PR #13, opened 2020-11-22
Fix Newton with finite diff

### LINKED ISSUE #12: jacobian erroneously calls finite_difference_derivative
This call should go to `FiniteDiff.finite_difference_jacobian`:

https://github.com/JuliaComputing/NonlinearSolve.jl/blob/cd00b8e4e3302be2cc724de126f5ca184d0e252e/src/jacobian.jl#L35

The reason the tests still pass is that the solution to the test problem is proportional to the initial guess, and the test function works for scalar `u`. If you change the initial guess so that it's not proportional to `[1.0, 1.0]` the algorithm will fail to converge, and if you use a function that fails for scalar `u` (e.g., by indexing into it) an exception will be thrown when trying to compute the jacobian.

YingboMa:
Fix #12

### DIFF
--- a/src/jacobian.jl
+++ b/src/jacobian.jl
@@ -32,7 +32,7 @@ function jacobian(f, x, solver)
     if alg_autodiff(solver.alg)
         J = ForwardDiff.jacobian(f, x)
     else
-        J = FiniteDiff.finite_difference_derivative(f, x, solver.alg.diff_type, eltype(x))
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
This is still non-allocating?

YingboMa:
FiniteDiff allocates.

PR #15, opened 2020-11-24
AD Support (Contd.)

kanav99:
@YingboMa


YingboMa:
This isn't the correct fix. It's go with Chris' idea of using a hybrid approach of using Falsi and Bisection together.

src/scalar.jl
```diff
@@ -21,26 +21,35 @@ end
 
 function scalar_nlsolve_ad(prob, alg, args...; kwargs...)
   f = prob.f
-  p = value(prob.p)
-  u0 = value(prob.u0)
+  p = value.(prob.p)
```
YingboMa, line 24: dot is not necessary here. See https://github.com/JuliaComputing/NonlinearSolve.jl/blob/master/src/utils.jl#L238-L240

kanav99: I think it doesn't handle the subtypes of AbstractArray well, it should be -
```julia
value(x::<:AbstractArray{<:Dual}) = map(ForwardDiff.value, x)
```
Am I right?

kanav99: Sorry my bad, that isn't even right syntax

PR #22, opened 2020-12-21
loosen types on prev/next floating handling

ChrisRackauckas:
Is there a reason why they were so tight?


kanav99:
I remember I loosened them a bit in some PR, maybe that didn't get merged. No reason for keeping them tight. This one looks good to me.

PR #28, opened 2021-01-31
Add SciML.build_solution

utkarsh530:
Hi,

I have currently added `build_solution` in solve.jl for NewtonRaphson. I did understand what was the purpose of `scalar.jl`. Do I need to add `build_solution` there too? And how to handle bracketing methods for `build_solution`?
@ChrisRackauckas, please have a look.

### DIFF
--- a/src/scalar.jl
+++ b/src/scalar.jl
@@ -52,12 +52,11 @@ end
 function SciMLBase.solve(prob::NonlinearProblem{<:Number, iip, <:Dual{T,V,P}}, alg::NewtonRaphson, args...; kwargs...) where {iip, T, V, P}
   sol, partials = scalar_nlsolve_ad(prob, alg, args...; kwargs...)
   return SciMLBase.build_solution(prob, alg, Dual{T,V,P}(sol.u, partials), sol.resid; retcode=sol.retcode)
-  #return NewtonSolution(Dual{T,V,P}(sol.u, partials), sol.retcode, sol.resid)
+  
 end
 function SciMLBase.solve(prob::NonlinearProblem{<:Number, iip, <:AbstractArray{<:Dual{T,V,P}}}, alg::NewtonRaphson, args...; kwargs...) where {iip, T, V, P}
   sol, partials = scalar_nlsolve_ad(prob, alg, args...; kwargs...)
   return SciMLBase.build_solution(prob, alg, Dual{T,V,P}(sol.u, partials), sol.resid; retcode=sol.retcode)
-  #return NewtonSolution(Dual{T,V,P}(sol.u, partials), sol.retcode, sol.resid)
 end
 
 # avoid ambiguities
--- a/src/scalar.jl
+++ b/src/scalar.jl
@@ -1,6 +1,7 @@
 function SciMLBase.solve(prob::NonlinearProblem{<:Number}, alg::NewtonRaphson, args...; xatol = nothing, xrtol = nothing, maxiters = 1000, kwargs...)
   f = Base.Fix2(prob.f, prob.p)
   x = float(prob.u0)
+  fx = float(prob.u0)
   T = typeof(x)
   atol = xatol !== nothing ? xatol : oneunit(T) * (eps(one(T)))^(4//5)
   rtol = xrtol !== nothing ? xrtol : eps(one(T))^(4//5)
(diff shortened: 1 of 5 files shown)


src/solve.jl
```diff
@@ -3,7 +3,15 @@ function SciMLBase.solve(prob::NonlinearProblem,
                          kwargs...)
   solver = init(prob, alg, args...; kwargs...)
   sol = solve!(solver)
-  return sol
+  if typeof(sol) <: NewtonSolution
+    resid = zero(prob.u0)
```
YingboMa, line 7: The allocation should be inside the `if` statement to save memory.

YingboMa (commented):
Doing one extra function evaluation for the residual seems pretty inefficient. I am fine with merge this as is, but we need to let the nonlinear solver itself forward the last computed residual.

ChrisRackauckas:
yeah it's already doing it to exit, so it might as well use the calculation of the exit criterion.

src/solve.jl
```diff
@@ -2,14 +2,8 @@ function SciMLBase.solve(prob::NonlinearProblem,
                          alg::AbstractNonlinearSolveAlgorithm, args...;
                          kwargs...)
   solver = init(prob, alg, args...; kwargs...)
-  sol = solve!(solver)
+  sol, resid = solve!(solver)
```
ChrisRackauckas, line 5: `solve!` has a standard interface that shouldn't be broken. We should just stick the `resid` in the `sol`.

src/solve.jl
```diff
@@ -2,8 +2,10 @@ function SciMLBase.solve(prob::NonlinearProblem,
                          alg::AbstractNonlinearSolveAlgorithm, args...;
                          kwargs...)
   solver = init(prob, alg, args...; kwargs...)
-  sol = solve!(solver)
-  return sol
+  sol, resid = solve!(solver)
+  if typeof(sol) <: NewtonSolution
```
YingboMa, line 6: Why is this branching needed? It makes the solution type unstable and break non-Newton solvers. We can just assume that `sol.u` exists.

utkarsh530: There are two types of solver, Newton solver and Bracketing methods, which outputs `sol.left` and `sol.right`. That's why I wanted to ask how to pass bracketing methods to `build_solution`?

src/solve.jl
utkarsh530: Okay, I will update the `sol` struct with `resid`.

src/scalar.jl
```diff
@@ -75,14 +76,14 @@ function SciMLBase.solve(prob::NonlinearProblem, ::Bisection, args...; maxiters
   fl, fr = f(left), f(right)
 
   if iszero(fl)
-    return BracketingSolution(left, right, EXACT_SOLUTION_LEFT)
+    return BracketingSolution(left, right, EXACT_SOLUTION_LEFT,fl)
```
utkarsh530, line 79: Do I need to return `SciMLBase.build_solution` here too? Because if we need to do, we need to handle how `u0` as a Tuple will be handled here: https://github.com/SciML/SciMLBase.jl/blob/master/src/solutions/nonlinear_solutions.jl#L26? What should it be in that case? @ChrisRackauckas

ChrisRackauckas: Maybe change it to `N = length((size(u)...,))` since it should not be able to change over the solution. Yes, it should use `build_solution` like all others, and treat `u` as the `left`.

ChrisRackauckas: or well, based on EXACT_SOLUTION_LEFT choose `u` as left etc.

utkarsh530:
With https://github.com/SciML/SciMLBase.jl/pull/17, this will complete the PR.

src/types.jl
```diff
@@ -41,22 +42,25 @@ struct NewtonImmutableSolver{fType, algType, uType, resType, pType, INType, tolT
...
+    prob::probType
 end
 
 # function NewtonImmutableSolver{iip}(iter, f, alg, u, fu, p, force_stop, maxiters, internalnorm, retcode, tol, cache) where iip
 #     NewtonImmutableSolver{iip, typeof(f), typeof(alg), typeof(u),
 #         typeof(fu), typeof(p), typeof(internalnorm), typeof(tol), typeof(cache)}(iter, f, alg, u, fu, p, force_stop, maxiters, internalnorm, retcode, tol, cache)
 # end
 
-struct BracketingSolution{uType}
+struct BracketingSolution{uType,resType}
```
YingboMa, line 53: Are we still using the original solution type?

utkarsh530: Yep, we don't need to it. Refactored the code. Thanks for the suggestion.

PR #30, opened 2021-03-26
Adds JuliaSim registry key to TagBot

ven-k:


ChrisRackauckas:
This is a public repo.

PR #56, opened 2022-03-10
Fix string concatenation error

KronosTheLate:
What is says on the box


ChrisRackauckas:
Awesome. We should probably add a test sometime that tests the error path.

PR #57, opened 2022-03-11
Fix docs typo

ericphanson:
There's no ODE here, right?


ChrisRackauckas:
Yeah, makes me sad too.

PR #79, opened 2022-08-02
More docs unification

ArnoStrouwen:


docs/src/index.md
```diff
@@ -1,6 +1,6 @@
-# NonlinearSolve.jl: High-Performance Unified Nonlinear Solvers
```
ChrisRackauckas, line 1: it should keep its title

ArnoStrouwen: But then it will look similar to this in the dropdown menu.
Capture
which is a bit ugly in my opinion.
So how do you want it? **Overview or home** on the dropdown and the actual name of the package on the corresponding page?

ChrisRackauckas: That doesn't come from the top of the page, it comes from the title description here: https://github.com/SciML/Optimization.jl/blob/master/docs/pages.jl#L2. But it's actually rather important for SEO reasons that the top file says "Optimization.jl" instead of "Home", otherwise Google will never pick up the page and put it to the top of the ranking as "Optimization.jl". So we really should be fixing the other ones to be more like DifferentialEquations.jl in that respect to fix how it interacts with search engines. Though I agree it would be nice to decouple that from what ends up in the sidebar.

ArnoStrouwen: Ok hopefully this can serve as a template for the other packages to be  made uniform.

docs/pages.jl
```diff
@@ -1,7 +1,7 @@
 # Put in a separate page so it can be used by SciMLDocs.jl
 
 pages = [
-    "Home" => "index.md",
+    "Overview" => "index.md",
```
ChrisRackauckas, line 4: We do need some kind of tag line.

Screenshot 2022-08-02 094516
 
Otherwise it won't do well in search. This doesn't necessarily need to be the label in the SciMLDocs.

ArnoStrouwen: I don't immediately see a way for the two to be separated?
The <title> element of https://nonlinearsolve.sciml.ai/ seems to be "first item on dropdown menu · NonLinearSolve.jl"
where NonLinearSolve.jl is what is used as the sitename keyword argument in the makedocs function.

ChrisRackauckas: We can make a bool that is set before `pages.jl` is loaded that then allowed for `is_scimldocs` and to switch whether to place the title on or not.

But I would greatly prefer just always having the title so it shows up on search engines, and making an issue to Documenter.jl to decouple the sidebar name from the title.

ArnoStrouwen: You can give a different sidebar name using that technique, but you still have to fill in something for the sidebar for `is_scimldocs=false`. 

Currently for DifferentialEquations.jl it is 

<title>DifferentialEquations.jl: Scientific Machine Learning (SciML) Enabled Simulation and Estimation · DifferentialEquations.jl</title>

Which has DifferentialEquations in there twice, once from the a name in the dropdown menu and once from the sitename keyword argument. 

So you could make the sitename argument an empty string. But then if you go to down the dropdown menu e.g

<title>Solving Large Stiff Equations · DifferentialEquations.jl</title>

It would also lose the DifferentialEquations, which is not great if somebody is searching for: "stiff equations DifferentialEquations.jl". 

I'll make an issue with Documenter.jl .

ArnoStrouwen:
The way it is done now should give:
<title>NonlinearSolve.jl: High-Performance Unified Nonlinear Solvers · SciML</title>
which is in my opinion the prettiest option given Documenter's current limitations.
Dropdown menu to be fixed later.

PR #88, opened 2022-09-15
Handle downstream usage issues

ChrisRackauckas:
It seems this library wasn't properly tested...


andreasnoack:
The tests here are not running at all. The code coverage has been zero for more than half a year https://app.codecov.io/gh/SciML/NonlinearSolve.jl. The reason is that you only run the tests for https://github.com/SciML/NonlinearSolve.jl/blob/22ed463491d2dd86758d90e79e297943e0b5914d/test/runtests.jl#L10 but `GROUP` never has any of those values https://github.com/SciML/NonlinearSolve.jl/blob/22ed463491d2dd86758d90e79e297943e0b5914d/.github/workflows/CI.yml#L15-L16. Might good to add an else branch that errors.

ChrisRackauckas:
Finally.

andreasnoack:
Can the [downstream failure](https://github.com/SciML/NonlinearSolve.jl/actions/runs/3539148420/jobs/5940729733#step:6:808) safely be ignored?

ChrisRackauckas:
Yes, this is getting a v1.0 release, and MTK is moving to SimpleNonlinerSolve.jl. It took awhile because this library had to be split to get the dependencies correct, which is a very breaking change. It also unifies the arguments (abstol instead of tol), so it's a very breaking update. Basically, it's finally updated from being an overlooked intern project to a full compliant piece of SciML, with all of the breakage required to make it conform to the rules of the interface. The update should be relatively sane though since most things would actually just move to SimpleNonlinearSolve.jl: the "bigger" Newton here is really just for big systems which isn't used all that much downstream.

PR #90, opened 2022-09-21
Missing LinearSolve dependency and compat

vpuri3:


vpuri3:
ok this package needs a lot of work. ill get to it later

PR #92, opened 2022-10-20
Change NonlinearSolve enums to SciMLBase enums

utkarsh530:
Provides compatibility to execute solves on GPU.

### DIFF
--- a/src/falsi.jl
+++ b/src/falsi.jl
@@ -21,7 +21,7 @@ function perform_step(solver, alg::Falsi, cache)
 
     if right == mid || right == mid
         @set! solver.force_stop = true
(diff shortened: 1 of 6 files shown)


src/falsi.jl
```diff
@@ -32,7 +32,7 @@ function perform_step(solver, alg::Falsi, cache)
         @set! solver.force_stop = true
         @set! solver.left = mid
         @set! solver.fl = fm
-        @set! solver.retcode = EXACT_SOLUTION_LEFT
```
ChrisRackauckas, line 35: Let's add these to SciMLBase and mark then as tokens of successful

PR #93, opened 2022-10-24
Start fixing CI

ArnoStrouwen:
The CI in some of the repos does not do anything, e.g.
https://github.com/SciML/SciMLOperators.jl/actions/runs/3119548680/jobs/5059578239#step:6:194
This is because there is insufficient control flow for `GROUP`.

It is not clear to me how _interface_ differs from _downstream_.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -7,7 +7,7 @@ const is_APPVEYOR = Sys.iswindows() && haskey(ENV, "APPVEYOR")
 
 @time begin
 
-if GROUP == "All" || GROUP == "Interface"
+if GROUP == "All" || GROUP == "Core" || GROUP == "Downstream" || GROUP == "Interface"
     #@time @safetestset "Linear Solver Tests" begin include("interface/linear_solver_test.jl") end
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Downstream always has a separate Project.toml

test/runtests.jl
```diff
@@ -7,7 +7,7 @@ const is_APPVEYOR = Sys.iswindows() && haskey(ENV, "APPVEYOR")
 
 @time begin
 
-if GROUP == "All" || GROUP == "Interface"
+if GROUP == "All" || GROUP == "Core" || GROUP == "Downstream" || GROUP == "Interface"
```
ChrisRackauckas, line 10: no test should ever be multi-group like this.

ArnoStrouwen: It would be nice if there was a design document that detailed things like this.
Currently,  you can not know based on this repo alone where some values of the environment variable, like Interface or Downstream will come from.

In any case, some CI in this repo passes trivially in a couple of microseconds because the environment variables are not set properly. This PR fixes it, but is probably too broad.

PR #94, opened 2022-11-04
add SnoopPrecompile and fix `_unwrap_val`

oscardssmith:
Adding precompile here should save time in the OrdinaryDiffEq precompile.


oscardssmith:
@YingboMa can we merge this? CI isn't passing but it wasn't before this PR either and this PR makes it so precompilation at least works.

ChrisRackauckas:
No merging here until https://github.com/SciML/NonlinearSolve.jl/pull/88

Issue #114, opened 2023-01-10
Error when using a neural network inside a system of nonlinear equations.


GevikUCL:
I've set up the following very simple system of nonlinear equations to solve.

```
using NonlinearSolve, Lux, LinearAlgebra, StableRNGs

rng = StableRNG(1111)

function func(u, p)
    [u[1].*u[1] .- p[1],
    u[2].*u[2] .- p[2]]
end

guess = [1., 1.]
params = [2.,3.]

prob = NonlinearProblem{false}(func, guess, params)
solution = solve(prob, NewtonRaphson())
```
This works fine. I am now trying to learn one of the terms from this system of equations using a neural network (similar to the UDE approach https://docs.sciml.ai/Overview/dev/showcase/missing_physics/).

I define a network.
```
ann = Lux.Chain(
    Lux.Dense(2,20,tanh), Lux.Dense(20,20,tanh), Lux.Dense(20,1)
)
p, st = Lux.setup(rng, ann)
```

And define the grey-box model, where I replace the `u[2]*u[2]` term with the output of the network. 

```
function GreyBox(u, p)
    z = ann(u, p, st)[1]
    [u[1].*u[1] .- params[1],
    z[1] .- params[2]]
end

NN_prob = NonlinearProblem{false}(GreyBox, guess, p)
NN_solution = solve(NN_prob NewtonRaphson())
```

In this case, `solve` isn't able to find the solution and the following error is returned.
```
ERROR: SingularException(2)
```

Was wondering if there would be a way to work around this issue.

ChrisRackauckas:
What's the error message? Please share the whole message.

CCsimon123:
I will make a PR tomorrow with a new `TrustRegion` solver to `NonlinearSolve.jl`
It will hopefully handle this by default, but I will make sure to test it!

ChrisRackauckas:
Indeed, this is just because a standard Newton cannot handle the singularity, but trust region can.

```julia
using NonlinearSolve, Lux, LinearAlgebra, StableRNGs

rng = StableRNG(1111)

function func(u, p)
    [u[1].*u[1] .- p[1],
    u[2].*u[2] .- p[2]]
end

guess = [1., 1.]
params = [2.,3.]

prob = NonlinearProblem{false}(func, guess, params)
solution = solve(prob)

ann = Lux.Chain(
    Lux.Dense(2,20,tanh), Lux.Dense(20,20,tanh), Lux.Dense(20,1)
)
[...]

PR #116, opened 2023-01-17
Added a Trust Region solver.

CCsimon123:
Implementation of a `TrustRegion` solver.
The solver is similar to the `TrustRegion` solver in `SimpleNonlinearSolve.jl`, but this new solver is implemented with `NewtonRaphson` as a template.

Questions:

1. Now the two `TrustRegion` solvers (in `SimpleNonlinearSolve.jl` and `NonlinearSolve.jl`) have the same name, should I change the name of the `TurstRegion` solver in `SimpleNonlinearSolve.jl` to `SimpleTrustRegion`?
2. Regarding https://github.com/SciML/NonlinearSolve.jl/issues/114, the `TurstRegion` solver does not give any error (as `NewtonRaphson` does). But the solver finds a solution with a "high" error. I suppose this is because it's getting stuck in a local minimum. But if you think that's not the case, please tell me.

As always, just tell me if there is anything I can improve/change.


src/NonlinearSolve.jl
```diff
@@ -31,22 +31,23 @@ end
...
+include("trustRegion.jl")
 include("ad.jl")
 
 import SnoopPrecompile
 
 SnoopPrecompile.@precompile_all_calls begin for T in (Float32, Float64)
     prob = NonlinearProblem{false}((u, p) -> u .* u .- p, T(0.1), T(2))
-    for alg in (NewtonRaphson,)
-        solve(prob, alg(), abstol = T(1e-2))
+    for alg in (NewtonRaphson(), TrustRegion(T(10.0)))
```
ChrisRackauckas, line 41: How is the max trust radius generally automated? Most other packages don't require this.

src/trustRegion.jl
```diff
@@ -0,0 +1,346 @@
...
+       AbstractNewtonAlgorithm{CS, AD, FDT, ST, CJ}
+    linsolve::L
+    precs::P
+    max_trust_radius::Number
+    initial_trust_radius::Number
+    step_threshold::Number
+    shrink_threshold::Number
+    expand_threshold::Number
+    shrink_factor::Number
+    expand_factor::Number
```
ChrisRackauckas, line 90: These should get specialized

src/trustRegion.jl
```diff
@@ -0,0 +1,346 @@
...
+    return nothing
+end
+
+function perform_step!(cache::TrustRegionCache{false})
+    @unpack make_new_J, fu, f, u, p = cache
+
+    if make_new_J
+        J = jacobian(cache, f)
+        cache.H = J * J
+        cache.g = J * fu
```
ChrisRackauckas, line 253:
```suggestion
        mul!(cache.H,J,J)
        mul!(cache.g,J,fu)
```

ChrisRackauckas: That first one might need a tmp to carry it in a non-aliasing way?

src/trustRegion.jl
```diff
@@ -0,0 +1,346 @@
...
+    return nothing
+end
+
+function perform_step!(cache::TrustRegionCache{false})
+    @unpack make_new_J, fu, f, u, p = cache
+
+    if make_new_J
+        J = jacobian(cache, f)
+        mul!(cache.H,J,J)
+        mul!(cache.g,J,fu)
```
ChrisRackauckas, line 253:
```suggestion
        cache.H = J * J
        cache.g = J * fu
```

src/trustRegion.jl
```diff
@@ -0,0 +1,346 @@
...
+                                 ReturnCode.Default, abstol, prob, alg.initial_trust_radius,
+                                 loss, loss, J, fu, 0, u, u, fu, true)
+end
+
+function perform_step!(cache::TrustRegionCache{true})
+    @unpack make_new_J, J, fu, f, u, p = cache
+    if cache.make_new_J
+        jacobian!(J, cache)
+        cache.H = J * J
+        cache.g = J * fu
```
ChrisRackauckas, line 234:
```suggestion
        mul!(cache.H,J,J)
        mul!(cache.g,J,fu)
```

src/trustRegion.jl
```diff
@@ -0,0 +1,346 @@
...
+    if cache.make_new_J
+        jacobian!(J, cache)
+        mul!(cache.H,J,J)
+        mul!(cache.g,J,fu)
+    end
+
+    dogleg!(cache)
+
+    # Compute the potentially new u
+    cache.u_new = u .+ cache.step_size
```
ChrisRackauckas, line 240:
```suggestion
    cache.u_new .= u .+ cache.step_size
```

src/trustRegion.jl
```diff
@@ -0,0 +1,346 @@
...
+    trust_region_step!(cache)
+    return nothing
+end
+
+function trust_region_step!(cache::TrustRegionCache)
+    @unpack fu_new, u_new, step_size, g, H, loss, alg = cache
+    cache.loss_new = get_loss(fu_new)
+
+    # Compute the ratio of the actual reduction to the predicted reduction.
+    model = -(step_size' * g + 0.5 * step_size' * H * step_size)
```
ChrisRackauckas, line 270: This needs to be written in a non-allocating style.

src/trustRegion.jl
```diff
@@ -0,0 +1,346 @@
...
+
+    if iszero(cache.fu) || cache.internalnorm(cache.step_size) < cache.abstol
+        cache.force_stop = true
+    end
+end
+
+function dogleg!(cache::TrustRegionCache)
+    @unpack g, H, trust_r = cache
+    # Compute the Newton step.
+    δN = -H \ g
```
ChrisRackauckas, line 306: This needs to use the LinearSolve choice.

src/utils.jl
```diff
@@ -124,3 +124,7 @@ function _nfcount(N, ::Type{diff_type}) where {diff_type}
     end
     tmp
 end
+
+function get_loss(fu)
+    return 0.5 * norm(fu)^2
```
ChrisRackauckas, line 129:
```suggestion
    return norm(fu)^2/2
```

ChrisRackauckas: The other way is not type-generic, i.e. Float32 would get promoted to Float64

ChrisRackauckas:
Note to @oscardssmith if he wants to take a look and help this get over the finish line.

ChrisRackauckas:
> Now the two TrustRegion solvers (in SimpleNonlinearSolve.jl and NonlinearSolve.jl) have the same name, should I change the name of the TurstRegion solver in SimpleNonlinearSolve.jl to SimpleTrustRegion?

I handled that. https://github.com/SciML/SimpleNonlinearSolve.jl/pull/28

src/trustRegion.jl
oscardssmith: specifically they should all be the same concrete type (and probably subtype `Real`)

oscardssmith:
How does this benchmark compared to NLSolve?

src/NonlinearSolve.jl
CCsimon123: When I've read about this it often seems quite ad-hoc. E.g. use fixed values for the specific problem or something like $||f(x)||=\Delta_{\text{max}}$ or $max(x) - min(x)$.

CCsimon123: But I'll look into it!

ChrisRackauckas: I think that's fine. Whatever value people would put in here would be more ad-hoc. I just wrote 10.0 because it seemed like a good number 😅

CCsimon123:
If the user inputs the exact solution as $u_0$ the forward diff seems to output the wrong value, Is this something I should handle as a special case?

CCsimon123:
And I'll have a look at benchmarking as soon as I finish this up :)

ChrisRackauckas:
>If the user inputs the exact solution as  the forward diff seems to output the wrong value, Is this something I should handle as a special case?

What do you mean? Example?

CCsimon123:
Sorry, just deleted the message.
give me 1 sec

CCsimon123:
This input gives error
```for p in 1.1:0.1:100.0
    @test g(p) ≈ sqrt(p)
    @test ForwardDiff.derivative(g, p) ≈ 1 / (2 * sqrt(p))
end```
ONLY for the last iteration (p = 100.0), when 
```g = function (p)
    probN = NonlinearProblem(f, oftype(p, 1.0), p)
    sol = solve(probN, TrustRegion(), abstol = 1e-10)
    return sol.u
end
```

ChrisRackauckas:
how far off is it?

CCsimon123:
```
  Expression: ForwardDiff.derivative(g, p) ≈ 1 / (2 * sqrt(p))
   Evaluated: 0.09090909090909091 ≈ 0.05
Stacktrace:
 [1] macro expansion
   @ C:\Users\ccsim\AppData\Local\Programs\julia-1.8.5\share\julia\stdlib\v1.8\Test\src\Test.jl:464 [inlined]
 [2] top-level scope
   @ c:\Users\ccsim\.julia\dev\NonlinearSolve.jl\test\basictests.jl:209
Test Summary:         | Pass  Fail  Total   Time
Basic Tests + Some AD | 7961     1   7962  53.6s
```

CCsimon123:
As off for now: I did the loop to `99.9`, and made a push on the comments above

ChrisRackauckas:
https://github.com/SciML/NonlinearSolve.jl/blob/master/src/ad.jl#L28-L37 expand this.

src/trustRegion.jl
```diff
@@ -0,0 +1,373 @@
...
+        mul!(cache.H, J, J)
+        mul!(cache.g, J, fu)
+    end
+
+    # u = u - J \ fu
+    linres = dolinsolve(alg.precs, linsolve, A = cache.H, b = _vec(cache.g),
+                        linu = _vec(du1),
+                        p = p, reltol = cache.abstol)
+    cache.linsolve = linres.cache
+    dogleg!(-du1, trust_r, cache)
```
ChrisRackauckas, line 262: `-du1` will allocate, needs a temporary.

src/trustRegion.jl
```diff
@@ -242,12 +243,12 @@ function SciMLBase.__init(prob::NonlinearProblem{uType, iip}, alg::TrustRegion,
     return TrustRegionCache{iip}(f, alg, u, fu, p, uf, linsolve, J, du1, jac_config,
                                  1, false, maxiters, internalnorm,
                                  ReturnCode.Default, abstol, prob, initial_trust_radius,
-                                 max_trust_radius, loss, loss, J, fu, 0, u, u, fu, true,
-                                 0.0)
+                                 max_trust_radius, loss, loss, copy(J), fu, 0, u, u, fu, true,
+                                 0.0, u)
```
ChrisRackauckas, line 247: a `u` term would alias the other?

src/trustRegion.jl
```diff
@@ -242,12 +243,12 @@ function SciMLBase.__init(prob::NonlinearProblem{uType, iip}, alg::TrustRegion,
     return TrustRegionCache{iip}(f, alg, u, fu, p, uf, linsolve, J, du1, jac_config,
                                  1, false, maxiters, internalnorm,
                                  ReturnCode.Default, abstol, prob, initial_trust_radius,
-                                 max_trust_radius, loss, loss, J, fu, 0, u, u, fu, true,
-                                 0.0)
+                                 max_trust_radius, loss, loss, copy(J), fu, 0, u, u, fu, true,
```
ChrisRackauckas, line 246: why is this copying? This should be fresh (line 205)

src/trustRegion.jl
```diff
@@ -0,0 +1,376 @@
...
+        mul!(cache.H, J, J)
+        mul!(cache.g, J, fu)
+    end
+
+    # u = u - J \ fu
+    linres = dolinsolve(alg.precs, linsolve, A = cache.H, b = _vec(cache.g),
+                        linu = _vec(du1),
+                        p = p, reltol = cache.abstol)
+    cache.linsolve = linres.cache
+    cache.δN = -du1
```
ChrisRackauckas, line 263:
```suggestion
    cache.δN .= -1 .* du1
```

src/trustRegion.jl
```diff
@@ -0,0 +1,376 @@
...
+        cache.make_new_J = true
+    else
+        # No need to make a new J, no step was taken, so we try again with a smaller trust_r
+        cache.make_new_J = false
+    end
+
+    if iszero(cache.fu) || cache.internalnorm(cache.step_size) < cache.abstol
+        cache.force_stop = true
+    end
+end
```
ChrisRackauckas, line 332: I don't see `du1` used in here.  So can't we just do `cache.du1 .= -1 .* du1` and then use it in here to eliminate a tmp vector?

CCsimon123: sure

CCsimon123: Hmm... Tried to replace $\delta N$, but did you have another vector in mind?

CCsimon123:
whops, give me 1 sec, forgot to run the tests... (error atm)

src/trustRegion.jl
CCsimon123: I've now replaced $\delta N$, du1 and u_new with a single vector u_tmp

CCsimon123:
> https://github.com/SciML/NonlinearSolve.jl/blob/master/src/ad.jl#L28-L37 expand this.

Solved the problem with `alg::Union{NewtonRaphson, TrustRegion}` Not 100% sure if this is ok, but it works

ChrisRackauckas:
> Solved the problem with alg::Union{NewtonRaphson, TrustRegion} Not 100% sure if this is ok, but it works

That's the correct thing to do.

test/basictests.jl
```diff
@@ -121,3 +123,167 @@ for u0 in [1.0, [1, 1.0]]
...
+function benchmark_immutable(f, u0)
+    probN = NonlinearProblem(f, u0)
+    solver = init(probN, TrustRegion(), abstol = 1e-9)
+    sol = solve!(solver)
+end
+
+function benchmark_mutable(f, u0)
+    probN = NonlinearProblem(f, u0)
+    solver = init(probN, TrustRegion(), abstol = 1e-9)
+    sol = solve!(solver)
```
ChrisRackauckas, line 138: Can you add a test that `solve!` on the mutable code path does not allocate? I think if that passes then it's done.

src/trustRegion.jl
```diff
@@ -0,0 +1,374 @@
...
+    trust_region_step!(cache)
+    return nothing
+end
+
+function trust_region_step!(cache::TrustRegionCache)
+    @unpack fu_new, u_tmp, step_size, g, H, loss, alg, max_trust_r = cache
+    cache.loss_new = get_loss(fu_new)
+
+    # Compute the ratio of the actual reduction to the predicted reduction.
+    cache.r = -(loss - cache.loss_new) / (step_size' * g + step_size' * H * step_size / 2)
```
ChrisRackauckas, line 299: This still needs to be made non-allocating.

CCsimon123:
Tried to fix the allocation problem (get them < 400), but why did you make `benchmark_mutable` and `benchmark_immutable` the same with the same input in your code. I just used yours as a template, but I can't really understand why you do this.

test/basictests.jl
```diff
@@ -127,19 +127,19 @@ end
...
 function benchmark_immutable(f, u0)
-    probN = NonlinearProblem(f, u0)
+    probN = NonlinearProblem{false}(f, u0)
     solver = init(probN, TrustRegion(), abstol = 1e-9)
     sol = solve!(solver)
 end
 
 function benchmark_mutable(f, u0)
-    probN = NonlinearProblem(f, u0)
+    probN = NonlinearProblem{false}(f, u0)
```
ChrisRackauckas, line 136: this one should be true

test/basictests.jl
```diff
@@ -161,6 +161,10 @@ sol = benchmark_scalar(sf, csu0)
 @test sol.retcode === ReturnCode.Success
 @test sol.u * sol.u - 2 < 1e-9
 
+@test (@ballocated benchmark_immutable(ff, cu0)) < 400
+@test (@ballocated benchmark_mutable(ff, cu0)) < 400
```
ChrisRackauckas, line 165: the allocation test we should check on this one shouldn't include the problem building.

ChrisRackauckas:
> but why did you make benchmark_mutable and benchmark_immutable the same with the same input in your code

That looks like it's a typo.

CCsimon123:
`NewtonRaphson` allocates 48 while `TrustRegion` allocates 112 (for the inplace case)

ChrisRackauckas:
yeah my guess is that it's mostly line 299.

Let's get this into a form that can be merged and follow up with fixing that.

CCsimon123:
Sure!
What needs to be done to get it into that form?

CCsimon123:
(I don't really understand why some test cases, e.g "IntegrationTest / OrdinaryDiffEq.jl" fail and what to do about it)

ChrisRackauckas:
It seems like it's a v1.6 only error?

CCsimon123:
yeah, sorry for not understanding... but I don't know what to do to fix it :(

ChrisRackauckas:
Precompilation of `TrustRegion` causes segfaults only on v1.6. That seems like it's a Julia bug that was fixed. Let's merge like this if tests pass, but then follow up by enabling it for `VERSION >= v"1.7"`

CCsimon123:
I want to say a huge thank you for all the help @ChrisRackauckas!
I'm learning so much from this (and is a lot of fun) :)
And sorry for all the time you have to put on me, hopefully, it will pay off soon as I intend to continue to do PRs.

If you would like me to continue with something specific, please tell me, otherwise I assume a good starting point is to implement the new solvers you just added to the issues.

PS. Thank you for the invite to SciML, means a lot to me :D

PR #130, opened 2023-01-22
format docstrings

ArnoStrouwen:
When https://github.com/domluna/JuliaFormatter.jl/blob/d89ab9e2638b9273b9d25321cbed5c9f22719e6d/src/styles/sciml/pretty.jl#L41 is changed to true.

I'll go through the repos and format the docstrings,
if that works for all the repos we can change it in the style?


ChrisRackauckas:
> if that works for all the repos we can change it in the style?

Yes

Issue #119, opened 2023-01-17
Implement Levenburg-Marquedt


ChrisRackauckas:
https://docs.scipy.org/doc/scipy/reference/optimize.root-lm.html

https://en.wikipedia.org/wiki/Levenberg%E2%80%93Marquardt_algorithm

oscardssmith:
this should be closed.

Issue #128, opened 2023-01-22
package loading time regression


daviehh:
Looks like it's due to https://github.com/SciML/NonlinearSolve.jl/issues/121 and #127

With version 1.1.1
```
@time_imports using NonlinearSolve
```
gives
> 142.4 ms  NonlinearSolve

while with the 1.2.0 version, this shot up to ~13 seconds,
> 13191.1 ms  NonlinearSolve

Tested w/
```
julia> versioninfo()
Julia Version 1.8.5
Commit 17cfb8e65ea (2023-01-08 06:45 UTC)
Platform Info:
  OS: macOS (arm64-apple-darwin21.5.0)
  CPU: 10 × Apple M1 Pro
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, apple-m1)
  Threads: 1 on 8 virtual cores
```

---

By reverting the changes in #127, the long `using NonlinearSolve` time is fixed. So maybe it would be better to revert https://github.com/SciML/NonlinearSolve.jl/pull/127 until it can be figured out where it went wrong?

Thanks!

ChrisRackauckas:
Can you check v1.9-beta3 and test a full TTFX example? `using` time needs to be put into context as increasing `using` time can be associated with decreased first call times.

daviehh:
With 1.9-beta3, the TTFX does not appear to be affected much by including or not including the `TrustRegion()` in, but loading the new 1.9 cached pkgimage also takes a long time with `TrustRegion()` in the precompile.

```julia
@time begin
    prob = NonlinearProblem{false}((u, p) -> u .* u .- p, 0.1, 2)
    solve(prob, TrustRegion(), abstol = 1e-2)
end
```

*With* the `precompile_algs = (NewtonRaphson(), TrustRegion())` in the precompile script, the problem solving example above gives
> 0.177028 seconds (778.01 k allocations: 55.147 MiB, 5.30% gc time)

and the total using time (after precompile with a fresh start of julia) is
> 14.901253 seconds (71.82 M allocations: 14.932 GiB, 8.68% gc time, 0.04% compilation time)

---

*Without* the `TrustRegion()` algo in precompile: `precompile_algs = (NewtonRaphson(),)`, the example takes
> 0.174154 seconds (791.75 k allocations: 56.127 MiB, 5.85% gc time)

and the total using time
[...]

PR #131, opened 2023-01-22
Implementation of a Levenberg–Marquardt algorithm

CCsimon123:
Implementation of a Levenberg–Marquardt algorithm #119, according to the paper [Improvements to the Levenberg-Marquardt algorithm for nonlinear least-squares minimization](https://arxiv.org/abs/1201.5885).
The solver in `raphson.jl` was used as a template for this solver.

### DIFF
--- a/src/ad.jl
+++ b/src/ad.jl
@@ -26,7 +26,7 @@ end
 function SciMLBase.solve(prob::NonlinearProblem{<:Union{Number, StaticArraysCore.SVector},
                                                 iip,
                                                 <:Dual{T, V, P}},
-                         alg::Union{NewtonRaphson, TrustRegion},
+                         alg::Union{NewtonRaphson, TrustRegion, LevenbergMarquardt},
                          args...; kwargs...) where {iip, T, V, P}
     sol, partials = scalar_nlsolve_ad(prob, alg, args...; kwargs...)
     return SciMLBase.build_solution(prob, alg, Dual{T, V, P}(sol.u, partials), sol.resid;
@@ -35,7 +35,8 @@ end
 function SciMLBase.solve(prob::NonlinearProblem{<:Union{Number, StaticArraysCore.SVector},
                                                 iip,
                                                 <:AbstractArray{<:Dual{T, V, P}}},
-                         alg::Union{NewtonRaphson, TrustRegion}, args...;
+                         alg::Union{NewtonRaphson, TrustRegion, LevenbergMarquardt},
+                         args...;
                          kwargs...) where {iip, T, V, P}
     sol, partials = scalar_nlsolve_ad(prob, alg, args...; kwargs...)
     return SciMLBase.build_solution(prob, alg, Dual{T, V, P}(sol.u, partials), sol.resid;
--- a/src/ad.jl
+++ b/src/ad.jl
@@ -26,7 +26,7 @@ end
 function SciMLBase.solve(prob::NonlinearProblem{<:Union{Number, StaticArraysCore.SVector},
                                                 iip,
                                                 <:Dual{T, V, P}},
-                         alg::Union{NewtonRaphson, TrustRegion, LevenbergMarquardt},
+                         alg::AbstractNewtonAlgorithm,
                          args...; kwargs...) where {iip, T, V, P}
     sol, partials = scalar_nlsolve_ad(prob, alg, args...; kwargs...)
     return SciMLBase.build_solution(prob, alg, Dual{T, V, P}(sol.u, partials), sol.resid;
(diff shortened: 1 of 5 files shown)


src/ad.jl
```diff
@@ -26,7 +26,7 @@ end
 function SciMLBase.solve(prob::NonlinearProblem{<:Union{Number, StaticArraysCore.SVector},
                                                 iip,
                                                 <:Dual{T, V, P}},
-                         alg::Union{NewtonRaphson, TrustRegion},
+                         alg::Union{NewtonRaphson, TrustRegion, LevenbergMarquardt},
```
ChrisRackauckas, line 29:
```suggestion
                         alg::AbstractNewtonAlgorithm,
```

src/ad.jl
```diff
@@ -35,7 +35,8 @@ end
 function SciMLBase.solve(prob::NonlinearProblem{<:Union{Number, StaticArraysCore.SVector},
                                                 iip,
                                                 <:AbstractArray{<:Dual{T, V, P}}},
-                         alg::Union{NewtonRaphson, TrustRegion}, args...;
+                         alg::Union{NewtonRaphson, TrustRegion, LevenbergMarquardt},
```
ChrisRackauckas, line 38:
```suggestion
                         alg::AbstractNewtonAlgorithm,
```

src/levenberg.jl
```diff
@@ -0,0 +1,408 @@
...
+    @unpack u, p, λ, JᵀJ, DᵀD, J = cache
+
+    # Usual Levenberg-Marquardt step ("velocity").
+    cache.v = -(JᵀJ .+ λ .* DᵀD) \ mul!(cache.du_tmp, J', fu)
+
+    # Geodesic acceleration (step_size = v + a / 2).
+    @unpack v, alg = cache
+    h = alg.finite_diff_step_geodesic
+    f(cache.fu_tmp, u .+ h .* v, p)
+    cache.a = -J \ ((2 / h) .* ((cache.fu_tmp .- fu) ./ h .- mul!(cache.du_tmp, J, v)))
```
ChrisRackauckas, line 306: The two `\`'s should be changed to `linsolve` handlings.

src/levenberg.jl
```diff
@@ -0,0 +1,408 @@
...
+    if make_new_J
+        jacobian!(cache.J, cache)
+        mul!(cache.JᵀJ, cache.J', cache.J)
+        cache.DᵀD .= max.(cache.DᵀD, Diagonal(cache.JᵀJ))
+        cache.make_new_J = false
+    end
+    @unpack u, p, λ, JᵀJ, DᵀD, J = cache
+
+    # Usual Levenberg-Marquardt step ("velocity").
+    cache.v = -(JᵀJ .+ λ .* DᵀD) \ mul!(cache.du_tmp, J', fu)
```
ChrisRackauckas, line 300: `-(JᵀJ .+ λ .* DᵀD)` should be put into a temporary

ChrisRackauckas: `mul!(cache.du_tmp, J', fu)` I'd move that to a different line for clarity. Though it returns `cache.du_tmp` in most dispatches, it's usually best to mutate in one line and then use in the next

CCsimon123:
Edit:
I will include this algorithm in `precompile_algs`. I naively thought I would have the same loading time as the 13 sec stated in #128, and when I got 20 sec for all three algorithms I thought the last 7 sec was due to the newly added `LevenbergMarquardt`.
And when I removed both `LevenbergMarquardt` and `TrustRegion` the loading time vanished.
But I now tested with only `LevenbergMarquardt` and `NewtonRaphson` and the loading time was good. While for `TrustRegion` and `NewtonRaphson` Is was 20 sec. Therefore I'll now include `LevenbergMarquardt` in the `precompile_algs`.

CCsimon123:
Testing to remove the precompilation of `LevenbergMarquardt` to see if it fixes the error in the check "IntegrationTest / ModelingToolkit.jl/All/1 (pull_request)"

CCsimon123:
Think I've found the issue from #128.
Have to go now, unfortunately, but I'll be back in 12h again.

src/levenberg.jl
```diff
@@ -0,0 +1,424 @@
...
+    # Geodesic acceleration (step_size = v + a / 2).
+    @unpack v, alg = cache
+    h = alg.finite_diff_step_geodesic
+    f(cache.fu_tmp, u .+ h .* v, p)
+
+    # The following lines do: cache.a = -J \ cache.fu_tmp
+    mul!(cache.du_tmp, J, v)
+    @. cache.fu_tmp = (2 / h) * ((cache.fu_tmp - fu) / h - cache.du_tmp)
+    linres = dolinsolve(alg.precs, linsolve, A = J, b = _vec(cache.fu_tmp),
+                        linu = _vec(cache.du_tmp), p = p, reltol = cache.abstol)
```
ChrisRackauckas, line 320: We may want to split the preconditioners between these two. But that's for follow up.

ChrisRackauckas:
This looks good for a first pass. More to be done, but we should split correctness from improvements.

PR #132, opened 2023-01-23
type instability fix

### LINKED ISSUE #128: package loading time regression

daviehh:
Tentative PR for the type instability/runtime dispatch and the resulting long loading time, may fix #128
Maybe it may be handled in a more clean/elegant fashion..

Also, since `eltype(<:Number)` just returns the type of the number, the line
```
u_elType = uType <: Number ? uType : eltype(u)
```
maybe just changed to `u_elType = eltype(u)`. Not sure how robust it can be


CCsimon123:
Nice one!
Just before I had to leave yesterday I also found the problem and solved it with 
```
max_trust_radius = convert(typeof(max_trust_radius), max(norm(fu), maximum(u) - minimum(u)))
```
in #131. Sorry for not telling you, but I thought this comment https://github.com/SciML/NonlinearSolve.jl/pull/131#issuecomment-1399562351 would trigger a note in #128.
Anyhow, maybe your way to solve it is more elegant, nice work :)

ChrisRackauckas:
```julia
                     max_trust_radius::Real = 0.0,
                     initial_trust_radius::Real = 0.0,
                     step_threshold::Real = 0.1,
                     shrink_threshold::Real = 0.25,
                     expand_threshold::Real = 0.75,
                     shrink_factor::Real = 0.25,
                     expand_factor::Real = 2.0,
```

what we really should be doing is change those values to `Rational` numbers. If that is done, they will convert on-demand when hitting `u` types, for example `1//2 * 0.33 == 0.165` and `1//2 * 3.3f-2 == 0.0165f0`. That makes the default values "type neutral". I wouldn't require that the numbers are rational, I would expect users to give Float64 value (or correctly type-matching values), but since the overwhelming default will be for the user to just use the default values of the algorithm parameters, this would be a safer option. The correct conversions should also then be done when the values are put into the cache object as well, doing a `convert(eltype(u),x)` on all of them.

CCsimon123:
hmm, I made the changes suggested by Chris, but I wasn't allowed to push to @daviehh
(Think its because: "The author of the PR also may need to explicitly allow you to push to their branch." in https://gist.github.com/wtbarnes/56b942641d314522094d312bbaf33a81)

ChrisRackauckas:
You can PR to his branch.

ChrisRackauckas:
Hmm something happened with the merge ambiguity fix. It's probably just a simple thing with the type parameters.

CCsimon123:
The problem is that `step_threshold` and `step_size` got different types (`trustType` and `suType`) see the "arrows" (<-----------------)
```julia
    step_threshold::trustType <-----------------
    shrink_threshold::trustType
    expand_threshold::trustType
    shrink_factor::trustType
    expand_factor::trustType
    loss::floatType
    loss_new::floatType
    H::jType
    g::resType
    shrink_counter::Int
    step_size::suType  <-----------------
    u_tmp::tmpType
    fu_new::resType
    make_new_J::Bool
    r::floatType

    function TrustRegionCache{iip}(f::fType, alg::algType, u::uType, fu::resType, p::pType,
                                   uf::ufType, linsolve::L, J::jType,
                                   jac_config::JC, iter::Int,
                                   force_stop::Bool, maxiters::Int, internalnorm::INType,
                                   retcode::SciMLBase.ReturnCode.T, abstol::tolType,
                                   prob::probType, trust_r::trustType,
                                   max_trust_r::trustType, step_threshold::suType, <-----------------
                                   shrink_threshold::trustType, expand_threshold::trustType,
                                   shrink_factor::trustType, expand_factor::trustType,
                                   loss::floatType, loss_new::floatType, H::jType,
                                   g::resType, shrink_counter::Int, step_size::uType, <-----------------
                                   u_tmp::tmpType, fu_new::resType, make_new_J::Bool,
```
Unfortunately I got no time to fix it right now, but if its not fixed in ~8h I'll have time to do it

src/trustRegion.jl
```diff
@@ -140,6 +140,11 @@ mutable struct TrustRegionCache{iip, fType, algType, uType, resType, pType,
     prob::probType
     trust_r::trustType
     max_trust_r::trustType
+    step_threshold::trustType
```
ChrisRackauckas, line 143:
```suggestion
    step_threshold::suType
```

CCsimon123: I don't think this will work, `step_threshold` is a scalar and `suType` seems to be a vector.

I suppose https://github.com/SciML/NonlinearSolve.jl/pull/135 was for a good reason, but when I try the program with
```julia
step_size::uType
    u_tmp::uType
```
and not the new types, the first execution is about 4 times faster...
Will look more into the timing now, but is it necessary to have the new types?

CCsimon123: (and `step_threshold::trustType`)

ChrisRackauckas: Oh okay if you're back cool. I was just poking at it through quick commits and checking CI results seeing if I can get this solved while hacking on something else, But if you have a second to push this over the line, it probably just needs a different type parameter. I cannot say I was looking too closely.

CCsimon123: Yeah, I've got some time now. I'll probably have to make a PR to daviehh (original PR creator) but hopefully, he accepts it soon enough.

PR #134, opened 2023-01-23
Updating the types in Levenberg

CCsimon123:
Changing the types in `levenberg.jl` to match the input `u0` given by the user as https://github.com/SciML/NonlinearSolve.jl/pull/132#issuecomment-1399911499 suggests for `trustRegion.jl`


CCsimon123:
I don't find any place where making the input rational would change anything.
In all cases, I just take the input given by the user or the default values
```
damping_initial::Real = 1.0,
damping_increase_factor::Real = 2.0,
damping_decrease_factor::Real = 3.0,
finite_diff_step_geodesic::Real = 0.1,
α_geodesic::Real = 0.75,
b_uphill::Real = 1.0,
min_damping_D::AbstractFloat = 1e-8)
```
and then convert it to the element type of `u` by `x = convert(eltype(u), alg.x)` before putting `x` into `cache`.

PR #159, opened 2023-03-06
Implementing some new Trust region radius update schemes

yash2798:


yash2798:
@ChrisRackauckas  is this the correct way to implement this? here, we take the input from the user regarding what update scheme they would like to use as a String `solve(probN, TrustRegion(radius_update_scheme="<>"), abstol = 1e-9)`. and then use this input subsequently?

ChrisRackauckas:
>we take the input from the user regarding what update scheme they would like to use as a String solve(probN, TrustRegion(radius_update_scheme="<>"), abstol = 1e-9). and then use this input subsequently?

No, never take in a string. Take in a type instance, like the algorithm choice.

yash2798:
> > we take the input from the user regarding what update scheme they would like to use as a String solve(probN, TrustRegion(radius_update_scheme="<>"), abstol = 1e-9). and then use this input subsequently?
> 
> No, never take in a string. Take in a type instance, like the algorithm choice.

okay, got it, so I define all the radius update schemes as composite types? and then use them? just like the `TrustRegion` type is defined?

yash2798:
Hi @ChrisRackauckas , I made some changes regarding how we choose the relevant radius update scheme. Is this approach alright?

src/trustRegion.jl
```diff
@@ -94,10 +95,33 @@ struct TrustRegion{CS, AD, FDT, L, P, ST, CJ, MTR} <:
     max_shrink_times::Int
 end
 
+struct RadiusUpdate{B}
+    simple::B
+    hei::B
+    yuan::B
+    bastin::B
```
ChrisRackauckas, line 102:
```suggestion
struct RadiusUpdate
    simple::Bool
    hei::Bool
    yuan::Bool
    bastin::Bool
```

src/trustRegion.jl
```diff
@@ -94,10 +95,33 @@ struct TrustRegion{CS, AD, FDT, L, P, ST, CJ, MTR} <:
...
+                      yuan::Bool = Val{false}(),
+                      bastin::Bool = Val{false}())
+    if simple
+      return RadiusUpdate{Bool}(true, false, false, false)
+    elseif hei
+      return RadiusUpdate{Bool}(false, true, false, false)
+    elseif yuan
+      return RadiusUpdate{Bool}(false, false, true, false)
+    elseif bastin
+      return RadiusUpdate{Bool}(false, false, false, true)
```
ChrisRackauckas, line 116:
```suggestion
      return RadiusUpdate{Bool}(true, false, false, false)
    elseif hei
      return RadiusUpdate{Bool}(false, true, false, false)
    elseif yuan
      return RadiusUpdate(false, false, true, false)
    elseif bastin
      return RadiusUpdate(false, false, false, true)
```

src/trustRegion.jl
```diff
@@ -94,10 +95,33 @@ struct TrustRegion{CS, AD, FDT, L, P, ST, CJ, MTR} <:
...
+
+function RadiusUpdate(;simple::Bool = Val{true}(),  # 3 different radius update schemes
+                      hei::Bool = Val{false}(),
+                      yuan::Bool = Val{false}(),
+                      bastin::Bool = Val{false}())
+    if simple
+      return RadiusUpdate{Bool}(true, false, false, false)
+    elseif hei
+      return RadiusUpdate{Bool}(false, true, false, false)
+    elseif yuan
```
ChrisRackauckas, line 113: This doesn't make sense. If it's going to be non-dispatching logic, use an EnumX.

yash2798: Oh yes, I am sorry I misinterpreted the use of `Val`, but now it's clear. I have changed the overall approach.

yash2798:
@ChrisRackauckas  I think this approach would be better! It would be great if you can take a look.

I have a question though:
In line 103, the `radius_update_scheme` input is a Symbol. Another option is to use `Val{:simple}()` which will change the function call for `trust_region_step!` in lines 299 and 322. So, should the user input be kept as a Symbol (:simple, :hei, etc.) or should I change it to `Val{:simple}()`. I mean, there is not a lot of difference between the two but what method would be better from a user's pov?

yash2798:
and thanks for all the feedback that you have been giving, I am learning a lot :)

ChrisRackauckas:
> In line 103, the radius_update_scheme input is a Symbol. Another option is to use Val{:simple}() which will change the function call for trust_region_step! in lines 299 and 322. So, should the user input be kept as a Symbol (:simple, :hei, etc.) or should I change it to Val{:simple}(). I mean, there is not a lot of difference between the two but what method would be better from a user's pov?

`Val{:sample}()` is a valid construction, so that doesn't give good errors for typos. 

See the ReturnCodes implementation and EnumX.jl 

https://github.com/SciML/SciMLBase.jl/blob/master/src/retcodes.jl

yash2798:
Okay, so by this you mean that if the user makes a typo `:sample` instead of the required `:simple`, then the program will show an exception something like this `no method matching trust_region_step!(cache::TrustRegionCache, ::Val{:sample})` and this error is not very clear for the user? Is this what you mean?

If yes, couldn't we simply check for the validity of the input beforehand and raise an exception in case of any mismatch?

yash2798:
@ChrisRackauckas Is this approach with EnumX alright? I have used my previous approach of creating a `RadiusUpdate` type variable and then used enumx just for multiple dispatching.

src/trustRegion.jl
```diff
@@ -80,10 +80,25 @@ for large-scale and numerically-difficult nonlinear systems.
...
 """
-struct TrustRegion{CS, AD, FDT, L, P, ST, CJ, MTR} <:
+EnumX.@enumx RadiusUpdateSchemes begin
+    Simple
+    Hei
+    Yuan
+    Bastin
+end
+
+struct RadiusUpdate
```
YingboMa, line 90: Aren't they mutually exclusive? We should just use enum.

ChrisRackauckas:
Everything should use the enums and the structs should be deleted.

src/trustRegion.jl
```diff
@@ -80,10 +80,18 @@ for large-scale and numerically-difficult nonlinear systems.
...
+    Hei
+    Yuan
+    Bastin
+end
+
+struct TrustRegion{CS, AD, FDT, L, P, ST, CJ, MTR, RUS} <:
        AbstractNewtonAlgorithm{CS, AD, FDT, ST, CJ}
     linsolve::L
     precs::P
+    radius_update_scheme::RUS
```
ChrisRackauckas, line 94:
```suggestion
    radius_update_scheme::RadiusUpdateSchemes.T
```

It doesn't need to be parametric.,

src/trustRegion.jl
```diff
@@ -138,6 +147,7 @@ mutable struct TrustRegionCache{iip, fType, algType, uType, resType, pType,
     retcode::SciMLBase.ReturnCode.T
     abstol::tolType
     prob::probType
+    radius_update_scheme::radType
```
ChrisRackauckas, line 150: here too

src/trustRegion.jl
```diff
@@ -312,11 +324,12 @@ function perform_step!(cache::TrustRegionCache{false})
...
     cache.fu_new = f(cache.u_tmp, p)
 
-    trust_region_step!(cache)
+    @unpack radius_update_scheme = cache
+    trust_region_step!(cache, Val(Int(radius_update_scheme)))
     return nothing
 end
 
-function trust_region_step!(cache::TrustRegionCache)
+function trust_region_step!(cache::TrustRegionCache, ::Val{0}) # conventional radius update scheme
```
ChrisRackauckas, line 332: Make a single dispatch and branch

src/NonlinearSolve.jl
```diff
@@ -10,6 +10,7 @@ using ForwardDiff: Dual
 using LinearAlgebra
 using StaticArraysCore
 using RecursiveArrayTools
+import EnumX
```
ChrisRackauckas, line 13: missing from the Project.toml

yash2798: yeah i added the pkg in the env but didn't commit it yet. will do it in the next commit

src/trustRegion.jl
ChrisRackauckas: yes

yash2798: cool! also if I need some helper functions, like in Hei's scheme, there is a monotonic function used which depends on the actual:predicted change ratio. So these kind of helper functions can go in `utils.jl`, right?

src/trustRegion.jl
```diff
@@ -312,42 +324,55 @@ function perform_step!(cache::TrustRegionCache{false})
...
-    if r < cache.shrink_threshold
-        cache.trust_r *= cache.shrink_factor
-        cache.shrink_counter += 1
-    else
-        cache.shrink_counter = 0
-    end
-    if r > cache.step_threshold
-        take_step!(cache)
-        cache.loss = cache.loss_new
+    if radius_update_scheme == RadiusUpdateSchemes.Simple
```
ChrisRackauckas, line 340:
```suggestion
    if radius_update_scheme === RadiusUpdateSchemes.Simple 
```

src/trustRegion.jl
```diff
@@ -312,42 +324,55 @@ function perform_step!(cache::TrustRegionCache{false})
...
+          cache.force_stop = true
+      end
+    
+    elseif radius_update_scheme == RadiusUpdateSchemes.Hei
+
+
+    elseif radius_update_scheme == RadiusUpdateSchemes.Yuan
+
+
+    elseif radius_update_scheme == RadiusUpdateSchemes.Bastin
```
ChrisRackauckas, line 373:
```suggestion
    elseif radius_update_scheme === RadiusUpdateSchemes.Hei


    elseif radius_update_scheme === RadiusUpdateSchemes.Yuan


    elseif radius_update_scheme === RadiusUpdateSchemes.Bastin
```

src/trustRegion.jl
```diff
@@ -312,42 +324,55 @@ function perform_step!(cache::TrustRegionCache{false})
...
+      else
+          # No need to make a new J, no step was taken, so we try again with a smaller trust_r
+          cache.make_new_J = false
+      end
+
+      if iszero(cache.fu) || cache.internalnorm(cache.fu) < cache.abstol
+          cache.force_stop = true
+      end
+    
+    elseif radius_update_scheme === RadiusUpdateSchemes.Hei
```
ChrisRackauckas, line 367: error for not implemented

yash2798: okay, like a TypeError when the `radius_update_scheme` type doesn't match `RadiusUpdateSchemes.T` ?

src/trustRegion.jl
```diff
@@ -262,13 +282,37 @@ function SciMLBase.__init(prob::NonlinearProblem{uType, iip}, alg::TrustRegion,
...
+      p4 = 0.15 # γ2
+    elseif radius_update_scheme === RadiusUpdateSchemes.Yuan
+      step_threshold = 0.0001
+      shrink_threshold = 0.25
+      expand_threshold = 0.25
+      p1 = 2.0 # μ
+      p2 = 1/6 # c5
+      p3 = 6 # c6
+      p4 = 0
+    end
```
ChrisRackauckas, line 307: these values should match the type of `u0`.

src/trustRegion.jl
```diff
@@ -312,42 +357,89 @@ function perform_step!(cache::TrustRegionCache{false})
...
+      if r < cache.shrink_threshold
+        cache.p1 = p2 * cache.p1
+      elseif r >= cache.shrink_threshold && cache.internalnorm(step_size) > cache.trust_r / 2
+        cache.p1 = p3 * cache.p1
+      end
+      @unpack p1 = cache.p1
+
+      # yuan's scheme
+      @unpack fu = cache
+      cache.trust_r = p1 * cache.internalnorm(jacobian(cache, f) * fu) # we need the gradient at the new (k+1)th point  WILL THIS BECOME ALLOCATING?
```
ChrisRackauckas, line 434: you don't needto calculate the whole Jacobian here, we can fix that in a future PR though.

ChrisRackauckas:
Last step here is to make sure the tests don't fail, and fix the constants definitions so that they match the state types.

yash2798:
> Last step here is to make sure the tests don't fail, and fix the constants definitions so that they match the state types.

Oh yes thanks, I understood it. I was just testing the two new methods myself and will make a commit real soon.

yash2798:
Hi @ChrisRackauckas  I can't quite get how can we get the gradient in line 421 without the whole jacobian? Did you mean something through jvp?

PR #172, opened 2023-03-24
Including jvp for radius update schemes

yash2798:


src/jacobian.jl
```diff
@@ -144,3 +144,14 @@ function jacobian_autodiff(f, x::AbstractArray, nonlinfun, alg)
                                 jac_prototype = jac_prototype, chunksize = chunk_size),
      num_of_chunks)
 end
+
+function jvp(cache::TrustRegionCache{false})
```
ChrisRackauckas, line 148: Put a `!` on there because it's mutating.

src/trustRegion.jl
```diff
@@ -417,17 +417,18 @@ function trust_region_step!(cache::TrustRegionCache)
...
         take_step!(cache)
         cache.loss = cache.loss_new
         cache.make_new_J = true
       else
         cache.make_new_J = false
       end
-      
+
+      @unpack p1= cache
+      cache.trust_r = p1 * cache.internalnorm(jvp(cache)) # we need the gradient at the new (k+1)th point  WILL THIS BECOME ALLOCATING?
```
ChrisRackauckas, line 430: The comment can be removed.

ChrisRackauckas:
Needs to add the dependency.

src/jacobian.jl
yash2798: we are actually not modifying the cache, we are just returning the jvp. Should that still be done?

src/NonlinearSolve.jl
```diff
@@ -61,6 +61,7 @@ SnoopPrecompile.@precompile_all_calls begin for T in (Float32, Float64)
 end end
 
 export RadiusUpdateSchemes
+export auto_jacvec, auto_jacvec!
```
ChrisRackauckas, line 64:
```suggestion
```

src/jacobian.jl
ChrisRackauckas: You are modifying the cache in one of them

src/trustRegion.jl
```diff
@@ -441,7 +445,7 @@ function dogleg!(cache::TrustRegionCache)
 
     # Test if the full step is within the trust region.
     if norm(u_tmp) ≤ trust_r
-        cache.step_size = u_tmp
+        cache.step_size = deepcopy(u_tmp)
```
ChrisRackauckas, line 448: Is this supposed to be a `.=`?

yash2798:
The issue is resolved, if there's nothing else, i think we can merge this pr.

ChrisRackauckas:
Test failed, looks like a tolerance issue.

yash2798:
works now

ChrisRackauckas (approved):

PR #178, opened 2023-04-08
Trust Region - Fan's method

yash2798:


yash2798:
Hi @ChrisRackauckas , line 307 and 309, is it alright if I calculate jacobian through direct `ForwardDiff.jacobian` calls? If Hei's scheme is used they'll be calculated only once in the cache `init` function

yash2798:
nvm i think jvp will work here too.

yash2798:
@ChrisRackauckas  if this seems alright, this can be merged.

ChrisRackauckas:
Next step, can we get a test that like 4 iterations of in-place and out-of-place match results?

yash2798:
> Next step, can we get a test that like 4 iterations of in-place and out-of-place match results?

Sure I will make a pr for this soon.

PR #179, opened 2023-04-19
iip vs oop iteration test

yash2798:


ChrisRackauckas:
I don't understand what you're doing here. Just run the code with maxiters = 1,2,3,4,5 and then check that the results are the same between the two modes?

yash2798:
> I don't understand what you're doing here. Just run the code with maxiters = 1,2,3,4,5 and then check that the results are the same between the two modes?

Yeah, but I was thinking maybe this would be beneficial when we benchmark different schemes later, like to see how different methods converge differently. So, instead of doing maxiters=1,2,3... again and again, we can just simply store the values? Does this make sense?

ChrisRackauckas:
No because this will increase the memory requirement of the method. We shouldn't add cruft to the method for the benchmarking/testing purpose here when we don't need to.

yash2798:
> No because this will increase the memory requirement of the method. We shouldn't add cruft to the method for the benchmarking/testing purpose here when we don't need to.

okay sure got it, i'll just add the test then

yash2798:
these tests pass. maybe this can be merged and i can add tests for more complex functions in the next PRs?

ChrisRackauckas (approved):

PR #191, opened 2023-06-20
Bastin's radius update scheme

yash2798:
The last pr got messed up a little so opening a new one. tests pass, formatted with no conflicts!

### DIFF
--- a/src/trustRegion.jl
+++ b/src/trustRegion.jl
@@ -249,7 +249,7 @@ function SciMLBase.__init(prob::NonlinearProblem{uType, iip}, alg::TrustRegion,
     else
         u = deepcopy(prob.u0)
     end
-    u_prev = deepcopy(u)
+    u_prev = zero(u)
     f = prob.f
     p = prob.p
     if iip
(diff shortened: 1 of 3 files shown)


src/trustRegion.jl
```diff
@@ -246,6 +249,7 @@ function SciMLBase.__init(prob::NonlinearProblem{uType, iip}, alg::TrustRegion,
     else
         u = deepcopy(prob.u0)
     end
+    u_prev = deepcopy(u)
```
ChrisRackauckas, line 252:
```suggestion
    u_prev = zero(u)
```

src/trustRegion.jl
```diff
@@ -254,6 +258,7 @@ function SciMLBase.__init(prob::NonlinearProblem{uType, iip}, alg::TrustRegion,
     else
         fu = f(u, p)
     end
+    fu_prev = deepcopy(fu)
```
ChrisRackauckas, line 261:
```suggestion
    fu_prev = zero(fu)
```

src/trustRegion.jl
ChrisRackauckas, line 252: This is only needed by some algorithms though, right? So in many cases it can be made as an empty array?

ChrisRackauckas:
This should be fine,  we can look into optimizing that copyto! operation later when we get to profiling.

PR #194, opened 2023-07-11
Line Search for Newton-Raphson

yash2798:


src/jacobian.jl
```diff
@@ -149,3 +149,19 @@ function jacobian_autodiff(f, x::AbstractArray, nonlinfun, alg)
             jac_prototype = jac_prototype, chunksize = chunk_size),
         num_of_chunks)
 end
+
+function simple_jacobian(cache, x::Number)
```
oscardssmith, line 153: shouldn't the nonlinear solve be responsible for providing the jacobian? What if the problem doesn't support autodiff?

src/raphson.jl
```diff
@@ -48,29 +49,33 @@ for large-scale and numerically-difficult nonlinear systems.
...
+mutable struct NewtonRaphsonCache{iip, fType, algType, uType, duType, resType, pType, foType, goType,
     INType, tolType,
     probType, ufType, L, jType, JC}
     f::fType
     alg::algType
     u::uType
     fu::resType
     p::pType
+    α::Real
+    f_o::foType # stores value of objective
```
oscardssmith, line 77: this should be restype

src/raphson.jl
```diff
@@ -181,7 +189,9 @@ end
 function perform_step!(cache::NewtonRaphsonCache{false})
     @unpack u, fu, f, p = cache
     J = jacobian(cache, f)
-    cache.u = u - J \ fu
+    cache.du1 = J \ fu
```
oscardssmith, line 192: Should this be using LinearSolve?

oscardssmith: sounds good

src/utils.jl
```diff
@@ -163,3 +163,9 @@ function rfunc(r::R, c2::R, M::R, γ1::R, γ2::R, β::R) where {R <: Real} # R-f
         return (1 - γ1 - β) * (exp(r - c2) + β / (1 - γ1 - β))
     end
 end
+
+function get_iip(prob::NonlinearProblem{uType, iip}) where {uType, iip}
+    iip
+end
```
avik-pal, line 169: This is probably just `SciMLBase.isinplace`

yash2798:
@ChrisRackauckas i did some preliminary tests and they seem to perform well but I have concerns regarding the efficiency (or "code performance") in general which I am highlighting.

src/raphson.jl
```diff
@@ -172,6 +182,43 @@ function perform_step!(cache::NewtonRaphsonCache{false})
     return nothing
 end
 
+function objective_linesearch(cache::NewtonRaphsonCache) ## returns the objective functions required in LS
```
yash2798, line 185: This function `objective_linesearch` returns 3 functions and is called in a loop (so it might get called multiple times). is this an efficient approach?

src/raphson.jl
```diff
@@ -172,6 +182,43 @@ function perform_step!(cache::NewtonRaphsonCache{false})
...
+    ϕ(α) = fo(u .- α .* du1)
+
+    function dϕ(α)
+        g!(g_o, u .- α .* du1)
+        return dot(g_o, -du1)
+    end
+
+    function ϕdϕ(α)
+        return (fg!(g_o, u .- α .* du1), dot(g_o, -du1))
+    end
```
yash2798, line 216: Similarly, these 3 functions are defined each time the `perform_linesearch!` function is called. Can this lead to performance issues?

src/jacobian.jl
ChrisRackauckas: This is already handled by other dispatches. Why does this exist?

src/raphson.jl
```diff
@@ -172,6 +182,44 @@ function perform_step!(cache::NewtonRaphsonCache{false})
...
 
+function objective_linesearch(cache::NewtonRaphsonCache) ## returns the objective functions required in LS
+    @unpack f, p = cache
+
+    function fo(x)
+        return dot(value_f(cache, x), value_f(cache, x)) / 2
+    end
+
+    function g!(g_o, x)
+        mul!(g_o, simple_jacobian(cache, x)', value_f(cache, x))
```
ChrisRackauckas, line 193: This is not a good way to compute this.

src/raphson.jl
```diff
@@ -172,6 +182,44 @@ function perform_step!(cache::NewtonRaphsonCache{false})
     return nothing
 end
 
+function objective_linesearch(cache::NewtonRaphsonCache) ## returns the objective functions required in LS
+    @unpack f, p = cache
+
+    function fo(x)
+        return dot(value_f(cache, x), value_f(cache, x)) / 2
```
ChrisRackauckas, line 189: Why is the same function called twice?

ChrisRackauckas:
@avik-pal I assume you'll need to incorporate this into https://github.com/SciML/NonlinearSolve.jl/pull/206

src/raphson.jl
ChrisRackauckas: This should be a vjp with reversemode ad?

src/jacobian.jl
yash2798: yes, however in the current dispatch, the jacobian is calculated at `cache.u`, i.e. the dispatch takes in `cache` as an argument and not the `x`. But in when linesearch is performed, we need the jacobian at multiple `x`'s (for multiple step lengths $\alpha_{k}$), so we can't just do `jacobian(f, cache)` as in the current dispatch. 

for example like here (https://github.com/JuliaNLSolvers/LineSearches.jl/blob/ded667a80f47886c77d67e8890f6adb127679ab4/src/strongwolfe.jl#L84)

i can't think of any other way to approach this, any inputs will be helpful?

yash2798: although i have an idea like this roughly
`function simple_jacobian(cache, x)
    cache.tmp_var = cache.u
    cache.u = x
    J = jacobian(cache, f) ## use our current dispatch
    cache.u = cache.tmp_var
end`

yash2798: this way we don't have to deal with sparsity and AD/Finite diff separately

ChrisRackauckas: Yes, it should be something like that reusing the existing machinery. But the vjp shouldn't build the Jacobian at all: you don't actually need to calculate the Jacobian for most of this.

yash2798: yes you are absolutely right. so we need $J^{T}v = (v^{T}J)^{T}$. We can calculate vjp $v^{T}J$. are there any routines in `SparseDiffTools.jl` or do we need to do this kind of 'in-situ'

yash2798: if in-situ, can you point me to something that i can take help from regarding reverse-mode AD, i am not very well versed with it.

ChrisRackauckas: Yes, that's precisely what Reverse-Mode AD calculates without calculating the Jacobian. This is why I want @avik-pal to take a look and see how that would mesh with https://github.com/SciML/NonlinearSolve.jl/pull/206

avik-pal:
Can be closed in favor of https://github.com/SciML/NonlinearSolve.jl/pull/203.

yash2798:
sorry i don't understand. Do we not need line search anymore? or do we do it after #203 gets merged?

avik-pal:
I incorporated the line search in that PR. It doesn't construct Jacobians and instead uses VJPs and also caches the function construction during the nonlinear solver cache construction.

PR #197, opened 2023-07-25
Trust Region docstrings improvement

yash2798:


yash2798:
@ChrisRackauckas  this can be merged

src/trustRegion.jl
```diff
@@ -88,6 +85,14 @@ for large-scale and numerically-difficult nonlinear systems.
     Currently, the linear solver and chunk size choice only applies to in-place defined
     `NonlinearProblem`s. That is expected to change in the future.
 """
+EnumX.@enumx RadiusUpdateSchemes begin
```
ChrisRackauckas, line 88: why was this moved? This puts the docstrong on the wrong object.

yash2798: Umm right now I dont see the trust region docstring on the main docs: https://docs.sciml.ai/NonlinearSolve/stable/api/nonlinearsolve/ . I thought it is because the docstring is not at the beginning.

yash2798: So should the Enum definition be on the main `src/NonlinearSolve.jl` file instead?

ChrisRackauckas: It's because the docstring wasn't right before the function. It has to be exactly the line before the function definition, no space.

ChrisRackauckas: It's already working on the dev version: https://docs.sciml.ai/NonlinearSolve/dev/api/nonlinearsolve/

PR #200, opened 2023-08-18
Newton Raphson used to ignore jac-prototype

avik-pal:


ChrisRackauckas:
Needs a test

avik-pal:
There was already a test, but it never tested that the `jac_prototype` was being used :sweat:. Updated that to check that the prototype is being mutated.

PR #202, opened 2023-08-29
Splitting the methods based on problem type for SimpleNonlinearSolve docs

yash2798:


yash2798:
can be merged @ChrisRackauckas

Issue #158, opened 2023-02-27
Solvers with KrylovJL fail


DanielVandH:
Should the code below be expected to work? 

```
using OrdinaryDiffEq, LinearSolve, Krylov, NonlinearSolve
function lorenz!(du,u,p,t)
    @show t
 du[1] = 10.0*(u[2]-u[1])
 du[2] = u[1]*(28.0-u[3]) - u[2]
 du[3] = u[1]*u[2] - (8/3)*u[3]
end
u0 = [1.0;0.0;0.0]
tspan = (0.0,100.0)
prob = ODEProblem(lorenz!,u0,tspan) |> NonlinearProblem
alg = NewtonRaphson(linsolve=KrylovJL_GMRES())
sol = solve(prob, alg)
```
```
ERROR: MethodError: no method matching forwarddiff_color_jacobian!(::Matrix{Float64}, ::NonlinearSolve.JacobianWrapper{NonlinearFunction{true,SciMLBase.AutoSpecialize,…}, SciMLBase.NullParameters}, ::Vector{Float64}, ::Nothing)

Some of the types have been truncated in the stacktrace for improved reading. To emit complete information
in the stack trace, evaluate `TruncatedStacktraces.VERBOSE[] = true` and re-run the code.

  [6] #solve_call#22
    @ C:\Users\User\.julia\packages\DiffEqBase\JH4gt\src\solve.jl:509 [inlined]
  [7] solve_call
    @ C:\Users\User\.julia\packages\DiffEqBase\JH4gt\src\solve.jl:479 [inlined]
  [8] #solve_up#29    @ C:\Users\User\.julia\packages\DiffEqBase\JH4gt\src\solve.jl:940 [inlined]  [9] solve_up    @ C:\Users\User\.julia\packages\DiffEqBase\JH4gt\src\solve.jl:905 [inlined]
 [10] #solve#28
[...]

Issue #140, opened 2023-01-31
Specialize Jacobians when doing Krylov methods to avoid building the matrices


ChrisRackauckas:
Just doing the AD tricks for Newton-Krylov. Should probably wait for SciMLOperators.jl integration.

ChrisRackauckas:
Fixed by https://github.com/SciML/NonlinearSolve.jl/pull/203

Issue #63, opened 2022-04-28
Maxiters seems to be offset by 1


KronosTheLate:
Check out the following example:
```
julia> using NonlinearSolve, LinearAlgebra

julia> f(u,p) = u .* u .- p;

julia> u0 = [1.0, 1.0];

julia> p = 2.0;

julia> probN = NonlinearProblem{false}(f, u0, p);

julia> bla0 = solve(probN, NewtonRaphson(), maxiters = 0).u;

julia> bla1 = solve(probN, NewtonRaphson(), maxiters = 1).u;

julia> bla2 = solve(probN, NewtonRaphson(), maxiters = 2).u;

julia> norm(f(u0, p))
1.4142135623730951

julia> norm(f(bla0, p))
1.4142135623730951

julia> norm(f(bla1, p))
1.4142135623730951

julia> norm(f(bla2, p))
0.3535533905932738
```

It seems as if setting maxter to 1 acutally results in zero iterations occuring. And I am left to wonder if maxter=2 then means 2 iterations, or actually 1 iteration...

This is the same for my actual application, which makes me more certain that the issue is not specific to this problem. The other example is:
```
julia> using NonlinearSolve, LinearAlgebra

julia> v = 120;  #kg

julia> k = 2.5;  #m

julia> w = 4.0;  #km/m

julia> α = 2e-7; #kg⁻¹

julia> function F(x⃗, parameters)
           L₀, L, p, x, θ, φ, a, H = x⃗
           (;d, n) = parameters
           return [
           a*(cosh(x/a)-1) - p,
           2a * sinh(x/a) - L ,
           2x + 2k*cos(θ) - d,
           p+k*sin(θ) - n,
           sinh(x/a) - tan(φ),
           (1+v/(w*L₀)) * tan(φ) - tan(θ),
           L₀ * (1 + α*H) - L,
           w*L₀ / 2sin(φ) - H,
           ]
       end;

julia> params = (d=30, n=5);

julia> u0 =    [29, 27, 1, params.d/2, deg2rad(45),  deg2rad(22.5), 40, 500];

julia> probN = NonlinearProblem{false}(F, u0, params);

julia> bla0 = solve(probN, NewtonRaphson(), maxiters = 0).u;

julia> bla1 = solve(probN, NewtonRaphson(), maxiters = 1).u;

julia> bla2 = solve(probN, NewtonRaphson(), maxiters = 2).u;

julia> norm(F(u0, params))
348.4941911206226

[...]

PR #203, opened 2023-09-07
Towards a cleaner and more maintainable internals of NonlinearSolve.jl

### LINKED ISSUE #195: Inplace Problem Specification for Methods with `size(residual) != size(u)`
Currently, for inplace problems the residual vector is determined by the size of the input `u`. This is not necessarily correct for problems that support overconstrained / underconstrained setups. For example:
https://github.com/SciML/NonlinearSolve.jl/blob/691e10a25c0b20b3349770a4d54f93b41e5235a4/src/levenberg.jl#L269-L270

Could we potentially add a kwarg like `residual_size` which could be used to initialize this?

avik-pal:
Fixes #195 #158 #140 #63 

* No special jacobian logic is now present in this package
  * Delegates everything to SparseDiffTools
* Uses `ADTypes` -- constructs the ADType from the old argument style!

## TODOs:

- [x] Fix Scalar
- [x] Solvers
  - [x] Raphson
  - [x] Trust Region
  - [x] Levenberg
- [x] Make tests pass again
- [x] Verify that all allocations are necessary
  - [x] Raphson
  - [x] Trust Region
  - [x] Levenberg
- [x] Compat Entries
- [x] [upstream] Add `resid_prototype` to properly construct `fu` (https://github.com/SciML/SciMLBase.jl/pull/490)
- [x] [upstream] Add a precomputed SparsityDetection type in SparseDiffTools.jl (https://github.com/JuliaDiff/SparseDiffTools.jl/pull/259)
- [x] Forward Mode for Abstract Array (was broken before as well)

**Holding Off Merge till we have some upstream benchmarks**


ChrisRackauckas:
Can we setup a PDE benchmark before merging this so we know that the change is not a performance regression? It shouldn't be, but it would be good to know that the new sparse diff stuff is on par in performance with the setup we had before.

src/ad.jl
```diff
@@ -23,22 +23,18 @@ function scalar_nlsolve_ad(prob, alg, args...; kwargs...)
     return sol, partials
```
avik-pal, line 23: @ChrisRackauckas, what does this code (mathematically) exactly do? I have never messed around with ForwardDiff so not really sure.

The code doesn't work if $u$ is an SVector. The previous tests had a typo and never tested that case.

ChrisRackauckas (approved):

src/ad.jl
ChrisRackauckas: It's handling the derivative rule for nonlinear solve, just skipping the dual numbers in the nonlinear solve and then applying the partials at the solution point.

avik-pal: Done!

src/ad.jl
```diff
@@ -1,40 +1,61 @@
...
+    f_p = scalar_nlsolve_∂f_∂p(f, uu, p)
+    f_x = scalar_nlsolve_∂f_∂u(f, uu, p)
+
+    z_arr = -inv(f_x) * f_p
 
-    f_x = ForwardDiff.derivative(Base.Fix2(f, p), uu)
     pp = prob.p
-    sumfun = let f_x′ = -f_x
-        ((fp, p),) -> (fp / f_x′) * ForwardDiff.partials(p)
+    sumfun = ((z, p),) -> [zᵢ * ForwardDiff.partials(p) for zᵢ in z]
```
ChrisRackauckas, line 16: This allocates?

avik-pal: Now it doesn't for scalars (both u and p). If either isn't a scalar, it will allocate

ChrisRackauckas: and static arrays?

src/linesearch.jl
```diff
@@ -0,0 +1,156 @@
...
+    eval_f(u) = evaluate_f(f, u, p, IIP; fu)
+
+    ls.method isa Static && return LineSearchCache(eval_f, nothing, nothing, nothing,
+        convert(eltype(u), ls.α), ls)
+
+    g₀ = _mutable_zero(u)
+
+    autodiff = if iip && (ls.autodiff isa AutoZygote || ls.autodiff isa AutoSparseZygote)
+        @warn "Attempting to use Zygote.jl for linesearch on an in-place problem. Falling back to finite differencing."
+        AutoFiniteDiff()
```
ChrisRackauckas, line 96: why not fallback to ForwardDiff if it's being used elsewhere?

avik-pal: I wanted to use `VecJac`, which doesn't have the forwarddiff yet (I think?). See https://github.com/JuliaDiff/SparseDiffTools.jl/blob/04810f5caea10633504ad44071f1a6b82cb9bb5a/src/differentiation/vecjac_products.jl#L51-L52

I can fix it upstream to just construct jacobian and then multiply, and then change the default here later

avik-pal:
Needs https://github.com/SciML/DiffEqBase.jl/pull/926

PR #207, opened 2023-09-13
Add Gauss Newton and make LM work for NLS Problems

avik-pal:
Don't merge it before https://github.com/SciML/NonlinearSolve.jl/pull/203


ChrisRackauckas:
Should LeastSquaresProblem be its own problem type?

avik-pal:
> Should LeastSquaresProblem be its own problem type?

That's actually a good point. I was thinking about this from a correctness standpoint. Currently if using a Nonlinear Solver for NLS, we will directly hit SteadyStateAdjoint which is clearly incorrect.

ChrisRackauckas:
Also, a LeastSquaresProblem could use other methods like Optimization.jl stuff. So it just seems like a prudent way to handle it.

We should make Levenburg-Marquedt specialize on it too. Worth an issue, etc.

avik-pal:
Damn it. Messed up the history. I will open a new PR

PR #208, opened 2023-09-13
Start using termination conditions from DiffEqBase

utkarsh530:
Requires #203


utkarsh530:
Based on: https://github.com/SciML/SimpleNonlinearSolve.jl/pull/45

Should we keep the existing interface, i.e., providing options to specify `abstol` and `reltol` from the `solve`? Currently the handling is similar to that in `Broyden` in `SimpleNonlinearSolve`.

ChrisRackauckas:
> Should we keep the existing interface, i.e., providing options to specify abstol and reltol from the solve? 

Yes, for simplicity and keeping the same interface. If `abstol` or `reltol` are used, that should specify the termination condition. If the termination condition is used, then that allows for a much more detailed specification. If both are specified, then error.

utkarsh530:
> If both are specified, then error.

I based this check on the SimpleNonlinearSolve.jl behavior:

https://github.com/SciML/SimpleNonlinearSolve.jl/blob/main/src/batched/utils.jl#L17-L20

ChrisRackauckas:
@avik-pal can you work with @utkarsh530 to rebase this and get it in? Are the termination conditions all uniform after this?

src/levenberg.jl
```diff
@@ -176,15 +186,29 @@ function SciMLBase.__init(prob::Union{NonlinearProblem{uType, iip},
...
 
-    return LevenbergMarquardtCache{iip}(f, alg, u, fu1, fu2, du, p, uf, linsolve, J,
-        jac_cache, false, maxiters, internalnorm, ReturnCode.Default, abstol, prob, DᵀD,
+    tc = alg.termination_condition
+    mode = DiffEqBase.get_termination_mode(tc)
+
+    atol = _get_tolerance(abstol, tc.abstol, eltype(u))
+    rtol = _get_tolerance(reltol, tc.reltol, eltype(u))
+
+    storage = mode ∈ DiffEqBase.SAFE_TERMINATION_MODES ? NLSolveSafeTerminationResult() :
```
ChrisRackauckas, line 195: Is this type stable?

utkarsh530: I believe so; the types are inferred in the cache struct. Something similar to want we did here: https://github.com/SciML/SimpleNonlinearSolve.jl/blob/21f14ff8e1bca2e20ee0de9158050a154ce3d585/src/broyden.jl#L53

ChrisRackauckas:
I think this is just about done, but we need to discuss a few last interface issues:

1. Usually general solver controls go into `solve`. We should make this a solve keyword argument?
2. We should make sure `abstol` and `reltol` lower in clearly documented ways that match how it's done in SteadyStateDiffEq.jl, i.e. if `abstol` and/or `reltol` is given, it defines a `TerminationCondition` that is sensible, and if `abstol` and/or `reltol` is given but also a `TerminationCondition` is given, it should check compatibility or error that incompatible termination condition combinations were seen.

What this does is effectively allow for the "simple" interface of `abstol`/`reltol` that everyone knows and loves to carry over to this domain, while allowing a much more detailed interface to control convergence if desired. However, we need to make sure it's fully uniform before adopting it.

@avik-pal can you chime in.

avik-pal:
Yeah I agree with the suggestions. The original implementation lived inside DEQs.jl so it was't fleshed out carefully.

ChrisRackauckas:
Is this done?

src/gaussnewton.jl
```diff
@@ -36,26 +36,34 @@ for large-scale and numerically-difficult nonlinear least squares problems.
...
     return GaussNewton{CJ}(ad, alg.linsolve, alg.precs)
 end
 
-function GaussNewton(; concrete_jac = nothing, linsolve = CholeskyFactorization(),
-    precs = DEFAULT_PRECS, adkwargs...)
+function GaussNewton(; concrete_jac = nothing, linsolve = NormalCholeskyFactorization(),
+    precs = DEFAULT_PRECS,
+    termination_condition = NLSolveTerminationCondition(NLSolveTerminationMode.AbsNorm;
+        abstol = nothing,
+        reltol = nothing), adkwargs...)
```
avik-pal, line 54: I think we decided not to go down this route right? We want to pass the everything to the solve / init and directly do it there.

avik-pal (changes requested):

utkarsh530:
No, not done yet

utkarsh530:
@ChrisRackauckas @avik-pal  this should be good to merge now.

ChrisRackauckas (approved):

Issue #209, opened 2023-03-21
Update README and docs tutorial examples to use new bracketing methods?


sjdaines:
Currently the examples in [README.md ](https://github.com/SciML/SimpleNonlinearSolve.jl#readme)and https://docs.sciml.ai/NonlinearSolve/stable/tutorials/nonlinear/#Using-Bracketing-Methods 
still use the `Falsi` method which is not a good new user experience.

Suggest a better choice would be `Brent` ?

ChrisRackauckas:
Fixed

PR #211, opened 2023-09-15
Tutorial change for Bracketing method

yash2798:
changed the tutorial method to itp [#209]


yash2798:
@ChrisRackauckas is something else required here?

ChrisRackauckas (approved):

PR #214, opened 2023-09-18
WIP: Add DFSane method

axla-io:
This PR adds a DFSane solver, similar to the ones in SimpleNonlinearSolve, [here](https://github.com/SciML/SimpleNonlinearSolve.jl/blob/main/src/dfsane.jl) and [here](https://github.com/SciML/SimpleNonlinearSolve.jl/blob/main/src/batched/dfsane.jl).

The implementation in this PR improves on the SimpleNonlinearSolve version by adding a cached solver with non allocating iterations. 

Checklist:

- [x] Algorithm and cache
- [x] Init function
- [x] In place steps
- [x] Bounds check for spectral parameter
- [x] Out of place steps
- [x] Benchmark against SimpleNonlinearSolve
- [x] Add tests
- [x] Docstrings


src/dfsane.jl
```diff
@@ -0,0 +1,314 @@
...
+    nₑₓₚ = alg.nₑₓₚ
+    𝒹, uₙ₋₁, fuₙ, fuₙ₋₁ = copy(uₙ), copy(uₙ), copy(uₙ), copy(uₙ)
+
+    if iip
+        f(dx, x) = prob.f(dx, x, p)
+        f(fuₙ₋₁, uₙ₋₁)
+
+    else
+        f(x) = prob.f(x, p)
+        fuₙ₋₁ = f(uₙ₋₁)
```
avik-pal, line 118: The error is because `f` is being overwritten due to branching. Changing the name might fix it.

axla-io: Changing the name does sort of fix it. But I'm sorry, I really don't understand why this happens. I never had this problem before. Do you have some quick reference?

ChrisRackauckas: You just cannot define standard functions in branches, make them anonymous.

src/dfsane.jl
```diff
@@ -0,0 +1,314 @@
...
+    M = alg.M
+    nₑₓₚ = alg.nₑₓₚ
+    𝒹, uₙ₋₁, fuₙ, fuₙ₋₁ = copy(uₙ), copy(uₙ), copy(uₙ), copy(uₙ)
+
+    if iip
+        f(dx, x) = prob.f(dx, x, p)
+        f(fuₙ₋₁, uₙ₋₁)
+
+    else
+        f(x) = prob.f(x, p)
```
ChrisRackauckas, line 117:
```suggestion
        f = (x) -> prob.f(x, p)
```

src/dfsane.jl
```diff
@@ -0,0 +1,314 @@
...
+    α₁ = one(T)
+    γ = T(alg.γ)
+    f₍ₙₒᵣₘ₎ₙ₋₁ = α₁
+    σₙ = T(alg.σ₁)
+    M = alg.M
+    nₑₓₚ = alg.nₑₓₚ
+    𝒹, uₙ₋₁, fuₙ, fuₙ₋₁ = copy(uₙ), copy(uₙ), copy(uₙ), copy(uₙ)
+
+    if iip
+        f(dx, x) = prob.f(dx, x, p)
```
ChrisRackauckas, line 113:
```suggestion
        f( = dx, x) -> prob.f(dx, x, p)
```

src/dfsane.jl
```diff
@@ -0,0 +1,314 @@
...
+    α₁ = one(T)
+    γ = T(alg.γ)
+    f₍ₙₒᵣₘ₎ₙ₋₁ = α₁
+    σₙ = T(alg.σ₁)
+    M = alg.M
+    nₑₓₚ = alg.nₑₓₚ
+    𝒹, uₙ₋₁, fuₙ, fuₙ₋₁ = copy(uₙ), copy(uₙ), copy(uₙ), copy(uₙ)
+
+    if iip
+        f( = dx, x) -> prob.f(dx, x, p)
```
ChrisRackauckas, line 113:
```suggestion
        f =  (dx, x) -> prob.f(dx, x, p)
```

src/dfsane.jl
axla-io: Thanks!

src/dfsane.jl
```diff
@@ -0,0 +1,313 @@
...
+    @. cache.uₙ₋₁ = cache.uₙ₋₁ * cache.fuₙ₋₁
+    α₋ = sum(cache.uₙ₋₁)
+    cache.σₙ = α₊ / α₋
+
+    # Spectral parameter bounds check
+    if abs(cache.σₙ) > σₘₐₓ || abs(cache.σₙ) < σₘᵢₙ
+        test_norm = sqrt(sum(abs2, cache.fuₙ₋₁))
+        if test_norm > 1
+            cache.σₙ = 1.0
+        elseif testnorm < 1e-5
```
avik-pal, line 281:
```suggestion
        elseif test_norm < 1e-5
```

src/dfsane.jl
```diff
@@ -0,0 +1,313 @@
...
+            cache.σₙ = 1e5
+        else
+            cache.σₙ = 1.0 / test_norm
```
avik-pal, line 284: is this not just a `clamp` op?

axla-io: Yes, you're right haha!

axla-io: Hmm, I don't think so. We are testing on the norm of `cache.fuₙ₋₁` to update `cache.σₙ`. Or maybe you are seeing something I'm not?

avik-pal: seems like  a clamp on `1/test_norm` to be between 1 and 1e5

axla-io: AH! Now I get what you're saying. Like: `cache.σₙ = clamp(1.0 / test_norm, 1, 1e5)`
Yes that makes sense.

avik-pal:
How close are we from getting this in?

src/dfsane.jl
```diff
@@ -0,0 +1,346 @@
...
+                                            σₘₐₓ,
+                                            σ₁,
+                                            M,
+                                            γ,
+                                            τₘᵢₙ,
+                                            τₘₐₓ,
+                                            nₑₓₚ,
+                                            ηₛ,
+                                            max_inner_iterations)
+end
```
avik-pal, line 80: Let's rewrite this to not use unicode. The Batched version was an internal implementation detail never meant to be exposed to the user, so it was fine there.

avik-pal (changes requested):

src/dfsane.jl
axla-io: Sure! Didn't know about this distinction before

axla-io:
Tests are added! Everything works except for that ForwardDiff fails in some cases, see this MWE:

```julia
using NonlinearSolve
using FiniteDiff, ForwardDiff

quadratic_f(u, p) = u .* u .- p

function benchmark_nlsolve_oop(f, u0, p=2.0)
    prob = NonlinearProblem{false}(f, u0, p)
    return solve(prob, DFSane(), abstol=1e-9)
end

broken_forwarddiff = [3.0, 4.0, 81.0]
for p in broken_forwarddiff
    analytical_derivative = 1 / (2 * sqrt(p))
    forward_diff = abs(ForwardDiff.derivative(p -> benchmark_nlsolve_oop(quadratic_f, 1.0, p).u, p))
    finite_diff = abs(FiniteDiff.finite_difference_derivative(p -> benchmark_nlsolve_oop(quadratic_f, 1.0, p).u, p))
    println("p = $p, Analytical: $analytical_derivative, ForwardDiff: $forward_diff, FiniteDiff: $finite_diff")
end
```

Which prints out:
```
p = 3.0, Analytical: 0.2886751345948129, ForwardDiff: 1776.530469223857, FiniteDiff: 0.2886751347781091
p = 4.0, Analytical: 0.25, ForwardDiff: 1.0, FiniteDiff: 0.25000000015714613
p = 81.0, Analytical: 0.05555555555555555, ForwardDiff: 0.1, FiniteDiff: 0.05555555555505331
```

ChrisRackauckas (approved):

ChrisRackauckas:
Specifically https://github.com/SciML/NonlinearSolve.jl/issues/245

PR #210, opened 2023-09-14
TrustRegion bugs

FHoltorf:
I had a look at https://github.com/SciML/NonlinearSolve.jl/issues/142. Indeed there were some mistakes in how the Gauss-Newton step and Cauchy point were calculated. This is still WIP, however, since TrustRegion is still not up to par with nlsolve's implementation after these changes. 

I suspect a large issue for the poor performance is the default TR adaptation and step acceptance/rejection routine; for instance the current default is rather conservative with respect to which steps are being accepted. Lowering the `step_threshold` default to 1e-4 from 0.1 (and after correcting the GN/Cauchy step computations) the performance on the test set is already quite a bit better.

Before
```
nr  Problem                                            n     NewtonRaphson       TrustRegion         LevenbergMarquardt  nlsolve
1   Generalized Rosenbrock function                    10    3.9833e-311         4.9193              4.9193              0.0
2   Powell singular function                           4     7.0435e-16          7.941               0.021633            7.0217e-16
3   Powell badly scaled function                       2     0.0                 26897.0             2.1983e-14          0.0
4   Wood function                                      4     4.1832e-14          19081.0             8550.5              8.2157e-15
5   Helical valley function                            3     6.2775e-30          50.0                50.0                4.3667e-26
6   Watson function                                    2     18860.0             3004.4              129.15              114.89
7   Chebyquad function                                 2     5.5511e-17          0.075044            0.34127             9.4369e-16
8   Brown almost linear function                       10    0.0                 0.01533             16.53               4.4409e-16
9   Discrete boundary value function                   10    5.4548e-16          0.0047026           6.5869e-15          2.7756e-16
10  Discrete integral equation function                10    2.7807e-15          1.9166e-15          9.3334e-15          5.5511e-17
11  Trigonometric function                             10    1.4753e-15          2.2808e-15          0.084117            6.8695e-16
12  Variably dimensioned function                      10    0.0                 2.8636e105          1.376e-12           0.0
13  Broyden tridiagonal function                       10    1.2803e-15          2.7487e-15          1.1633e-13          6.6613e-16
14  Broyden banded function                            10    1.4002e-15          2.1729e-15          1.6865e-13          5.5511e-16
15  Hammarling 2 by 2 matrix square root problem       4     2.2204e-16          0.32355             0.91487             2.2204e-16
16  Hammarling 3 by 3 matrix square root problem       9     2.2204e-16          0.30203             1.0506              2.2204e-16
[...]

PR #222, opened 2023-09-27
Test problem library for regression tests

FHoltorf:
My interpretation of @avik-pal's proposal (https://github.com/SciML/NonlinearSolve.jl/pull/210#issuecomment-1737518425) for how to include the test problem library as part of the tests. If someone has an idea for a more elegant way to flag which tests are allowed to fail (i.e., problems for which we do not expect a given solver to converge), that would be very much appreciated.  Here it needs to be done by hand.


test/testproblems.jl
```diff
@@ -0,0 +1,570 @@
+using NonlinearSolve, LinearAlgebra
```
avik-pal, line 1: We can just use the NonlinearProblemLibrary here?

avik-pal:
I pulled the changes here into https://github.com/SciML/NonlinearSolve.jl/pull/226

PR #226, opened 2023-10-03
Fix bug in Levenberg

avik-pal:
## Updated Version

```julia
nr  Problem                                            n     NewtonRaphson       TrustRegion         LevenbergMarquardt  LevenberdMarquardt (LSOptim)
1   Generalized Rosenbrock function                    10    NaN                 0.0                 1.0214e-14          5.8944e-11          
2   Powell singular function                           4     7.0435e-16          6.8342e-16          1.0538e-15          2.3573e-7           
3   Powell badly scaled function                       2     0.0                 0.0                 8.7431e-5           2.5957e-11          
4   Wood function                                      4     1.8586e-14          1.0663e-14          1.2104e-14          3.8075e-12          
5   Helical valley function                            3     6.2775e-30          9.4526e-23          5.2326e-19          6.8803e-13          
6   Watson function                                    2     NaN                 NaN                 NaN                 NaN                 
7   Chebyquad function                                 2     5.5511e-17          1.6653e-16          2.0053e-15          4.4854e-12          
8   Brown almost linear function                       10    0.0                 6.6613e-16          4.2061e-14          2.1603e-11          
9   Discrete boundary value function                   10    5.4548e-16          6.9388e-17          3.1544e-15          3.7241e-11          
10  Discrete integral equation function                10    2.7807e-15          3.4694e-17          3.1457e-15          1.2691e-11          
11  Trigonometric function                             10    1.8884e-15          0.0052868           0.0052874           0.0052868           
12  Variably dimensioned function                      10    0.0                 0.0                 5.124e-12           1.345e-13           
13  Broyden tridiagonal function                       10    1.2803e-15          1.3414e-15          6.4287e-15          1.5726e-12          
14  Broyden banded function                            10    1.4002e-15          1.4002e-15          6.37e-15            1.4637e-13          
15  Hammarling 2 by 2 matrix square root problem       4     1.1102e-16          2.4973e-6           7.3386e-6           2.9137e-6           
16  Hammarling 3 by 3 matrix square root problem       9     1.1102e-16          2.1418e-6           7.3386e-6           2.9119e-6           
17  Dennis and Schnabel 2 by 2 example                 2     0.0                 0.0                 13.143              3.7942e-15          
18  Sample problem 18                                  2     0.0                 5.2861e-18          2.8025e-10          7.6511e-11          
19  Sample problem 19                                  2     9.5092e-16          9.5092e-16          7.0792e-6           1.058e-6            
[...]

PR #229, opened 2023-10-06
Fix Jacobian Construction for concrete_jac

avik-pal:
Needs #226


avik-pal:
Needs https://github.com/JuliaDiff/SparseDiffTools.jl/pull/263

PR #213, opened 2023-09-16
Add advanced tutorials

ErikQQY:
What is the correct way of adding a preconditioner for linear solvers in NonlinearSolve.jl? I wrote this tutorial based on [solving large stiff ODE in diffeq docs](https://docs.sciml.ai/DiffEqDocs/stable/tutorials/advanced_ode_example/). I think the preconditioners should be the same since we have a common LinearSolve.jl interface, but the preconditioners failed when I passed them to `solve`.

MWE:
```julia
using NonlinearSolve, LinearAlgebra, SparseArrays, LinearSolve

const N = 32
const xyd_brusselator = range(0, stop = 1, length = N)
brusselator_f(x, y) = (((x - 0.3)^2 + (y - 0.6)^2) <= 0.1^2) * 5.0
limit(a, N) = a == N + 1 ? 1 : a == 0 ? N : a
function brusselator_2d_loop(du, u, p)
    A, B, alpha, dx = p
    alpha = alpha / dx^2
    @inbounds for I in CartesianIndices((N, N))
        i, j = Tuple(I)
        x, y = xyd_brusselator[I[1]], xyd_brusselator[I[2]]
        ip1, im1, jp1, jm1 = limit(i + 1, N), limit(i - 1, N), limit(j + 1, N),
        limit(j - 1, N)
        du[i, j, 1] = alpha * (u[im1, j, 1] + u[ip1, j, 1] + u[i, jp1, 1] + u[i, jm1, 1] -
                       4u[i, j, 1]) +
                      B + u[i, j, 1]^2 * u[i, j, 2] - (A + 1) * u[i, j, 1] +
                      brusselator_f(x, y)
        du[i, j, 2] = alpha * (u[im1, j, 2] + u[ip1, j, 2] + u[i, jp1, 2] + u[i, jm1, 2] -
                       4u[i, j, 2]) +
                      A * u[i, j, 1] - u[i, j, 1]^2 * u[i, j, 2]
    end
end
p = (3.4, 1.0, 10.0, step(xyd_brusselator))

function init_brusselator_2d(xyd)
    N = length(xyd)
    u = zeros(N, N, 2)
    for I in CartesianIndices((N, N))
        x = xyd[I[1]]
        y = xyd[I[2]]
        u[I, 1] = 22 * (y * (1 - y))^(3 / 2)
        u[I, 2] = 27 * (x * (1 - x))^(3 / 2)
    end
    u
end
u0 = init_brusselator_2d(xyd_brusselator)
[...]

PR #231, opened 2023-10-08
Add docs of AutoSparseForwardDiff and matrix coloring

ErikQQY:
Continue #213

### DIFF
--- a/docs/src/tutorials/advanced.md
+++ b/docs/src/tutorials/advanced.md
@@ -55,7 +55,7 @@ are then applied at each point in space (they are broadcast). Use `dx=dy=1/32`.
 The resulting `NonlinearProblem` definition is:
 
 ```@example ill_conditioned_nlprob
-using NonlinearSolve, LinearAlgebra, SparseArrays, LinearSolve
+using NonlinearSolve, LinearAlgebra, SparseArrays, LinearSolve, Symbolics
 
 const N = 32
 const xyd_brusselator = range(0, stop = 1, length = N)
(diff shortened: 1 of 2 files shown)


docs/src/tutorials/advanced.md
```diff
@@ -275,3 +275,28 @@ nothing # hide
...
+        autodiff=AutoSparseForwardDiff()));
+nothing # hide
+```
+
+To setup matrix coloring for the jacobian sparsity pattern, we can simply get the coloring vector by using [ArrayInterface.jl](https://github.com/JuliaArrays/ArrayInterface.jl) for the sparsity pattern of `jac_prototype`:
+
+```@example ill_conditioned_nlprob
+using ArrayInterface
+colorvec = ArrayInterface.matrix_colors(jac_sparsity)
+ff = NonlinearFunction(brusselator_2d_loop; jac_prototype=float.(jac_sparsity), colorvec)
```
ChrisRackauckas, line 295: Aren't the color vectors done automatically so this step is unneccessary?

ChrisRackauckas, line 295: Aren't the color vectors done automatically so this step is unneccessary?

docs/src/tutorials/advanced.md
```diff
@@ -275,3 +275,28 @@ nothing # hide
 
 For more information on the preconditioner interface, see the
 [linear solver documentation](https://docs.sciml.ai/LinearSolve/stable/basics/Preconditioners/).
+
+## Speed up Jacobian computation with sparsity exploitation and matrix coloring
```
ChrisRackauckas, line 279: This seems a bit oddly placed since the parts above already did sparsity and matrix coloring. This is just a less manual route for it

ErikQQY: Avik said the above part https://docs.sciml.ai/NonlinearSolve/dev/tutorials/advanced/#Declaring-a-Sparse-Jacobian-with-Automatic-Sparsity-Detection didn't actually use sparsity, we still need to use sparse AD type for matrix coloring. Did I understand this wrong?

avik-pal: @ErikQQY is right. Right now even if you have colorvecs and stuff but use a non-sparse AD type, we construct a dense jacobian based on how SparseDiffTools is setup. But this can be modified.

ChrisRackauckas: Oh. We should change that. If you pass a sparse matrix into jac_prototype it should always use sparse diff on it. The SparseDiffTools stuff should only override it to force sparse in the case of `jac_prototype=nothing`, otherwise it should respect the user's type and color on-demand (since coloring is super cheap).

PR #234, opened 2023-10-10
remove iip function parameter

simonbyrne:
Fixes warning
```
┌ NonlinearSolve [8913a72c-1f9b-4ce2-8d82-65094dcecaec]
│  WARNING: method definition for #solve#52 at /central/scratch/esm/slurm-buildkite/climacore-ci/2648/depot/default/packages/NonlinearSolve/lOnjn/src/ad.jl:36 declares type variable iip but does not use it.
└  
```


ChrisRackauckas:
Same as https://github.com/SciML/NonlinearSolve.jl/pull/233

PR #236, opened 2023-10-14
Impoving NLS Solvers

avik-pal:
- [x] Add a wrapper over LeastSquaresOptim
- [x] Sparse `mul!(A, J', J)` seems slower than `A .= J' * J` (@ChrisRackauckas any idea why?)
- [ ] Add tests


ChrisRackauckas:
> A .= J' * J

What's this actually used for? Generally that's a bad idea

avik-pal:
For GaussNewton and LM if it is NLLS. For a NLProblem we can get around by dropping the $J^T$ term on both sides.

avik-pal:
I did that dispatch for the sparse arrays case, because the in place hits the generic matmul code

avik-pal:
For Matrices it seems to hit the correct BLAS dispatches of `syrk`, but doesn't do it for Banded Matrices or Sparse Matrices (there seems to be a version in Intel MKL docs but don't know how to access it)

avik-pal:
Let's hold off a merge on this. We should be specializing the `J'J` matmul to use symmetric factorizations

avik-pal:
Should be good to go!

PR #228, opened 2023-10-04
[WIP] Pseudo-Transient Method

yonatanwesen:
This PR adds pseudo-transient  methods using adaptive step size described [here](https://ctk.math.ncsu.edu/PAPERS/Kelley_Liao_2013.pdf) created this new PR since #215 had too many merge conflicts

PR #251, opened 2023-10-19
Pseudo-Transient Method

yonatanwesen:
This PR adds pseudo-transient  methods using adaptive step size described [here](https://ctk.math.ncsu.edu/PAPERS/Kelley_Liao_2013.pdf) created this new PR since #228 had too many merge conflicts when rebasing

### DIFF
--- a/src/pseudotransient.jl
+++ b/src/pseudotransient.jl
@@ -1,0 +1,180 @@
+"""
+PseudoTransient{nothing, AutoForwardDiff{0, Bool}, Nothing, typeof(NonlinearSolve.DEFAULT_PRECS), Float64}(AutoForwardDiff{0, Bool}(true), 
+nothing, NonlinearSolve.DEFAULT_PRECS, 0.001)
+
+An implementation of PseudoTransient method that is used to solve steady state problems in an accelerated manner. It uses an adaptive time-stepping to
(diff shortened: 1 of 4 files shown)


src/pseudotransient.jl
```diff
@@ -0,0 +1,180 @@
+"""
+PseudoTransient{nothing, AutoForwardDiff{0, Bool}, Nothing, typeof(NonlinearSolve.DEFAULT_PRECS), Float64}(AutoForwardDiff{0, Bool}(true), 
+nothing, NonlinearSolve.DEFAULT_PRECS, 0.001)
+
+An implementation of PseudoTransient method that is used to solve steady state problems in an accelerated manner. It uses an adaptive time-stepping to
```
ChrisRackauckas, line 5:
```suggestion
    PseudoTransient(; concrete_jac = nothing, linsolve = nothing,
        precs = DEFAULT_PRECS, alpha_initial = 1e-3, adkwargs...)

An implementation of PseudoTransient method that is used to solve steady state problems in an accelerated manner. It uses an adaptive time-stepping to
```

ChrisRackauckas:
After this, adding alternative choices for alpha scheduling should be done. But we should merge this with just the simple alpha schedule.

PR #257, opened 2023-10-20
Porting over some low cost Algorithms

avik-pal:
- [x] Broyden
  - [x] Inplace
  - [x] OOP
  - [x] Integrate into the Test Suite
    - [x] Basic Tests
    - [x] 23 Test Problems
- [x] Klement
  - [x] Inplace
  - [x] OOP
  - [x] Integrate into the Test Suite
    - [x] Basic Tests
    - [x] 23 Test Problems

PR #258, opened 2023-10-23
Gauss Newton & LM Robustness Fixes

avik-pal:
Merge after #257 

For `GaussNewton` defaults to `QRFactorization(ColumnNorm())`, to ensure we can directly solve for
$$J \delta x = -f(x)$$

For `LM` we solve

$$ \begin{bmatrix}
    J\\
    \lambda D^TD
\end{bmatrix} v = \begin{bmatrix}
    -f(x)\\
    0
\end{bmatrix}  $$


src/gaussnewton.jl
```diff
@@ -81,15 +81,31 @@ function SciMLBase.__init(prob::NonlinearLeastSquaresProblem{uType, iip}, alg_::
     kwargs...) where {uType, iip}
     alg = get_concrete_algorithm(alg_, prob)
     @unpack f, u0, p = prob
+
+    # Use QR if the user did not specify a linear solver
+    if alg.linsolve === nothing || alg.linsolve isa QRFactorization ||
+       alg.linsolve isa FastQRFactorization
```
ChrisRackauckas, line 87: This doesn't make sense. It should be open so other methods like GMRES apply.

src/levenberg.jl
```diff
@@ -155,8 +163,26 @@ function SciMLBase.__init(prob::Union{NonlinearProblem{uType, iip},
     @unpack f, u0, p = prob
     u = alias_u0 ? u0 : deepcopy(u0)
     fu1 = evaluate_f(prob, u)
-    uf, linsolve, J, fu2, jac_cache, du, JᵀJ, v = jacobian_caches(alg, f, u, p, Val(iip);
-        linsolve_kwargs, linsolve_with_JᵀJ = Val(true))
+
+    # Use QR if the user did not specify a linear solver
```
ChrisRackauckas, line 167: Iinearsolve will default on non-square

src/levenberg.jl
```diff
@@ -155,8 +163,26 @@ function SciMLBase.__init(prob::Union{NonlinearProblem{uType, iip},
...
+    if (alg.linsolve === nothing || alg.linsolve isa QRFactorization ||
```
ChrisRackauckas, line 168: missing GMRES, but leave open

src/gaussnewton.jl
avik-pal: GMRES works on non-square? https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/main/src/gmres.jl

avik-pal: is there a `can_handle_non_square` trait?

src/utils.jl
```diff
@@ -206,8 +205,75 @@ function __get_concrete_algorithm(alg, prob)
...
+        b = ones(Float64, 3)
+        solve(LinearProblem(A, b), alg)
+        return false
+    catch err
+        return true
+    end
+end
+for alg in (:QRFactorization, :FastQRFactorization)
+    @eval needs_square_A(::$(alg)) = false
+end
```
avik-pal, line 279: @ChrisRackauckas is this a good check? :sweat:

src/utils.jl
```diff
@@ -206,8 +205,81 @@ function __get_concrete_algorithm(alg, prob)
...
+_try_factorize_and_check_singular!(::Nothing, x) = _issingular(x), false
+
+# Needs Square Matrix
+"""
+    needs_square_A(alg)
+
+Returns `true` if the algorithm requires a square matrix.
+"""
+needs_square_A(::Nothing) = false
```
ChrisRackauckas, line 266: Should be in LinearSolve?

src/utils.jl
```diff
@@ -206,8 +205,81 @@ function __get_concrete_algorithm(alg, prob)
...
+        return false
+    catch err
+        return true
+    end
+end
+for alg in (:QRFactorization, :FastQRFactorization, NormalCholeskyFactorization,
+    NormalBunchKaufmanFactorization)
+    @eval needs_square_A(::$(alg)) = false
+end
+for kralg in (LinearSolve.Krylov.lsmr!, LinearSolve.Krylov.craigmr!)
```
ChrisRackauckas, line 283: GMRES

avik-pal: GMRES requires square matrix

src/utils.jl
avik-pal: Yes. Once all the tests pass I will upstream

src/gaussnewton.jl
```diff
@@ -81,15 +81,29 @@ function SciMLBase.__init(prob::NonlinearLeastSquaresProblem{uType, iip}, alg_::
     kwargs...) where {uType, iip}
     alg = get_concrete_algorithm(alg_, prob)
     @unpack f, u0, p = prob
+
+    if !needs_square_A(alg.linsolve) && !(u isa Number) && !(u isa StaticArray)
```
axla-io, line 85: Should this be `u0` instead of `u`?

avik-pal: yes

src/gaussnewton.jl
```diff
@@ -81,15 +81,29 @@ function SciMLBase.__init(prob::NonlinearLeastSquaresProblem{uType, iip}, alg_::
     kwargs...) where {uType, iip}
     alg = get_concrete_algorithm(alg_, prob)
     @unpack f, u0, p = prob
+
+    if !needs_square_A(alg.linsolve) && !(u isa Number) && !(u0 isa StaticArray)
```
axla-io, line 85: Should this be `u0` instead of `u`? I get a `u not defined` error

src/gaussnewton.jl
```diff
@@ -99,12 +113,20 @@ end
...
-    linres = dolinsolve(alg.precs, linsolve; A = __maybe_symmetric(JᵀJ), b = _vec(Jᵀf),
-        linu = _vec(du), p, reltol = cache.abstol)
+    if JᵀJ !== nothing
+        __matmul!(JᵀJ, J', J)
+        __matmul!(Jᵀf, J', fu1)
+    end
+
+    # u = u - JᵀJ \ Jᵀfu
+    if cache.JᵀJ === nothing
+        linres = dolinsolve(alg.precs, linsolve; A = J, b = _vec(fu1), linu = _vec(du),
```
axla-io, line 124: I get an error stemming from here because the Jacobian is not square

avik-pal: what is the `linsolve.alg`? Is `A` or `b` some non-standard array type?

avik-pal: Also does specifying `linsolve = QRFactorization()` fix it?

axla-io: `linsolve.alg` is `nothing`. Changing to `QRFactorization()` gives a `not implemented `error. There could be some weird types yes, I'll try purging them...

axla-io: Still doesn't work with only standard arrays

avik-pal: Can you post the code for that?

src/utils.jl
ChrisRackauckas: That's interesting 😅 . Algorithmically that's not required. It is an algorithm that provably gives the L2 solution on non-square matrices, just like QR.

src/raphson.jl
```diff
@@ -96,7 +97,7 @@ function perform_step!(cache::NewtonRaphsonCache{true})
 
     # Line Search
     α = perform_linesearch!(cache.lscache, u, du)
-    @. u = u - α * du
+    axpy!(-α, du, u)
```
ChrisRackauckas, line 100: does this have appropriate generic fallbacks?

avik-pal: Done

ChrisRackauckas:
This needs a rebase?

avik-pal:
I am really unsure what to do with the Broyden/Klement failures. They fail on 1.9 but magically pass on 1.10

PR #264, opened 2023-10-26
Don't change the default termination condition

avik-pal:


avik-pal:
`NLSolveDefault` is from SimpleNonlinearSolve which works fine work simple usecases but for general problems it is almost always a terrible choice

ChrisRackauckas (approved):

PR #265, opened 2023-10-27
Proper handling of complex numbers and failures

avik-pal:
Fixes https://github.com/SciML/NonlinearSolve.jl/issues/269

* `__findmin` ignores NaNs
* TR returns `ConvergenceFailure` if it stops due to shrinking
* complex numbers gave a complex abstol/reltol
* TR solves an ill conditioned GN step which is not needed.
* Test for an infeasible case!
* Reworks the Termination Conditions and makes everything type stable (if chunksize and such are specified)!
* Return Codes are more meaningful

## TODO

- [x] Delete the Manifest before merging
- [x] https://github.com/JuliaDiff/SparseDiffTools.jl/pull/271
- [x] https://github.com/SciML/DiffEqBase.jl/pull/954

PR #268, opened 2023-11-01
Gauss Newton with Line Search

avik-pal:
# TODOs

- [ ] Fix Tests
- [x] Wait on https://github.com/SciML/NonlinearSolve.jl/pull/265
- [x] Upstream patch to allow non-square function operators for VecJac https://github.com/JuliaDiff/SparseDiffTools.jl/pull/272
- [x] Remove Manifest


vpuri3:
Any reason to not use `LineSearches.jl`?

avik-pal:
It is using LineSearches internally

avik-pal:
This is done. just need the previous PRs to be merged

Manifest.toml
```diff
@@ -0,0 +1,954 @@
+# This file is machine-generated - editing it directly is not advised
+
+julia_version = "1.10.0-beta3"
```
ChrisRackauckas, line 3: remove?

vpuri3:
would it be possible to pass an initial guess for alpha and line search? example? InitialStatic(), InitialPrevious(), etc...

Manifest.toml
avik-pal: it is to run the CI with the SparseDiffTools PR

avik-pal:
> would it be possible to pass an initial guess for alpha and line search? example? InitialStatic(), InitialPrevious(), etc...

Not currently. I have no clue how that API currently works.

PR #270, opened 2023-11-01
Replace `.*(Number, Diagonal)` with `*(Number, Diagonal)`

vpuri3:
The result with / without broadcasting here is the same. Removing the broadcast here would avoid https://github.com/JuliaGPU/CUDA.jl/issues/2140. Further, when `cache.DtD.diag isa ComponentVector` that CUDA bug leads to scalar indexing error on GPU.


oscardssmith (approved):

vpuri3:
are the failures related to this PR @avik-pal ?

avik-pal:
Rebase this should have been fixed

avik-pal (approved):

PR #272, opened 2023-11-02
Fix LM for banded jacobian

avik-pal:
- [ ] Make Multiple Shooting in https://github.com/SciML/BoundaryValueDiffEq.jl/pull/122 work Levenberg


avik-pal:
I think we can go ahead with this. The 23 test problems for some seem a bit flaky. This PR doesn't even tough Broyden and it broke...

PR #278, opened 2023-11-06
Use a proper polyalgorithm algorithm type

avik-pal:
Adding this since it makes it easy to write type-stable polyalgorithms. Once this is in we can simply define a custom polyalgorithm algorithm for https://github.com/SciML/NonlinearSolve.jl/pull/277 as well.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "NonlinearSolve"
 uuid = "8913a72c-1f9b-4ce2-8d82-65094dcecaec"
 authors = ["SciML"]
-version = "2.7.0"
+version = "2.8.0"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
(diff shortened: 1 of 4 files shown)


ChrisRackauckas (approved):

avik-pal:
I thing that might break is if someone was writing `::RobustMultiNewton`. Do we guarantee that the algorithms are types that can be dispatched on?

ChrisRackauckas:
no

avik-pal:
@ChrisRackauckas  is the forwarddiff linear solve failure from nonlinear solve or something changed in linear solve?

ChrisRackauckas:
That's a LinearSolve thing, don't worry about it.

ChrisRackauckas:
Can you make the 23 tests more robust?

avik-pal:
Probably disable the `GeneralBroyden`? It fails only sometimes on CI and locally always passes. I can't even reproduce the failure reliably

ChrisRackauckas:
Or make it simpler, like number of successes > 5

PR #279, opened 2023-11-06
Use Jacobian Based Algorithms if user supplies custom Jacobian

avik-pal:


Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "NonlinearSolve"
 uuid = "8913a72c-1f9b-4ce2-8d82-65094dcecaec"
 authors = ["SciML"]
-version = "2.8.0"
+version = "2.9.0"
```
ChrisRackauckas, line 4: Did this need a minor version?

ChrisRackauckas, line 4:
```suggestion
version = "2.8.1"
```

avik-pal: No really patch is fine

avik-pal:
I want to test this with BoundaryValueDiffEq once before merging

ChrisRackauckas (approved):

PR #282, opened 2023-11-13
Jacobian-Free Krylov Versions for TR/LM/GN

### LINKED ISSUE #167: TrustRegion is missing optimizations of using vjps directly
https://github.com/SciML/NonlinearSolve.jl/pull/159/files/17abaf0801fb7e61f7a5fafbfb89c2266aec9f40#diff-d4ded13aeff73c93c2977a4227ff0ba4a434dc462ff8cc433761133aff882405

This can be computed using jvps to avoid the full Jacobian

avik-pal:
- [x] Trust Region
- [x] Levenberg Marquardt
  - [x] LM can't support Krylov versions due to the $D^TD$ updates
  - [x] LinearSolve aliases `u`. Should we allocate 2 caches? Inefficient for factorizations where we want to reuse factorization. How do we prevent the aliasing
- [x] Gauss Newton
  - [x] Non-Square JacVec https://github.com/JuliaDiff/SparseDiffTools.jl/pull/274
- [x] Allow external control of the AD
- [x] Testing
- [x] Default to Zygote for Line Search if Zygote is loaded

Fixes #167


avik-pal:
@ChrisRackauckas this is good to go on my end

src/jacobian.jl
```diff
@@ -120,21 +125,72 @@ function __setup_linsolve(A, b, u, p, alg)
...
 __get_nonsparse_ad(::AutoSparseFiniteDiff) = AutoFiniteDiff()
 __get_nonsparse_ad(::AutoSparseZygote) = AutoZygote()
 __get_nonsparse_ad(ad) = ad
 
-__init_JᵀJ(J::Number) = zero(J)
-__init_JᵀJ(J::AbstractArray) = J' * J
-__init_JᵀJ(J::StaticArray) = MArray{Tuple{size(J, 2), size(J, 2)}, eltype(J)}(undef)
+__init_JᵀJ(J::Number, args...; kwargs...) = zero(J), zero(J)
+function __init_JᵀJ(J::AbstractArray, fu, args...; kwargs...)
+    JᵀJ = J' * J
```
ChrisRackauckas, line 137: what is actually using this?

src/jacobian.jl
```diff
@@ -120,21 +125,72 @@ function __setup_linsolve(A, b, u, p, alg)
...
+function __init_JᵀJ(J::FunctionOperator, fu, uf, u, args...; f = nothing,
+        vjp_autodiff = nothing, jvp_autodiff = nothing, kwargs...)
+    # FIXME: Proper fix to this requires the FunctionOperator patch
+    if f !== nothing && f.vjp !== nothing
+        @warn "Currently we don't make use of user provided `jvp`. This is planned to be \
+               fixed in the near future."
+    end
+    autodiff = __concrete_vjp_autodiff(vjp_autodiff, jvp_autodiff, uf)
+    Jᵀ = VecJac(uf, u; fu, autodiff)
+    JᵀJ_op = SciMLOperators.cache_operator(Jᵀ * J, u)
```
ChrisRackauckas, line 154: This operator shouldn't need to be constructed.

src/jacobian.jl
```diff
@@ -120,21 +125,72 @@ function __setup_linsolve(A, b, u, p, alg)
...
+            # VecJac can be only FiniteDiff
+            return AutoFiniteDiff()
+        else
+            # Short circuit if we see that FiniteDiff was used for J computation
+            jvp_autodiff isa AutoFiniteDiff && return jvp_autodiff
+            # Check if Zygote is loaded then use Zygote else use FiniteDiff
+            if haskey(Base.loaded_modules,
+                Base.PkgId(Base.UUID("e88e6eb3-aa80-5325-afca-941959d7151f"), "Zygote"))
+                return AutoZygote()
+            else
```
ChrisRackauckas, line 172: does this give inference issues?

src/jacobian.jl
avik-pal: None of the algorithms by default. But if LM/GN/TR is forced to use a Linear Solve which only works with square matrices then this needs to be triggered.

src/jacobian.jl
avik-pal: without doing the cache thing, it complained that for in place operations we need to run set the cache (something along those lines)

src/jacobian.jl
avik-pal: yeah probably, I had it before I implemented a way to control the ad from outside. I can remove it.

There is a way to avoid runtime check by having an extension define `is_loaded(::Val{:Zygote}) = true` with the default `is_loaded(::Val) = false` but that will cause invalidations, right?

ChrisRackauckas: Small invalidations would likely be better here.

ChrisRackauckas (approved):

PR #295, opened 2023-11-28
add spellcheck CI

ArnoStrouwen:
I left one mistake unfixed to see if it detects it.


ArnoStrouwen:
Seems to work nicely with low false positives.

PR #310, opened 2023-12-08
Up GPU env

avik-pal:


avik-pal:
Needs https://github.com/SciML/LinearSolve.jl/pull/443

PR #320, opened 2023-12-12
Use approximate sparsity detection by default

avik-pal:
See https://github.com/SciML/NonlinearSolve.jl/pull/318. Opening a new PR so that I can see the documentation previews


avik-pal:
Don't merge this. I am working on a way to control this from outside and add a manual entry explaining the choices.

But results are very promising:

```julia
132.129 s (1713050 allocations: 161.91 MiB)  # No Sparsity
  7.050 s (1784519 allocations: 4.00 GiB)    # Approx Sparsity + Default Linsolve
  6.579 s (1751088 allocations: 224.89 MiB)  # Approx Sparsity + KLU
  103.329 ms (52777 allocations: 387.83 MiB) # Approx Sparsity + GMRES
  5.082 s (84270 allocations: 3.90 GiB)      # Default Linsolve with Precomputed Sparsity
  6.408 s (50843 allocations: 119.60 MiB)    # KLU with Precomputed Sparsity
```

avik-pal:
It's not using it, I just ran it with the option on

ChrisRackauckas:
> we are smart enough to switch to the corresponding non-sparse type if concrete_jac is false

By non-sparse you mean Jacobian-free?

PR #326, opened 2023-12-16
Rework How We Update Alpha

yonatanwesen:
Rather than hardcoding how we update alpha, we now allow the user to provide a function that specifies the schema for updating alpha. The default is still the same(i.e. we still use the switch evolution relaxation method). I have added the documentation and the appropriate tests.  We have decided to move robust PT to a different PR

@ChrisRackauckas  @avik-pal


src/utils.jl
```diff
@@ -494,3 +494,30 @@ end
...
 @inline __diag(x::Number) = x
+
+#functions for updating alpha for PseudoTransient
+function switched_evolution_relaxation(alpha::Number,
+        res_norm::Number,
+        nsteps::Int,
+        u,
+        u_prev,
+        fu,
+        norm::F) where {F}
```
avik-pal, line 505: Format it into 2 lines like:

```julia
switched_evolution_relaxation(alpha::Number, ....,
		......)
```

src/utils.jl
```diff
@@ -494,3 +494,30 @@ end
 @inline __diag(x::AbstractMatrix) = diag(x)
 @inline __diag(x::AbstractVector) = x
 @inline __diag(x::Number) = x
+
+#functions for updating alpha for PseudoTransient
```
avik-pal, line 498:
```suggestion
# Functions for updating alpha for PseudoTransient
```

src/utils.jl
```diff
@@ -494,3 +494,30 @@ end
...
+        nsteps::Int,
+        u,
+        u_prev,
+        fu,
+        norm::F) where {F}
+    if nsteps ≤ 100
+        return alpha
+    else
+        new_norm = norm(fu)
+        return alpha * (res_norm / new_norm)
```
avik-pal, line 521: Same here

test/23_test_problems.jl
```diff
@@ -123,3 +123,12 @@ end
...
 end
+
+@testset "RobustPseudoTransient 23 Test Problems" begin
+    alg_ops = (RobustPseudoTransient(; alpha_initial = 10.0),)
+
+    broken_tests = Dict(alg => Int[] for alg in alg_ops)
+    broken_tests[alg_ops[1]] = [1, 6, 22]
+
+    test_on_library(problems, dicts, alg_ops, broken_tests)
+end
```
avik-pal, line 134: Club this into PT, follow how the other tests are setup.

test/basictests.jl
```diff
@@ -685,6 +685,113 @@ end
...
+    end
+
+    @testset "Termination condition: $(termination_condition) u0: $(_nameof(u0))" for termination_condition in TERMINATION_CONDITIONS,
+        u0 in (1.0, [1.0, 1.0])
+
+        probN = NonlinearProblem(quadratic_f, u0, 2.0)
+        @test all(abs.(solve(probN, RobustPseudoTransient(; alpha_initial = 0.01);
+            termination_condition).u) .≈ sqrt(2.0))
+    end
+end
```
avik-pal, line 793: Same here don't maintain copies of the exact same code

src/pseudotransient.jl
```diff
@@ -139,7 +162,13 @@ function perform_step!(cache::PseudoTransientCache{iip}) where {iip}
...
 
     new_norm = cache.internalnorm(cache.fu)
-    cache.alpha *= cache.res_norm / new_norm
+    cache.alpha = cache.alg.update_alpha(cache.alpha,
+        cache.res_norm,
+        cache.stats.nsteps,
+        cache.u,
+        cache.u_cache,
+        cache.fu,
+        cache.internalnorm)
```
avik-pal, line 171: Same formatting problem

src/pseudotransient.jl
```diff
@@ -11,6 +11,12 @@ the time-stepping and algorithm, please see the paper:
 [Coffey, Todd S. and Kelley, C. T. and Keyes, David E. (2003), Pseudotransient Continuation and Differential-Algebraic Equations,
 SIAM Journal on Scientific Computing,25, 553-569.](https://doi.org/10.1137/S106482750241044X)
 
+RobustPseudoTransient(; concrete_jac = nothing, linsolve = nothing,
+precs = DEFAULT_PRECS, alpha_initial = 1e-6,update_alpha=robust_update_alpha, adkwargs...)
+
```
avik-pal, line 16: needs 4 spaces. check how the docs are being printed in the repl

src/pseudotransient.jl
```diff
@@ -1,6 +1,6 @@
 """
     PseudoTransient(; concrete_jac = nothing, linsolve = nothing,
-        precs = DEFAULT_PRECS, alpha_initial = 1e-3, adkwargs...)
+        precs = DEFAULT_PRECS, alpha_initial = 1e-3,update_alpha= switched_evolution_relaxation,adkwargs...)
```
avik-pal, line 3: weird spacing

src/pseudotransient.jl
```diff
@@ -11,6 +11,12 @@ the time-stepping and algorithm, please see the paper:
 [Coffey, Todd S. and Kelley, C. T. and Keyes, David E. (2003), Pseudotransient Continuation and Differential-Algebraic Equations,
 SIAM Journal on Scientific Computing,25, 553-569.](https://doi.org/10.1137/S106482750241044X)
 
+RobustPseudoTransient(; concrete_jac = nothing, linsolve = nothing,
+precs = DEFAULT_PRECS, alpha_initial = 1e-6,update_alpha=robust_update_alpha, adkwargs...)
```
avik-pal, line 15: Put this as the documentation for RobustPT and not in PT. Cross reference from here as `See also .....`

src/pseudotransient.jl
```diff
@@ -33,22 +39,39 @@ SIAM Journal on Scientific Computing,25, 553-569.](https://doi.org/10.1137/S1064
     [LinearSolve.jl documentation](https://docs.sciml.ai/LinearSolve/stable/).
   - `alpha_initial` : the initial pseudo time step. it defaults to 1e-3. If it is small,
     you are going to need more iterations to converge but it can be more stable.
+  - `update_alpha`  : a function that specifies the schema for updating alpha. The function
+    should take in alpha,the norm of the previous residual(i.e |fu|),number of steps so far,
+    u,u_prev,fu. The default is a function that uses "switched evolution relaxation" SER method to update alpha.
```
avik-pal, line 44: Show how the function is called rather than a textual descriptiion

src/utils.jl
```diff
@@ -494,3 +494,30 @@ end
...
+
+function robust_update_alpha(alpha::Number,
+        res_norm::Number,
+        nsteps::Int,
+        u,
+        u_prev,
+        fu,
+        norm::F) where {F}
+    if nsteps ≤ 100
+        return alpha
```
avik-pal, line 518: Why `100`? Is this supported by benchmarks? I would expose this as an option via RobustPT.

avik-pal (changes requested):

src/utils.jl
yonatanwesen: Yes. I played around with threshold and 100 is the best option. I think the user can define their own function easily if they want something else. I don't see how I can expose the threshold without affecting signficant portion of the pt methods.

avik-pal: Not really, it should affect PT at all. Create a closure with the outer function taking `n` as input, or use partial functions

src/pseudotransient.jl
```diff
@@ -33,22 +33,44 @@ SIAM Journal on Scientific Computing,25, 553-569.](https://doi.org/10.1137/S1064
...
+        update_alpha::F = switched_evolution_relaxation,
+        autodiff = nothing) where {F}
     return PseudoTransient{_unwrap_val(concrete_jac)}(autodiff, linsolve, precs,
-        alpha_initial)
+        alpha_initial, update_alpha)
+end
+
+"""
+    RobustPseudoTransient(; concrete_jac = nothing, linsolve = nothing,
+        precs = DEFAULT_PRECS, alpha_initial = 1e-6,update_alpha = robust_update_alpha, adkwargs...)
```
avik-pal, line 63: fix the inconsistent whitespace formattings. also `adkwargs` doesn't exist

test/basictests.jl
```diff
@@ -570,29 +570,38 @@ end
...
+        return solve(prob,
+            PseudoTransient(; linsolve, precs, alpha_initial, update_alpha),
+            abstol = 1e-9)
     end
 
-    @testset "PT: alpha_initial = 10.0 PT AD: $(ad)" for ad in (AutoFiniteDiff(),
-        AutoZygote())
+    @testset "update_alpha:$(update_alpha) alpha_initial = $(alpha_initial) AD: $(ad)" for (update_alpha, alpha_initial) in zip((switched_evolution_relaxation,
+                robust_update_alpha),
+            (10.0, 0.01)),
```
avik-pal, line 592: Why are you having to change the alpha for a "robust" method? A robust method is by definition supposed to work without having to tune alpha

test/basictests.jl
```diff
@@ -604,46 +613,75 @@ end
             if prec === :Random
                 prec = (args...) -> (Diagonal(randn!(similar(u0))), nothing)
             end
-            sol = benchmark_nlsolve_iip(quadratic_f!, u0; linsolve, precs = prec)
+            sol = benchmark_nlsolve_iip(quadratic_f!,
+                u0;
+                linsolve,
+                precs = prec,
+                alpha_initial,
+                update_alpha)
```
avik-pal, line 621: weird formatting

test/basictests.jl
```diff
@@ -604,46 +613,75 @@ end
...
+            robust_update_alpha),
+        (10.0, 0.01))
         for p in 1.0:0.1:100.0
             @test begin
-                res = benchmark_nlsolve_oop(quadratic_f, @SVector[1.0, 1.0], p)
+                res = benchmark_nlsolve_oop(quadratic_f,
+                    @SVector[1.0, 1.0],
+                    p;
+                    alpha_initial,
+                    update_alpha)
```
avik-pal, line 641: same

test/basictests.jl
```diff
@@ -604,46 +613,75 @@ end
...
+            p;
+            alpha_initial,
+            update_alpha).u) ≈ sqrt(p[2] / p[1])
+        @test ForwardDiff.jacobian(p -> [
+                abs.(benchmark_nlsolve_oop(quadratic_f2,
+                    0.5,
+                    p;
+                    alpha_initial,
+                    update_alpha).u),
+            ],
```
avik-pal, line 682: same

test/basictests.jl
```diff
@@ -661,27 +699,37 @@ end
     @test nlprob_iterator_interface(quadratic_f, p, Val(false)) ≈ sqrt.(p)
     @test nlprob_iterator_interface(quadratic_f!, p, Val(true)) ≈ sqrt.(p)
 
-    @testset "ADType: $(autodiff) u0: $(_nameof(u0))" for autodiff in (AutoSparseForwardDiff(),
+    @testset "update_alpha:$(update_alpha) alpha_initial = $(alpha_initial),ADType: $(autodiff) u0: $(_nameof(u0))" for (update_alpha, alpha_initial) in zip((switched_evolution_relaxation,
```
avik-pal, line 702: spaces

test/basictests.jl
```diff
@@ -661,27 +699,37 @@ end
...
+        @test all(abs.(solve(probN,
+            PseudoTransient(; alpha_initial, autodiff, update_alpha)).u) .≈
                   sqrt(2.0))
     end
-
-    @testset "NewtonRaphson Fails but PT passes" begin # Test that `PseudoTransient` passes a test that `NewtonRaphson` fails on.
+    # Test that `PseudoTransient` passes a test that `NewtonRaphson` fails on.
+    @testset "update_alpha: $(update_alpha) alpha_initial = $(alpha_initial) NewtonRaphson Fails but PT passes" for (update_alpha, alpha_initial) in zip((switched_evolution_relaxation,
+            robust_update_alpha),
+        (1.0, 0.1))
```
avik-pal, line 716: same here why tune alpha?

src/pseudotransient.jl
```diff
@@ -33,22 +33,44 @@ SIAM Journal on Scientific Computing,25, 553-569.](https://doi.org/10.1137/S1064
     [LinearSolve.jl documentation](https://docs.sciml.ai/LinearSolve/stable/).
   - `alpha_initial` : the initial pseudo time step. it defaults to 1e-3. If it is small,
     you are going to need more iterations to converge but it can be more stable.
+  - `update_alpha`  : a function that specifies the schema for updating alpha. The function should
+    like update_alpha(alpha::Number,res_norm::Number,nsteps::Int,u,u_prev,fu,norm::F) The default is
+    a function that uses "switched evolution relaxation" SER method to update alpha.
```
avik-pal, line 38: make the spacing consistent

src/NonlinearSolve.jl
```diff
@@ -233,9 +233,11 @@ include("default.jl")
 end
 
 export RadiusUpdateSchemes
+export switched_evolution_relaxation, robust_update_alpha
 
 export NewtonRaphson, TrustRegion, LevenbergMarquardt, DFSane, GaussNewton, PseudoTransient,
-    Broyden, Klement, LimitedMemoryBroyden
+    RobustPseudoTransient
+Broyden, Klement, LimitedMemoryBroyden
```
avik-pal, line 240: this breaks the export

avik-pal (changes requested):

test/basictests.jl
yonatanwesen: We still have to choose an initial alpha for these problems. The problems were designed to test NewtonRaphson, hence it helps to set the initial alpha slightly higher. There is no single initial alpha that works for every problem. I had to make a guess that works for these problems. The point here is that robust one converges even if we start with a much lower alpha compared to the regular PT with the default maxiters. I can set this to small alpha, but it would just take more iterations. From our discussion, what I understood was Robust one will be slower, but it will try to find a stable solution.

avik-pal: So this works if the alpha is not changed? In which case it should not be changed

avik-pal: Also you are testing for a problem with an unstable equilibrium, either change it to

1. Problem with a stable equilibrium
2. initialize correctly so that the vector field doesn't make the euler diverge

PR #337, opened 2023-12-22
FixedPointSolvers: Recommendations and Wrappers

### LINKED ISSUE #76: SpeedMapping.jl wrapper

### LINKED ISSUE #78: FixedPointAcceleration.jl
https://github.com/s-baumann/FixedPointAcceleration.jl

avik-pal:
Fixes #76, Fixes #78


docs/src/solvers/FixedPointSolvers.md
```diff
@@ -0,0 +1,42 @@
+# Fixed Point Solvers
+
+Currently we don't have an API to directly specify Fixed Point Solvers. However, a Fixed
+Point Problem can be triviall converted to a Root Finding Problem. Say we want to solve:
```
ChrisRackauckas, line 4:
```suggestion
Point Problem can be trivially converted to a Root Finding Problem. Say we want to solve:
```

avik-pal:
@ChrisRackauckas what do you think about introducing a `FixedPointProblem` type? I don't quite like the idea of doing the back and forth of adding and subtracting `u`, also I feel it is easy to make a mistake while providing the custom jacobian and such

ext/NonlinearSolveFixedPointAccelerationExt.jl
```diff
@@ -0,0 +1,56 @@
+module NonlinearSolveFixedPointAccelerationExt
+
+using NonlinearSolve, FixedPointAcceleration, DiffEqBase, SciMLBase
+
+function SciMLBase.__solve(prob::NonlinearProblem, alg::FixedPointAccelerationJL, args...;
+        abstol = nothing, maxiters = 1000, alias_u0::Bool = false,
+        show_trace::Val{PrintReports} = Val(false), termination_condition = nothing,
```
ErikQQY, line 7: ```show_trace``` keyword is passed during the method constructing in ```NLsolveJL``` and ```CMINPACK```, I think we'd better keep a unified interface here.

avik-pal: I honestly wanted to get rid of those and have them supported similar to https://docs.sciml.ai/NonlinearSolve/stable/basics/solve/#Tracing-Controls. Mostly because, I feel tracing is something that I would occasionally turn on for debugging and having it as a kwarg is more convenient.

But we could not forward those kwargs for non-native solvers

ErikQQY: Agreed, ```show_trace``` would be more convenient as a kwarg in solve when debugging

ChrisRackauckas:
> @ChrisRackauckas what do you think about introducing a FixedPointProblem type? I don't quite like the idea of doing the back and forth of adding and subtracting u, also I feel it is easy to make a mistake while providing the custom jacobian and such

Yeah that sounds like a good thing to add at some point, where some methods would work on that natively and others would convert.

ChrisRackauckas (approved):

PR #339, opened 2023-12-24
[WIP] Take on LsqFit.jl

avik-pal:
## Checklist

- [ ] Add documentation
- [ ] Support a Weight Matrix
- [ ] StatsAPI


src/curve_fit.jl
```diff
@@ -0,0 +1,32 @@
+function curve_fit end
+
+function curve_fit(f::F, xdata, ydata, p₀, args...; kwargs...) where {F <: Function}
```
ChrisRackauckas, line 3: I kind of think it should be its own package since it will clutter the docs to have this kind of alternative function around. Also for discoverability.

avik-pal: That makes sense. Do you want that to be an exclusive package for `curve_fit` or have wrappers around `root_find`, `fixed_point`, etc. as well?

ChrisRackauckas: Having those 3 would be fine. I don't think the other two are so important but might as well have them if it's easy.

avik-pal: name suggestions?

ChrisRackauckas: I was thinking CurveFit.jl, but I see that's taken 😅. RobustCurveFit.jl? SimpleCurveFit? Or we can just ask them https://github.com/pjabardo/CurveFit.jl if they would want to migrate this over to SciML and we could then add stuff to it. I think that has some interesting ideas.

PR #341, opened 2023-12-24
Standardize the Extension Algorithms

avik-pal:
## Checklist

- [x] Standardized Kwargs
- [x] Forward the tracing kwargs and deprecate the ones in the Solver Kwargs
- [x] Use Jacobian Handling via `jacobian!!` in FastLM


ext/NonlinearSolveSIAMFANLEquationsExt.jl
```diff
@@ -2,51 +2,58 @@ module NonlinearSolveSIAMFANLEquationsExt
...
+    atol = NonlinearSolve.DEFAULT_TOLERANCE(abstol, eltype(prob.u0))
+    rtol = NonlinearSolve.DEFAULT_TOLERANCE(reltol, eltype(prob.u0))
 
     if prob.u0 isa Number
-        f! = if iip
-            function (u)
-                du = similar(u)
-                prob.f(du, u, prob.p)
-                return du
-            end
```
avik-pal, line 26: This shouldn't originate in the first place.

docs/src/basics/solve.md
```diff
@@ -14,8 +14,8 @@ solve(prob::SciMLBase.NonlinearProblem, args...; kwargs...)
 ## Iteration Controls
 
   - `maxiters::Int`: The maximum number of iterations to perform. Defaults to `1000`.
-  - `abstol::Number`: The absolute tolerance.
-  - `reltol::Number`: The relative tolerance.
+  - `abstol::Number`: The absolute tolerance. Defaults to `real(oneunit(T)) * (eps(real(one(T))))^(4 // 5)`.
+  - `reltol::Number`: The relative tolerance. Defaults to `real(oneunit(T)) * (eps(real(one(T))))^(2 // 5)`.
```
ErikQQY, line 18: While now we are using ```_get_tolerance``` from DiffEqBase.jl, shouldn't both ```reltol``` and ```abstol``` are ```real(oneunit(T)) * (eps(real(one(T))))^(4 // 5)```?

ErikQQY: See:
https://github.com/SciML/DiffEqBase.jl/blob/898e6bf252dbd3e4a0561b19f2015b9d6e0a8ffb/src/termination_conditions.jl#L219-L226

avik-pal: Oops yeah typo

avik-pal, line 18:
```suggestion
  - `reltol::Number`: The relative tolerance. Defaults to `real(oneunit(T)) * (eps(real(one(T))))^(4 // 5)`.
```

ext/NonlinearSolveSIAMFANLEquationsExt.jl
ErikQQY: Yeah, the scalar case root finding doesn't need this

src/jacobian.jl
```diff
@@ -9,7 +9,7 @@ isinplace(JᵀJ::KrylovJᵀJ) = isinplace(JᵀJ.Jᵀ)
 
 # Select if we are going to use sparse differentiation or not
 sparsity_detection_alg(_, _) = NoSparsityDetection()
-function sparsity_detection_alg(f, ad::AbstractSparseADType)
+function sparsity_detection_alg(f::NonlinearFunction, ad::AbstractSparseADType)
```
ChrisRackauckas, line 12: Does this need to specialize?

ChrisRackauckas (approved):
I didn't go through all the details but it seems generally in the right direction with all of it.

ChrisRackauckas:
Are the 23 test problem failures known?

src/jacobian.jl
avik-pal: safety measure, we are using `f.sparsity` and stuff in there.

PR #342, opened 2023-12-25
Fix the 23 test problems

avik-pal:
Needs https://github.com/SciML/DiffEqProblemLibrary.jl/pull/122


ChrisRackauckas:
oh it's here 😅 . Rebase and merge.

Issue #221, opened 2023-09-23
Improve docstrings


ArnoStrouwen:
Such that `makedocs` `warnonly = [:missing_docs, :cross_references]` can be removed.

avik-pal:
`warnonly = [:missing_docs]` still present

PR #343, opened 2023-12-27
typo in makedocs

ArnoStrouwen:
https://github.com/SciML/NonlinearSolve.jl/issues/221 should likely not have been closed.
It likely throws no errors about missing docs because of the typo.


avik-pal:
Oops yeah. One thing, though, none of the "non included" ones should be included. Is there a way to pass ignore these.. to makedocs?

ArnoStrouwen:
I don't think so.
See, https://github.com/JuliaDocs/Documenter.jl/issues/2383.

avik-pal:
re-enabled the `warnonly` argument. Let's wait for https://github.com/JuliaDocs/Documenter.jl/issues/2383 before we can disable it

Issue #330, opened 2023-12-18
Add new algorithm, the semi-implicit Continuous Newton method


rzyu45:
**What kind of problems is it mostly used for? Please describe.**

Nonlinear algebraic equations, including but not limited to the [AC power flow models](https://en.wikipedia.org/wiki/Power-flow_study).

**Describe the algorithm you’d like**

As depicted in [1], for equations 

$$0=g(y),$$

we can convert it to

```math
\begin{aligned}
\dot{y}&=z\\
0&=J(y)z+g(y)
\end{aligned},
```
where $J$ is the jaobian, and then solve the DAEs with the new 4-stage 3rd-order L-stable RODAS3D constructed in [1]. The values of $z$ can be initialized by $-J^{-1}(y_0)g(y_0)$ for consistency.

It is proved in [1] that if the initial values are within the region of attraction of the converted dynamic systems, then the continuous Newton method is bounded to converge. So that the semi-implicit Continuous Newton method is very stable.

**Other implementations to know about**

Currently there is no other implementation to the best of my knowledge. Maybe I can do it here.

**References**
[1] [Semi-implicit Continuous Newton Method for Power Flow Analysis](https://arxiv.org/abs/2312.02809)
[2] [Implicit Continuous Newton Method for Power Flow Analysis](https://ieeexplore.ieee.org/document/8698293)
[3] [Continuous Newton's Method for Power Flow Analysis](https://ieeexplore.ieee.org/document/4682629)

avik-pal:
Yeah these would be interesting to have. Though we might have to implement this in something like SteadyStateDiffEq, since we can't take a dep on OrdinaryDiffEq (circular dep).

Let us know if you need help getting started. I am guessing the DAE solver goes in OrdinaryDiffEq (@ChrisRackauckas)?

ChrisRackauckas:
Yeah the DAE solver could go into OrdinaryDiffEq and we can have this in SteadyStateDiffEq.jl, I was thinking the same thing. It is a special DAE solver so it's not one that we have right now, but I think it's the right choice to focus it entirely on stability and not accuracy given accurately solving the ODE isn't the goal but making it stable for big dts is.

rzyu45:
@avik-pal  @ChrisRackauckas 
I have understood the dep relationships. Could you please offer me some dev tutorials? The way to use the existing interfaces to generate Jac and Hessian-Vector Products will be of significance.

avik-pal:
[...]

Issue #247, opened 2023-10-18
Completeness for use in OrdinaryDiffEq.jl


ChrisRackauckas:
This package is almost complete to be used in OrdinaryDiffEq.jl. For other discussions on this topic, see https://github.com/SciML/OrdinaryDiffEq.jl/issues/1570

However, there are some special things done in its nonlinear solvers which we would need to allow for here:

1. First step is completing the iterator interface https://github.com/SciML/NonlinearSolve.jl/issues/241
2. We need to have a version of a Quasi-Newton method which forces Jacobian reuse and terminates if convergence starts to fail. I think this can be done just with a `SingleFactorizeNewton` and some extended `TerminationCondition` stuff.
3. For this `SingleFactorizeNewton`, we'd need to allow it in the iterator form to have a mode where changing `u` does not cause a new Jacobian to be taken. Thus in the `cache` interface, we'd need a `reevalute_jacobian!` that can be handled outside the solver for such cases.

Once those are together, then I think we're good to try it in OrdinaryDiffEq?

oscardssmith:
@ChrisRackauckas I believe this can now be closed, right?

Issue #241, opened 2023-10-17
Complete the iterator interface


ChrisRackauckas:
The iterator interface for NonlinearSolve.jl was never completed in the past, but it's very close. Next steps are:

- [ ] add dispatches for the common solve `step!` interface
- [ ] Setup the abstract caches as iterators that automatically iterate, and test with for loops
- [ ] Update the iterator interface page to showcase some uses, like in for loops https://docs.sciml.ai/NonlinearSolve/stable/tutorials/iterator_interface/

As inspiration, see https://docs.sciml.ai/DiffEqDocs/stable/basics/integrator/

avik-pal:
This was done with 3.5 release

Issue #216, opened 2023-09-21
Simple high order multistep schemes


ChrisRackauckas:
https://link.springer.com/article/10.1007/s40314-022-02159-9

PR #345, opened 2023-12-28
What is a Nonlinear Solver? And How to easily build Newer Ones

### LINKED ISSUE #62: Can not iterate solver before solving
When attempting to iterate a solver that has not been solved, the user is told that there is no method matching `iterate` for the solver. Same goes for `getindex`:
```
julia> using NonlinearSolve

julia> f(u, p) = u .* u .- 2.0
f (generic function with 1 method)

julia> u0 = (1.0, 2.0) # brackets
(1.0, 2.0)

julia> probB = NonlinearProblem(f, u0)
NonlinearProblem with uType Tuple{Float64, Float64}. In-place: false
u0: (1.0, 2.0)

julia> solver = init(probB, Falsi()) # Can iterate the solver object
NonlinearSolve.BracketingImmutableSolver{NonlinearFunction{false, typeof(f), LinearAlgebra.UniformScaling{Bool}, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, typeof(SciMLBase.DEFAULT_OBSERVED_NO_TIME), Nothing}, Falsi, Float64, Float64, SciMLBase.NullParameters, Nothing, NonlinearProblem{Tuple{Float64, Float64}, false, SciMLBase.NullParameters, NonlinearFunction{false, typeof(f), LinearAlgebra.UniformScaling{Bool}, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, typeof(SciMLBase.DEFAULT_OBSERVED_NO_TIME), Nothing}, Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}}}(1, NonlinearFunction{false, typeof(f), LinearAlgebra.UniformScaling{Bool}, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, typeof(SciMLBase.DEFAULT_OBSERVED_NO_TIME), Nothing}(f, LinearAlgebra.UniformScaling{Bool}(true), nothing, nothing, nothing, nothing, nothing, nothing, nothing, nothing, nothing, nothing, nothing, SciMLBase.DEFAULT_OBSERVED_NO_TIME, nothing), Falsi(), 1.0, 2.0, -1.0, 2.0, SciMLBase.NullParameters(), false, 1000, NonlinearSolve.DEFAULT, nothing, false, NonlinearProblem{Tuple{Float64, Float64}, false, SciMLBase.NullParameters, NonlinearFunction{false, typeof(f), LinearAlgebra.UniformScaling{Bool}, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, typeof(SciMLBase.DEFAULT_OBSERVED_NO_TIME), Nothing}, Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}}(NonlinearFunction{false, typeof(f), LinearAlgebra.UniformScaling{Bool}, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, typeof(SciMLBase.DEFAULT_OBSERVED_NO_TIME), Nothing}(f, LinearAlgebra.UniformScaling{Bool}(true), nothing, nothing, nothing, nothing, nothing, nothing, nothing, nothing, nothing, nothing, nothing, SciMLBase.DEFAULT_OBSERVED_NO_TIME, nothing), (1.0, 2.0), SciMLBase.NullParameters(), Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}()))

julia> solver[1]
ERROR: MethodError: no method matching getindex(::NonlinearSolve.BracketingImmutableSolver{NonlinearFunction{false, typeof(f), LinearAlgebra.UniformScaling{Bool}, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, typeof(SciMLBase.DEFAULT_OBSERVED_NO_TIME), Nothing}, Falsi, Float64, Float64, SciMLBase.NullParameters, Nothing, NonlinearProblem{Tuple{Float64, Float64}, false, SciMLBase.NullParameters, NonlinearFunction{false, typeof(f), LinearAlgebra.UniformScaling{Bool}, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, typeof(SciMLBase.DEFAULT_OBSERVED_NO_TIME), Nothing}, Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}}}, ::Int64)
Stacktrace:
 [1] top-level scope
   @ REPL[6]:1

julia> for s in solver
           @show s
       end
ERROR: MethodError: no method matching iterate(::NonlinearSolve.BracketingImmutableSolver{NonlinearFunction{false, typeof(f), LinearAlgebra.UniformScaling{Bool}, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, typeof(SciMLBase.DEFAULT_OBSERVED_NO_TIME), Nothing}, Falsi, Float64, Float64, SciMLBase.NullParameters, Nothing, NonlinearProblem{Tuple{Float64, Float64}, false, SciMLBase.NullParameters, NonlinearFunction{false, typeof(f), LinearAlgebra.UniformScaling{Bool}, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, typeof(SciMLBase.DEFAULT_OBSERVED_NO_TIME), Nothing}, Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}}})
Closest candidates are:
  iterate(::Union{LinRange, StepRangeLen}) at C:\Users\Dennis Bal\.julia\juliaup\julia-1.7.1+0~x64\share\julia\base\range.jl:826
  iterate(::Union{LinRange, StepRangeLen}, ::Integer) at C:\Users\Dennis Bal\.julia\juliaup\julia-1.7.1+0~x64\share\julia\base\range.jl:826
  iterate(::T) where T<:Union{Base.KeySet{<:Any, <:Dict}, Base.ValueIterator{<:Dict}} at C:\Users\Dennis Bal\.julia\juliaup\julia-1.7.1+0~x64\share\julia\base\dict.jl:695
  ...
Stacktrace:
 [1] top-level scope
   @ .\REPL[7]:1

julia> solver = solve!(solver)
u: 1.414213562373095

julia> solver[1]
1.414213562373095

julia> for s in solver
           @show s
       end
s = 1.414213562373095
```

However, should the unsolved solver not be an empty collection?

This is especially confusing due to the documentation, which says that you can iterate the solver in the line before it is solved. From http://nonlinearsolve.sciml.ai/dev/tutorials/iterator_interface/ :
```
f(u, p) = u .* u .- 2.0
u0 = (1.0, 2.0) # brackets
probB = NonlinearProblem(f, u0)
solver = init(probB, Falsi()) # Can iterate the solver object
solver = solve!(solver)
```

### LINKED ISSUE #312: Normal Form Linear Solve in Trust Region Methods
Currently we cannot use Normal Form equations in TR. This can be added similar to how Gauss Newton and Levenberg work. Allows us to use Iterative Solvers like GMRES.

In the long run, we want to support LSLQ and such solvers but that requires a minor rewrite of JacVec and VecJac to provide the correct adjoint operators

avik-pal:
## Goals

- [x] Fully Functional Iterator Interface
- [x] Composable Internals. https://github.com/SciML/NonlinearSolve.jl/issues/330 and https://github.com/SciML/NonlinearSolve.jl/issues/216 got me thinking that our level of abstraction before was way off. All our algorithms are essentially the same (okay maybe except the spectral one, but even that can be made composable and we get a bunch of those methods for free) but we keep copying the code around
- [ ] ~Dispatch system with safety to disable NLLS with NewtonRaphson (and similarly for others). Currently it will work which is not the worst thing, but disabling this is a safer option IMO~
- [x] Actually functional `reinit` interface

## How close is this to being ready?

Once tests pass and downstream PRs are merged we should be good to go. I will follow this PR up with:

1. SingleFactorize<....> which can use the iterator interface and `recompute_jacobian` to Jacobian only when needed.
2. Multi-Step Methods
3. Default change to PolyesterForwardDiff similar to SimpleNonlinearSolve

## Documentation

- [x] Dev Docs

## Target Issues

- [x] Fixes #312
- [x] Setup for #247 
- [x] Setup for #241
- [x] Fixes #62 

## Modular Internals

- [x] Descent
  - [x] Newton -- Automatically handle normal form efficiently
  - [x] Cauchy
  - [x] Dogleg -- Mix Newton and Cauchy
  - [x] Damped Newton -- Yes Levenberg-Marquardt and PseudoTransient can share the same internals
    - [x] Rework a little bit for different schemes
- [x] Globalization: Only one can be activated at a time
  - [x] Line Search
  - [x] Trust Region


src/descent/steepest.jl
```diff
@@ -0,0 +1,66 @@
+"""
+    SteepestDescent(; linsolve = nothing, precs = DEFAULT_PRECS)
+
+Compute the descent direction as ``δu = -Jᵀfu``.
```
ChrisRackauckas, line 4: why is this steepest?

avik-pal: That's the gradient of $\frac{1}{2}||f(x)||_2$

termi-official:
I may have a special case here to consider and will write it here instead of as a separate issue due to the title of the PR. If I should transfer this information to an issue instead, I am happy to do it.

When discretizing PDEs we often end up with a nonlinear system of equations + constraints (coming from e.g. Dirichlet boundary conditions). For example in Newton methods we also know how to enforce the constraints efficiently: We can rewrite the linear equation system. In Ferrite.jl a typical Newton-Raphson implementation looks like this:
```julia
dh = ... # Holds info about finite element problem
un = zeros(ndofs(dh))
u = zeros(ndofs(dh))
Δu = zeros(ndofs(dh))
... # define remaining variables
dbcs = ... # Holds Dirichlet boundary condition information 
apply!(un, dbcs) # Apply Dirichlet conditions to solution vector
while true; newton_itr += 1
        # Construct the current guess
        u .= un .+ Δu
        # Compute residual and tangent for current guess
        assemble_global!(K, g, dh, cv, fv, mp, u, ΓN)
        # Apply boundary conditions
        apply_zero!(K, g, dbcs)
        # Compute the residual norm and compare with tolerance
        normg = norm(g)
        if normg < NEWTON_TOL
            break
        elseif newton_itr > NEWTON_MAXITER
            error("Reached maximum Newton iterations, aborting")
        end

        # Compute increment
        ΔΔu =  K \ g

        apply_zero!(ΔΔu, dbcs)
        Δu .-= ΔΔu
    end
```
(see https://ferrite-fem.github.io/Ferrite.jl/dev/tutorials/hyperelasticity/ for full example).

I am confident that other frameworks come with similar mechanisms. Basically the difference to a "normal" Newton-Raphson is the `apply!` before entering the iteration to e.g. set the boundary condition in the solution vector and the 2x `apply_zero!`, first to e.g. eliminate the boundary conditions from the linear system and second to remove numerical inaccurracies in the solver from the increment. If it is not clear: The increment must be zero, because the solution is already enforced to be correct by the first call to `apply!`. For e.g. gradient based solvers we need similar hooks.

With this in mind I would like to ask for considering this and possibly recommendations about how this integration can be accompished from user side when using NonlinearSolve.jl. I am happy to help here and also to answer any questions which might come up.

My first idea here was to provide a custom linear solver. This custom solver just wraps the actual linear solver and adds the apply_zero calls in the `solve` function (and when the linear system is regenerated). However, in this case it is not clear how to introduce the first `apply!` in a clean way before going into the iteration.

avik-pal:
One way might be to provide a `preamble` / `postamble` functions with the input being `cache`. I think that might solve your problem @termi-official? Then you can construct the `GeneralizedFirstOr....` method with `NewtonDescent` to get your `NewtonRaphson`.

The only concern here is that with custom functions, the inputs to the `step!` like `recompute_jacobian` might have to be set as incompatible -- unless we rework how the jacobian or `fu` is stored / cached.

termi-official:
So, if I understand correctly, then the current recommendation would be to define a custom linear solver wrapper, as I proposed, and then dispatch the preamble if the cache has this specific linear solver wrapper. I think this makes sense.

Indeed one thing to consider here is that the Dirichlet condition might be time-dependent themselves and this can possibly invalidate the Jacobian (depending on the reusage strategy). I am not sure how to deal with this one properly.

avik-pal:
> if I understand correctly, then the current recommendation would be to define a custom linear solver wrapper, as I proposed

I meant having the preamble function in the `GeneralizedFirstOrder...` algorithm itself. So you create your newton's method as:

```julia
descent = NewtonDescent(; linsolve, precs)
GeneralizedFirstOrderAlgorithm(; concrete_jac, name = :NewtonRaphson,
    linesearch, descent, jacobian_ad = autodiff, preamble = <....>)
```

avik-pal:
That said I am anyways not adding any new functionality in the PR. Most of what is being added is what was incomplete or inconsistent with docs and such. @termi-official can you open a separate issue for this and link this PR, I will open a follow up PR for that?

ChrisRackauckas:
> I am confident that other frameworks come with similar mechanisms. Basically the difference to a "normal" Newton-Raphson is the apply! before entering the iteration to e.g. set the boundary condition in the solution vector and the 2x apply_zero!, first to e.g. eliminate the boundary conditions from the linear system and second to remove numerical inaccurracies in the solver from the increment. If it is not clear: The increment must be zero, because the solution is already enforced to be correct by the first call to apply!. For e.g. gradient based solvers we need similar hooks.

Those are just nonlinear preconditioners, with pre and post preconditioning.

avik-pal:
Once https://github.com/SciML/DiffEqBase.jl/pull/994 is merged, this would be good to go

avik-pal:
Let's go! Everything is working :tada:

avik-pal:
battery_problem_work_precision
Things that were not working before are also working now

docs/src/basics/faq.md
```diff
@@ -90,3 +91,61 @@ end
...
+```@example type_unstable
+@code_warntype solve(prob, NewtonRaphson(; autodiff = AutoForwardDiff(; chunksize = 2)))
+nothing # hide
+```
+
+And boom! Type stable again. For selecting the chunksize the method is:
+
+ 1. For small inputs `≤ 12` use `chunksize = <length of input>`
+ 2. For larger inputs, use `chunksize = 12`
+
```
ChrisRackauckas, line 149: Maybe just recommend `ForwardDiff.chunksize`?

ChrisRackauckas: Or, since that's not public API, we can just make `NonlinearSolve.chunksize` which just calls the ForwardDiff one and is tested so if it moves we just keep this stable.

docs/src/basics/solve.md
```diff
@@ -8,14 +8,19 @@ solve(prob::SciMLBase.NonlinearProblem, args...; kwargs...)
...
 
   - `maxiters::Int`: The maximum number of iterations to perform. Defaults to `1000`.
-  - `abstol::Number`: The absolute tolerance. Defaults to `real(oneunit(T)) * (eps(real(one(T))))^(4 // 5)`.
-  - `reltol::Number`: The relative tolerance. Defaults to `real(oneunit(T)) * (eps(real(one(T))))^(4 // 5)`.
+  - `maxtime`: The maximum time for solving the nonlinear system of equations. Defaults to
+    `nothing` which means no time limit. Note that setting a time limit does have a small
+    overhead.
+  - `abstol::Number`: The absolute tolerance. Defaults to
+    `real(oneunit(T)) * (eps(real(one(T))))^(4 // 5)`.
+  - `reltol::Number`: The relative tolerance. Defaults to
```
ChrisRackauckas, line 22: These can be vectors matching the size as well?

docs/src/native/simplenonlinearsolve.md
```diff
@@ -0,0 +1,59 @@
...
+## General Methods
+
+These methods are suited for any general nonlinear root-finding problem, i.e.
+`NonlinearProblem`.
+
+| Solver                               | In-place | Out of Place | Non-Allocating (Scalars) | Non-Allocating (`SArray`) |
+|:------------------------------------ |:-------- |:------------ |:------------------------ |:------------------------- |
+| [`SimpleNewtonRaphson`](@ref)        | ✔️       | ✔️           | ✔️                       | ✔️                        |
+| [`SimpleBroyden`](@ref)              | ✔️       | ✔️           | ✔️                       | ✔️                        |
+| [`SimpleHalley`](@ref)               | ❌        | ✔️           | ✔️                       | ❌                         |
```
ChrisRackauckas, line 32:
```suggestion
| [`SimpleHalley`](@ref)               | ❌       | ✔️           | ✔️                       | ❌                         |
```

docs/src/native/simplenonlinearsolve.md
```diff
@@ -0,0 +1,59 @@
...
+## General Methods
+
+These methods are suited for any general nonlinear root-finding problem, i.e.
+`NonlinearProblem`.
+
+| Solver                               | In-place | Out of Place | Non-Allocating (Scalars) | Non-Allocating (`SArray`) |
+|:------------------------------------ |:-------- |:------------ |:------------------------ |:------------------------- |
+| [`SimpleNewtonRaphson`](@ref)        | ✔️       | ✔️           | ✔️                       | ✔️                        |
+| [`SimpleBroyden`](@ref)              | ✔️       | ✔️           | ✔️                       | ✔️                        |
+| [`SimpleHalley`](@ref)               | ❌       | ✔️           | ✔️                       | ❌                         |
```
ChrisRackauckas, line 32: ugh why is this one odd 😅

docs/src/solvers/nonlinear_system_solvers.md
```diff
@@ -0,0 +1,163 @@
...
+difficult yet need high performance, and  [`TrustRegion`](@ref) is a bit less performant but
+more stable. If the problem is well-conditioned, [`Klement`](@ref) or [`Broyden`](@ref) may
+be faster, but highly dependent on the eigenvalues of the Jacobian being sufficiently small.
+
+[`NewtonRaphson`](@ref) and [`TrustRegion`](@ref) are designed for for large systems. They
+can make use of sparsity patterns for sparse automatic differentiation and sparse linear
+solving of very large systems. Meanwhile, [`SimpleNewtonRaphson`](@ref) and
+[`SimpleTrustRegion`](@ref) are implementations which are specialized for small equations.
+They are non-allocating on static arrays and thus really well-optimized for small systems,
+thus usually outperforming the other methods when such types are used for `u0`.
```
ChrisRackauckas, line 34: Mention GPUs as well? That should probably get a tutorial too. I think I have left a blank tutorial on static compilation for that kind of stuff.

docs/src/solvers/nonlinear_system_solvers.md
```diff
@@ -0,0 +1,163 @@
...
+### SteadyStateDiffEq.jl
+
+SteadyStateDiffEq.jl uses ODE solvers to iteratively approach the steady state. It is a
+very stable method for solving nonlinear systems with stable equilibrium points, though
+often more computationally expensive than direct methods.
+
+  - [`DynamicSS()`](@ref): Uses an ODE solver to find the steady state. Automatically
+    terminates when close to the steady state.
+  - [`SSRootfind()`](@ref): Uses a NonlinearSolve compatible solver to find the steady
+    state.
```
ChrisRackauckas, line 120: This is effectively deprecated for just using NonlinearSolve

docs/src/solvers/nonlinear_system_solvers.md
```diff
@@ -0,0 +1,163 @@
...
+Submethod choices for this algorithm include:
+
+  - `:anderson`: Anderson-accelerated fixed-point iteration
+  - `:newton`: Classical Newton method with an optional line search
+  - `:trust_region`: Trust region Newton method (the default choice)
+
+### MINPACK.jl
+
+MINPACK.jl methods are good for medium-sized nonlinear solves. It does not scale due to
+the lack of sparse Jacobian support, though the methods are very robust and stable.
```
ChrisRackauckas, line 138: Are they? The algorithms aren't showing that in comparison 😅

docs/src/solvers/nonlinear_system_solvers.md
```diff
@@ -0,0 +1,163 @@
...
+implementations.
+
+  - [`KINSOL()`](@ref): The KINSOL method of the SUNDIALS C library
+
+### SIAMFANLEquations.jl
+
+SIAMFANLEquations.jl is a wrapper for the methods in the SIAMFANLEquations.jl library.
+
+  - [`SIAMFANLEquationsJL()`](@ref): A wrapper for using the methods in
+    [SIAMFANLEquations.jl](https://github.com/ctkelley/SIAMFANLEquations.jl)
```
ChrisRackauckas, line 163: Some of the newer wrappers are missing from this list. SpeedMapping.jl, LeastSquaresOptim.jl, FastLevenburgMarquedt, etc.

docs/src/solvers/steady_state_solvers.md
```diff
@@ -0,0 +1,68 @@
...
+
+The SteadyStateDiffEq.jl methods on a [`SteadyStateProblem`](@ref) respect the time
+definition in the nonlinear definition, i.e., `u' = f(u, t)` uses the correct values for
+`t` as the solution evolves. A conversion of a [`SteadyStateProblem`](@ref) to a
+[`NonlinearProblem`](@ref) replaces this with the nonlinear system `u' = f(u, ∞)`, and thus
+the direct [`SteadyStateProblem`](@ref) approach can give different answers (i.e., the
+correct unique fixed point) on ODEs with non-autonomous dynamics.
+
+If you have an unstable equilibrium and you want to solve for the unstable equilibrium,
+then [`DynamicSS`](@ref) might converge to the equilibrium based on the initial condition.
```
ChrisRackauckas, line 26:
```suggestion
then [`DynamicSS`](@ref) will not converge to that equilibrium for any initial condition.
```

docs/src/tutorials/large_systems.md
```diff
@@ -137,11 +137,14 @@ Symbolic Sparsity Detection. See the manual entry on
...
 @btime solve(prob_brusselator_2d,
-    NewtonRaphson(; autodiff = AutoSparseForwardDiff(), linsolve = KLUFactorization()));
+    NewtonRaphson(; autodiff = AutoSparseForwardDiff(; chunksize = 32)));
 @btime solve(prob_brusselator_2d,
-    NewtonRaphson(; autodiff = AutoSparseForwardDiff(), linsolve = KrylovJL_GMRES()));
+    NewtonRaphson(; autodiff = AutoSparseForwardDiff(; chunksize = 32),
+        linsolve = KLUFactorization()));
+@btime solve(prob_brusselator_2d,
+    NewtonRaphson(; autodiff = AutoSparseForwardDiff(; chunksize = 32),
+        linsolve = KrylovJL_GMRES()));
```
ChrisRackauckas, line 147: That's a huge chunk size. Is it actually more helpful than 7?

ext/NonlinearSolveLeastSquaresOptimExt.jl
```diff
@@ -4,45 +4,49 @@ using NonlinearSolve, SciMLBase
...
 
-function SciMLBase.__init(prob::NonlinearLeastSquaresProblem, alg::LeastSquaresOptimJL,
-        args...; alias_u0 = false, abstol = nothing, show_trace::Val{ShT} = Val(false),
-        trace_level = TraceMinimal(), store_trace::Val{StT} = Val(false), maxiters = 1000,
-        reltol = nothing, kwargs...) where {ShT, StT}
-    iip = SciMLBase.isinplace(prob)
-    u = NonlinearSolve.__maybe_unaliased(prob.u0, alias_u0)
+function SciMLBase.reinit!(cache::LeastSquaresOptimJLCache, args...; kwargs...)
+    error("Reinitialization not supported for LeastSquaresOptimJL.")
+end
```
ChrisRackauckas, line 28: why only this one but not the other extensions?

src/abstract_types.jl
```diff
@@ -0,0 +1,471 @@
...
+### `SciMLBase.init` specification
+
+```julia
+SciMLBase.init(prob::AbstractNonlinearProblem,
+    alg::AbstractNonlinearSolveLineSearchAlgorithm, f::F, fu, u, p, args...;
+    internalnorm::IN = DEFAULT_NORM,
+    kwargs...) where {F, IN} --> AbstractNonlinearSolveLineSearchCache
+```
+"""
+abstract type AbstractNonlinearSolveLineSearchAlgorithm end
```
ChrisRackauckas, line 124: Wasn't this moved to LineSearch.jl?

src/algorithms/klement.jl
```diff
@@ -0,0 +1,146 @@
+"""
+    Klement(; max_resets = 100, linsolve = NoLineSearch(), linesearch = nothing,
+        precs = DEFAULT_PRECS, alpha = nothing, init_jacobian::Val = Val(:identity),
+        autodiff = nothing)
+
+An implementation of `Klement` [klement2014using](@citep) with line search, preconditioning
+and customizable linear solves. It is recommended to use [`Broyden`](@ref) for most problems
+over this.
```
ChrisRackauckas, line 8: Is that still true? Since the fast polyalg defaults here.

src/core/generalized_first_order.jl
```diff
@@ -0,0 +1,279 @@
...
+            u_cache, p, du, J, alg, prob, jac_cache, descent_cache, linesearch_cache,
+            trustregion_cache, 0, 0, maxiters, maxtime, alg.max_shrink_times, timer, 0.0,
+            true, termination_cache, trace, ReturnCode.Default, false)
+    end
+end
+
+function __step!(cache::GeneralizedFirstOrderAlgorithmCache{iip, GB};
+        recompute_jacobian::Union{Nothing, Bool} = nothing, kwargs...) where {iip, GB}
+    @static_timeit cache.timer "jacobian" begin
+        if (recompute_jacobian === nothing || recompute_jacobian) && cache.make_new_jacobian
```
ChrisRackauckas, line 204: So what's the plan on the ODE solver for triggering new Jacobians? I didn't see the high level function that would be involved. I assume it's not just flipping `make_new_jacobian`?

src/default.jl
```diff
@@ -180,38 +182,18 @@ or more precision / more stable linear solver choice is required).
...
-    [LinearSolve.jl documentation](https://docs.sciml.ai/LinearSolve/stable/).
 """
 function RobustMultiNewton(::Type{T} = Float64; concrete_jac = nothing, linsolve = nothing,
         precs = DEFAULT_PRECS, autodiff = nothing) where {T}
     if __is_complex(T)
         # Let's atleast have something here for complex numbers
         algs = (NewtonRaphson(; concrete_jac, linsolve, precs, autodiff),)
     else
-        algs = (TrustRegion(; concrete_jac, linsolve, precs, autodiff),
+        algs = (TrustRegion(; concrete_jac, linsolve, precs),
```
ChrisRackauckas, line 192: why remove that autodiff here?

src/descent/dogleg.jl
```diff
@@ -0,0 +1,136 @@
...
+    normal_form = prob isa NonlinearLeastSquaresProblem &&
+                  __needs_square_A(alg.newton_descent.linsolve, u)
+    JᵀJ_cache = !normal_form ? J * _vec(δu) : nothing  # TODO: Rename
+
+    return DoglegCache{INV, normal_form}(δu, δus, newton_cache, cauchy_cache, internalnorm,
+        JᵀJ_cache, δu_cache_1, δu_cache_2, δu_cache_mul)
+end
+
+# If TrustRegion is not specified, then use a Gauss-Newton step
+function SciMLBase.solve!(cache::DoglegCache{INV, NF}, J, fu, u, idx::Val{N} = Val(1);
```
ChrisRackauckas, line 78: I don't know about putting these all on the main init and solve because they are meant to be internal?

src/internal/helpers.jl
```diff
@@ -0,0 +1,262 @@
...
+        check_reverse_mode = true, kwargs...) where {test_sparse}
+    if check_reverse_mode
+        @warn "$(autodiff)::$(typeof(autodiff)) is not a \
+               `Abstract(Forward/FiniteDifferences)Mode`. Use with caution." maxlog=1
+    end
+    return autodiff
+end
+function get_concrete_forward_ad(autodiff, prob, sp::Val{test_sparse} = True, args...;
+        kwargs...) where {test_sparse}
+    # TODO: Default to PolyesterForwardDiff for non sparse problems
```
ChrisRackauckas, line 56: Make an issue?

src/internal/helpers.jl
```diff
@@ -0,0 +1,262 @@
...
+        kwargs...) where {test_sparse}
+    if check_reverse_mode
+        @warn "$(autodiff)::$(typeof(autodiff)) is not a \
+               `Abstract(Forward/FiniteDifferences)Mode`. Use with caution." maxlog=1
+    end
+    return autodiff
+end
+function get_concrete_reverse_ad(autodiff, prob, sp::Val{test_sparse} = True, args...;
+        kwargs...) where {test_sparse}
+    # TODO: Default to Enzyme / ReverseDiff for inplace problems?
```
ChrisRackauckas, line 99: Make an issue?

src/internal/jacobian.jl
```diff
@@ -0,0 +1,191 @@
...
+
+# Sparsity Detection Choices
+@inline __sparsity_detection_alg(_, _) = NoSparsityDetection()
+@inline function __sparsity_detection_alg(f::NonlinearFunction, ad::AbstractSparseADType)
+    if f.sparsity === nothing
+        if f.jac_prototype === nothing
+            if is_extension_loaded(Val(:Symbolics))
+                return SymbolicsSparsityDetection()
+            else
+                return ApproximateJacobianSparsity()
```
ChrisRackauckas, line 150: As a default?

ChrisRackauckas:
A bunch of small comments and issues to open, otherwise good to go!

src/default.jl
avik-pal: Accidental will fix this

src/core/generalized_first_order.jl
avik-pal: There are 2 possibilities:

1. the ODE solver uses the iterator form, and passes in `recompute_jacobian = false` if it wants to prevent a new jacobian construction
2. we have an inbuilt SingleFactorizeNewton which is similar to KINSOL and sets the make_new_jacobian internally.

src/algorithms/klement.jl
avik-pal: Klement is a very interested solver. If it works it converges faster than Broyden but it works in very limited situations. We can consider removing it from the polyalgorithm.

src/abstract_types.jl
avik-pal: Not yet, but that is the eventual plan. We still need to figure out how to decouple the line searches like `RobustNonMonotoneLineSearch` which heavily ties into the cache.

docs/src/basics/solve.md
avik-pal: No that is not implemented.

src/internal/jacobian.jl
avik-pal: Symbolics is actually pretty good after a size threshold, and even faster than the approximate version.

docs/src/tutorials/large_systems.md
avik-pal: In this case it actually gave a speedup locally till 32 but larger than this made the code slow.

src/internal/helpers.jl
avik-pal: https://github.com/SciML/NonlinearSolve.jl/issues/353

src/internal/helpers.jl
avik-pal: https://github.com/SciML/NonlinearSolve.jl/issues/353

docs/src/basics/faq.md
avik-pal: Done, naming it `pickchunksize` similar to what ForwardDiff does

docs/src/solvers/nonlinear_system_solvers.md
avik-pal: Mentioned it, I might do a section on GPU compilation for the paper, I will port that over when done.

docs/src/solvers/nonlinear_system_solvers.md
avik-pal: Yeah. not really. Changed the text.

docs/src/solvers/nonlinear_system_solvers.md
avik-pal: Linked them from here to avoid repeating the text.

docs/src/solvers/steady_state_solvers.md
avik-pal: Done

docs/src/native/simplenonlinearsolve.md
avik-pal: I don't think so, the initial formatting came from my default markdown formatted in vscode.

avik-pal: actually JuliaFormatter formats it back to the original one before your commit.

src/descent/dogleg.jl
avik-pal: makes sense. dispatching on `__internal_init` and `__internal_solve!`

avik-pal:
the dae failure is real https://github.com/SciML/NonlinearSolve.jl/actions/runs/7543078285/job/20533293590?pr=345#step:6:1759. The cache is initialized with a Vector but later it gets a subarray

src/internal/jacobian.jl
ChrisRackauckas: I mean defaulting to approximate Jacobian sparsity seems a bit odd. Even if it's not AD?

src/core/generalized_first_order.jl
ChrisRackauckas: > the ODE solver uses the iterator form, and passes in recompute_jacobian = false if it wants to prevent a new jacobian construction

Oh I missed that. That works.

docs/src/basics/solve.md
ChrisRackauckas: We should make an issue on this then. It just needs to use broadcast?

ChrisRackauckas (approved):
Looks like my comments were addressed. Let's get some issues open on nonlinear preconditioning, Klement in the polyalg, linesearch removal, and non-scalar tolerances.

ChrisRackauckas:
> the dae failure is real https://github.com/SciML/NonlinearSolve.jl/actions/runs/7543078285/job/20533293590?pr=345#step:6:1759. The cache is initialized with a Vector but later it gets a subarray

We should just make the cache via a view too.

ChrisRackauckas:
Make an issue in OrdinaryDiffEq with the DAE one, I can take a look at that later today.

PR #347, opened 2024-01-05
make RobustMultiNewton always respect autodiff choice

oscardssmith:
This probably needs a test, but the TLDR is that it fixes use of `RobustMultiNewton` with autodiff incompatible functions. Before `TrustRegion(; concrete_jac, linsolve, precs)` would mean that it would throw duals at the problem even if you specified `autodiff=AutoFiniteDiff()`


avik-pal (approved):

avik-pal:
@oscardssmith can you rerun the formatter once?

ChrisRackauckas (approved):

ChrisRackauckas:
And add a test

oscardssmith:
Formatted and tests added.

PR #348, opened 2024-01-06
CI trigger

ArnoStrouwen:
Tests are not passing locally for me, so triggering tests

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -13,7 +13,7 @@ concurrency:
   cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


ArnoStrouwen:
These ones take much longer (stuck indefinitely?) on windows/mac compared to linux:
https://github.com/SciML/NonlinearSolve.jl/actions/runs/7430099659/job/20219416626?pr=348
This one fails on mac:
https://github.com/SciML/NonlinearSolve.jl/actions/runs/7430099659/job/20219417053?pr=348
Then there are a couple groups that fail all os.
@avik-pal can you take a look?

avik-pal:
Let's hope #345 fixes all of these. I don't have a mac or windows to test locally.

PR #354, opened 2024-01-16
AD Section in Documentation

avik-pal:
## TODOs

- [x] Merge https://github.com/SciML/NonlinearSolve.jl/pull/345
- [x] Finish https://github.com/JuliaDiff/SparseDiffTools.jl/pull/281
- [x] Add a section on AD tips in the documentation

## Good Old Brusselator Benchmark

```julia
julia> @btime solve(prob_brusselator_2d, NewtonRaphson());
  402.717 ms (6959 allocations: 64.87 MiB)

julia> using PolyesterForwardDiff

julia> @btime solve(prob_brusselator_2d, NewtonRaphson());
  348.450 ms (333 allocations: 77.20 MiB)
```

Not changing the default, here, but mentioned in the docs on how to use polyester.


avik-pal:
I want to verify that this is better in most cases before switching the default

docs/src/basics/autodiff.md
```diff
@@ -0,0 +1,51 @@
...
+  - [`AutoZygote`](@ref): The fastest choice for non-mutating array-based (BLAS) functions.
+  - [`AutoSparseZygote`](@ref): Sparse version of [`AutoZygote`](@ref).
+  - [`AutoEnzyme`](@ref): Uses `Enzyme.jl` Reverse Mode and should be considered
+    experimental.
+
+!!! note
+    
+    If `PolyesterForwardDiff.jl` is installed and loaded, then `SimpleNonlinearSolve.jl`
+    will automatically use `AutoPolyesterForwardDiff` as the default AD backend.
+
```
ChrisRackauckas, line 26:
```suggestion
!!! note
    
    If `PolyesterForwardDiff.jl` is installed and loaded, then `SimpleNonlinearSolve.jl`
    will automatically use `AutoPolyesterForwardDiff` as the default AD backend.

!!! note

    The `sparse` versions of the methods refers to automated sparsity detection. These methods
    automatically discover the sparse Jacobian form from the function `f`. Note that all methods
    specialize the differentiation on a sparse Jacobian if the sparse Jacobian is given as
    `prob.f.jac_prototype` in the `NonlinearFunction` definition, and the `autosparse` here simply
    refers to whether this `jac_prototype` should be generated automatically. Note too that
    this sparsity detection has a non-trivial overhead and thus as an optimization it's recommended
    that one runs the sparsity detection outside of the solver loop if many solves are required.
    For more details, see [SparseDiffTools.jl](https://github.com/JuliaDiff/SparseDiffTools.jl)
```

avik-pal: Should we link https://docs.sciml.ai/NonlinearSolve/stable/basics/sparsity_detection/ instead of SparseDiffTools directly?

avik-pal: If `AutoForwardDiff` is specified and `sparsity` is specified, we still use non-color AD. However, if the `autodiff = nothing`, we will choose `AutoSparseForwardDiff`. So the note "specialize the differentiation on a sparse Jacobian if the sparse Jacobian" seems a bit incorrect.

The reason to do this is in the last warning of https://docs.sciml.ai/NonlinearSolve/stable/basics/sparsity_detection/ (mostly to allow an external override of any default internal selection algorithm)

avik-pal: not for a performance consideration, it is to allow an external override. If only jac_prototype is specified, then sparse is used. But if the user manually decides to use AutoForwardDiff overriding the default autodiff choice, then I feel we should use the autodiff provided.

ChrisRackauckas: I guess so, though that's an odd combination 😅. Can you reword this and merge?

avik-pal:
@oscardssmith do you know what https://github.com/SciML/NonlinearSolve.jl/actions/runs/7786112451/job/21230175067?pr=354#step:6:14951 is? It seems to come from the `Base.Experimental.@opaque`, which I think you added.

oscardssmith:
It is. It means you are trying to AD a function even though the test specifically tells you not to AD it.

Issue #360, opened 2024-01-30
`TrustRegion(radius_update_scheme=RadiusUpdateSchemes.Bastin)` doesn't respect autodiff setting .


oscardssmith:
When solving a nonlinear problem that doesn't support autodiff, the Bastin update scheme uses autodiff anyways. This is especially problematic because Bastin is used in the default algorithm.

**Minimal Reproducible Example 👇**

Without MRE, we would only be able to help you to a limited extent, and attention to the issue would be limited. to know more about MRE refer to [wikipedia](https://en.wikipedia.org/wiki/Minimal_reproducible_example) and [stackoverflow](https://stackoverflow.com/help/minimal-reproducible-example).

```julia
using NonlinearSolve
g(u::Vector{Float64},p) = u
p = NonlinearProblem(g, [1.])
solve(p, TrustRegion(autodiff=AutoFiniteDiff(), radius_update_scheme=RadiusUpdateSchemes.Bastin))
```

**Error & Stacktrace ⚠️**

```julia
ERROR: MethodError: no method matching g(::Vector{ForwardDiff.Dual{ForwardDiff.Tag{…}, Float64, 1}}, ::SciMLBase.NullParameters)
An arithmetic operation was performed on a NullParameters object. This means no parameters were passed
into the AbstractSciMLProblem (e.x.: ODEProblem) but the parameters object `p` was used in an arithmetic
expression. Two common reasons for this issue are:

1. Forgetting to pass parameters into the problem constructor. For example, `ODEProblem(f,u0,tspan)` should
be `ODEProblem(f,u0,tspan,p)` in order to use parameters.

2. Using the wrong function signature. For example, with `ODEProblem`s the function signature is always
`f(du,u,p,t)` for the in-place form or `f(u,p,t)` for the out-of-place form. Note that the `p` argument
will always be in the function signature regardless of if the problem is defined with parameters!


ERROR: MethodError: no method matching g(::Vector{ForwardDiff.Dual{ForwardDiff.Tag{…}, Float64, 1}}, ::SciMLBase.NullParameters)

Closest candidates are:
  g(::Vector{Float64}, ::Any)
   @ Main REPL[65]:1

Stacktrace:
  [1] (::NonlinearFunction{…})(::Vector{…}, ::Vararg{…})
    @ SciMLBase ~/.julia/packages/SciMLBase/bp17m/src/scimlfunctions.jl:2356
[...]

PR #361, opened 2024-01-31
fix `autodiff=AutoFiniteDiff()` for `TrustRegion`

oscardssmith:
## Checklist
fixes https://github.com/SciML/NonlinearSolve.jl/issues/360

### DIFF
--- a/src/algorithms/trust_region.jl
+++ b/src/algorithms/trust_region.jl
@@ -24,9 +24,13 @@ function TrustRegion(; concrete_jac = nothing, linsolve = nothing, precs = DEFAU
         initial_trust_radius::Real = 0 // 1, step_threshold::Real = 1 // 10000,
         shrink_threshold::Real = 1 // 4, expand_threshold::Real = 3 // 4,
         shrink_factor::Real = 1 // 4, expand_factor::Real = 2 // 1,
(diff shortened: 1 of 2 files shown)


src/algorithms/trust_region.jl
```diff
@@ -24,9 +24,13 @@ function TrustRegion(; concrete_jac = nothing, linsolve = nothing, precs = DEFAU
         initial_trust_radius::Real = 0 // 1, step_threshold::Real = 1 // 10000,
         shrink_threshold::Real = 1 // 4, expand_threshold::Real = 3 // 4,
         shrink_factor::Real = 1 // 4, expand_factor::Real = 2 // 1,
-        max_shrink_times::Int = 32, vjp_autodiff = nothing, autodiff = nothing)
+        max_shrink_times::Int = 32, autodiff = nothing, vjp_autodiff = autodiff)
```
avik-pal, line 27: This will cause problems if `autodiff = AutoForwardDiff()`. The nicer way would be to keep `vjp_autodiff = nothing` and check that if `autodiff` is FiniteDiff then set vjp_autodiff to that.

oscardssmith: makes sense. Updated.

avik-pal (approved):

PR #382, opened 2024-02-20
Add caching for solvers without init

avik-pal:
This is needed because we want to write generic code with `init` and `solve!` without having to query every solver to check if it supports caching.

### DIFF
--- a/src/core/noinit.jl
+++ b/src/core/noinit.jl
@@ -1,0 +1,37 @@
+# Some algorithms don't support creating a cache and doing `solve!`, this unfortunately
+# makes it difficult to write generic code that supports caching. For the algorithms that
+# don't have a `__init` function defined, we create a "Fake Cache", which just calls
+# `__solve` from `solve!`
+@concrete mutable struct NonlinearSolveNoInitCache{iip, timeit} <:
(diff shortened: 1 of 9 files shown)


avik-pal:
I will do a little bit of type piracy for SimpleNonlinearSolve to avoid having to copy over this code there

src/core/noinit.jl
```diff
@@ -0,0 +1,37 @@
+# Some algorithms don't support creating a cache and doing `solve!`, this unfortunately
+# makes it difficult to write generic code that supports caching. For the algorithms that
+# don't have a `__init` function defined, we create a "Fake Cache", which just calls
+# `__solve` from `solve!`
+@concrete mutable struct NonlinearSolveNoInitCache{iip, timeit} <:
+                         AbstractNonlinearSolveCache{iip, timeit}
+    prob
+    alg
+    args
+    kwargs::Any
```
ChrisRackauckas, line 10: does this effect type inference? I don't see why this wouldn't be concrete as well.

ChrisRackauckas:
> I will do a little bit of type piracy for SimpleNonlinearSolve to avoid having to copy over this code there.

That seems reasonable.

PR #385, opened 2024-02-26
Make default polyalgs respect autodiff

ChrisRackauckas:


ChrisRackauckas:
Test failure unrelated. @avik-pal can you handle that? It should just be a quick ambiguity fix

oscardssmith:
Why wasn't this caught by the `@testitem "No AD" begin` tests?

PR #392, opened 2024-03-21
Handle polyalgorithm aliasing correctly

avik-pal:
Aliasing is messy :sweat:


ChrisRackauckas:
Yes, I want a whole aliasing system in general, but this'll do for now.

PR #395, opened 2024-03-26
Add a wrapper for Optimization.jl

avik-pal:
Unfortunately, this introduces a massive circular dependency if Optimization is loaded. But we only pay the price in CI rn.


ChrisRackauckas:
Why not just add this code to OptmizationBase and document it here?

avik-pal:
That's probably a better way I will do that

PR #404, opened 2024-04-08
Add Halley's method via descent API

tansongchen:
Add Halley's method by computing Hessian-vector-vector product with TaylorDiff.jl in $O(n)$ time. Note that this isn't structured as something like "HessianCache", but instead structured as a descent method, which can be viewed as "adding some correction term to the Newton's method". This is mainly because Hessian, either implicitly or explicitly, isn't likely to be used elsewhere.

Tests haven't been added because TaylorDiff.jl should be added to an extension. Once that is done, it is straightforward to add tests.

## Checklist


src/descent/halley.jl
```diff
@@ -0,0 +1,87 @@
...
+        @static_timeit cache.timer "linear solve 1" begin
+            δu = cache.lincache(;
+                A = J, b = _vec(fu), kwargs..., linu = _vec(δu), du = _vec(δu),
+                reuse_A_if_factorization = !new_jacobian || (idx !== Val(1)))
+            δu = _restructure(get_du(cache, idx), δu)
+        end
+    end
+    b = cache.b
+    # compute the hessian-vector-vector product
+    hvvp = derivative(x -> cache.f(x, cache.p), u, δu, 2)
```
avik-pal, line 71:
```suggestion
    hvvp = derivative(Base.Fix2(cache.f, cache.p), u, δu, 2)
```

avik-pal: This will also need a case for inplace `cache.f`

src/descent/halley.jl
```diff
@@ -0,0 +1,87 @@
+"""
+    HalleyDescent(; linsolve = nothing, precs = DEFAULT_PRECS)
+
+Compute the descent direction as ``J δu = -fu``. For non-square Jacobian problems, this is
+commonly referred to as the Gauss-Newton Descent.
```
avik-pal, line 5: Old docs

src/algorithms/halley.jl
```diff
@@ -0,0 +1,12 @@
+"""
+    Halley(; concrete_jac = nothing, linsolve = nothing, linesearch = NoLineSearch(),
+        precs = DEFAULT_PRECS, autodiff = nothing)
+
+An experimental Halley's method implementation.
```
avik-pal, line 5: Ig you will change this later. But keeping it marked so that it doesn't fall through.

src/descent/halley.jl
tansongchen: Got it. TaylorDiff.jl haven't got a in-place mode yet, but will do this week

tansongchen:
@avik-pal Support for in-place `f` and tests are added. Let me know if you have any questions!

I think there isn't an eqivalent for `Fix2` that fixes the third argument, which is needed by evaluating the derivatives given `p`. Is it ok to use an closure?

avik-pal:
> I think there isn't an eqivalent for Fix2 that fixes the third argument, which is needed by evaluating the derivatives given p. Is it ok to use a closure?

Yes mark it as `@closure` and do it

src/descent/halley.jl
```diff
@@ -0,0 +1,98 @@
...
+    @bb @. δu = δu * δu / (b / 2 - δu)
+    set_du!(cache, δu, idx)
+    cache.b = b
+    return δu, true, (;)
+end
+
+function evaluate_hvvp(
+        cache::HalleyDescentCache, f::NonlinearFunction{iip}, p, u, δu) where {iip}
+    if iip
+        binary_f = (y, x) -> f(y, x, p)
```
avik-pal, line 92:
```suggestion
        binary_f = @closure (y, x) -> f(y, x, p)
```

test/misc/halley_tests.jl
```diff
@@ -0,0 +1,16 @@
...
+    # out-of-place
+    prob1 = NonlinearProblem(f, u0, p)
+    sol1 = solve(prob1, Halley())
+    @test sol1.u ≈ [sqrt(2.0), sqrt(2.0)]
+
+    # in-place
+    prob2 = NonlinearProblem(f!, u0, p)
+    sol2 = solve(prob2, Halley())
+    @test sol2.u ≈ [sqrt(2.0), sqrt(2.0)]
+end
```
avik-pal, line 16: Let's add a bit more comprehensive tests here similar to SimpleNonlinearSolve

tansongchen:
I am trying to add this as a block in `23_test_problems_tests.jl` so I can make use of these problems. Do you know how to run a `@testitem` independently? With the vscode built-in test runner, I am getting the following error:

image

avik-pal:
Yeah ReTestItems doesn't yet work with VSCode Test Runner. In a julia REPL activate the test using TestEnv then do

```julia
using ReTestItems

ReTestItems.runtests(<path to file or test directory>; name="<name of the testitem>")
```

tansongchen:
Ok I added it to this file near `NewtonRaphson`. Some cases haven't work mainly due to limitations of TaylorDiff.jl

tansongchen:
@avik-pal Do you have an out-of-place version of 23 problems? When testing with `SimpleHallely`, I hit

```
12:48:34 | START (1/1) test item "SimpleHalley" at test/core/23_test_problems_tests.jl:53
┌ Error: ErrorException("SimpleHalley currently only supports out-of-place nonlinear problems")
└ @ Main.var"##RobustnessTesting#247" ~/Public/SimpleNonlinearSolve.jl/test/core/23_test_problems_tests.jl:28
```

avik-pal:
Don't the tests in simplenonlinearsolve use the oop version?

tansongchen:
So currently, the tests at https://github.com/SciML/SimpleNonlinearSolve.jl/blob/main/test/core/23_test_problems_tests.jl doesn't include Halley's method, and when I tried to add Halley in the same way as Newton, I get the error above.

And the file at https://github.com/SciML/DiffEqProblemLibrary.jl/blob/master/lib/NonlinearProblemLibrary/src/NonlinearProblemLibrary.jl doesn't have out-of-place `f`'s

tansongchen:
Ok so I manually ran `[1, 5, 8, 15, 16]` against SimpleHalley, and 1, 5, 15, 16 also failed. 8 is due to an error on TaylorDiff.jl and has been fixed. I need to release a new version of TaylorDiff to make the CIs pass though.

tansongchen:
Should be fine now, @avik-pal could you review that again?

avik-pal:
Handle the test failures, the `Misc` ones are real

tansongchen:
Ok, tried to fix the precompilation issue and added compat

tansongchen:
Now all fails are unrelated

avik-pal:
Are you sure? There are no failures on master. Some weird interaction of Taylor diff with zygote?

tansongchen:
hmm, that's weird. The Zygote stuff in TaylorDiff doesn't have type piracy so shouldn't have weird interactions

avik-pal:
@tansongchen TaylorDiff is definitely causing problems. See https://github.com/SciML/NonlinearSolve.jl/pull/441

Project.toml
```diff
@@ -28,6 +28,7 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 SparseDiffTools = "47a9eef4-7e08-11e9-0b38-333d64bd3804"
 StaticArraysCore = "1e83bf80-4336-4d27-bf5d-d5a4f845583c"
 SymbolicIndexingInterface = "2efcf032-c050-4f8e-a9bb-153293bab1f5"
+TaylorDiff = "b36ab563-344f-407b-a36a-4f200bebf99c"
```
avik-pal, line 31: TaylorDiff brings in Symbolics, which seems like quite a heavy dep. We have quite a few downstream users notably OrdinaryDiffEq.

Won't this affect load times? @ChrisRackauckas @oscardssmith thoughts on this? Should we make it a weak dep

ChrisRackauckas: Can we make this be an extension alg in a lib like what we're doing in OrdinaryDiffEq.jl? NonlinearSolve.jl is going to have the same issue of just having too many algorithms at some point, and this seems like something that can be a lib extension

tansongchen:
@avik-pal Could you please review this again?

Comparing https://github.com/SciML/NonlinearSolve.jl/actions/runs/11667334463/job/32484400696 and https://github.com/SciML/NonlinearSolve.jl/actions/runs/11667812262/job/32486004697?pr=404 , I think this is not introducing new failures

Project.toml
```diff
@@ -25,6 +25,7 @@ PrecompileTools = "aea7be01-6a6a-4083-8856-8a6e6704d82a"
 Preferences = "21216c6a-2e73-6563-6e65-726566657250"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 SciMLBase = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
+SciMLJacobianOperators = "19f34311-ddf3-4b8b-af20-060888a46c0e"
```
avik-pal, line 28: needed?

lib/NonlinearSolveBase/ext/NonlinearSolveBaseTaylorDiffExt.jl
```diff
@@ -0,0 +1,20 @@
+module NonlinearSolveBaseTaylorDiffExt
+using SciMLBase: NonlinearFunction
+using NonlinearSolveBase: HalleyDescentCache
+import NonlinearSolveBase: evaluate_hvvp
+using TaylorDiff: derivative, derivative!
+using FastClosures: @closure
+
+function evaluate_hvvp(
```
avik-pal, line 8:
```suggestion
using NonlinearSolveBase: NonlinearSolveBase, HalleyDescentCache
using TaylorDiff: derivative, derivative!
using FastClosures: @closure

function NonlinearSolveBase.evaluate_hvvp(
```

style nit

lib/NonlinearSolveFirstOrder/src/halley.jl
```diff
@@ -0,0 +1,15 @@
...
+
+Currently depends on TaylorDiff.jl to handle the correction terms,
+might have more general implementation in the future.
+"""
+function Halley(; concrete_jac = nothing, linsolve = nothing,
+        linesearch = missing, autodiff = nothing)
+    return GeneralizedFirstOrderAlgorithm(;
+        concrete_jac, name = :Halley, linesearch,
+        descent = HalleyDescent(; linsolve), autodiff)
+end
```
avik-pal, line 15: Not really a First Order method, we might want an additional split cc @ChrisRackauckas

avik-pal (changes requested):
Needs some testing aprt from the 23 test problems

PR #414, opened 2024-04-23
Robust Singular Jacobian Handling

avik-pal:
Current approach is:

1. Proceed as usual
2. On first linear solve failure:
    1. Construct a QRPivoted Cache and cache it
    2. Solve using Pivoted QR but warn the user that there is a potential singular Jacobian detected
    3. If successful, proceed as usual
    4. Otherwise:
        1. For "current" jacobian: stop nonlinear solve
        2. For approximate methods: follow a policy similar to line search routines
        3. For non-current exact jacobian: force a reset
3. On later linear solve failure:
    1. Goto 2.2

## TODOs

- [x] Switching to Pivoted QR
- [x] Add tests
- [x] Use `verbose` to turn off printing
- [x] Integrate it correctly into the algorithms -- currently assumes that switching is always successful.
- [x] How do we deal with non-standard Arrays? Basically how do we determine that PivotedQR is defined for the matrix A
- [x] Add missing docstrings to the documentation

@ChrisRackauckas any idea how to handle this last part?


avik-pal:
With this PR all of the following work:

```julia
using NonlinearSolve, LinearSolve

function nlls!(du, u, p)
    du[1] = 2u[1] - 2
    du[2] = (u[1] - 4u[2])^2 + 0.1
    du[3] = 0
end

u0 = [0.0, 0.0]
prob = NonlinearLeastSquaresProblem(
    NonlinearFunction(nlls!, resid_prototype = zeros(3)), u0)

solve(prob)

solve(prob, GaussNewton(; linsolve = QRFactorization()))

solve(prob, GaussNewton(; linsolve = LUFactorization()))
```

I always display a warning rn, but I will configure the `verbose` to not show the warning if needed.

avik-pal:
Needs https://github.com/SciML/LinearSolve.jl/pull/494 before the tests pass

ChrisRackauckas (approved):

PR #437, opened 2024-05-25
Consolidate Stats Handling

### LINKED ISSUE #410: Consolidate `stats` handling
As pointed out by @oscardssmith, it would be good to store a centralized stats in the cache and update references to it instead of adding up the stats at the end.

Would probably make https://github.com/SciML/OrdinaryDiffEq.jl/pull/2167 easier

avik-pal:
Fixes #410

Oops lost my local changes it seems...

Merge after https://github.com/SciML/NonlinearSolve.jl/pull/435

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -32,7 +32,6 @@ TimerOutputs = "a759f4b9-e2f1-59dc-863e-4aeb61b1ea8f"
 
 [weakdeps]
 BandedMatrices = "aae01518-5342-5314-be14-df237901396f"
-Enlsip = "d5306a6b-d590-428d-a53a-eb3bb2d36f2d"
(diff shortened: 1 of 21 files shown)


avik-pal:
Now `cache.stats` gives the total counts correctly. For polyalgorithms this means we get the accumulated stats from all the solve calls.

avik-pal:
cc @oscardssmith because you needed this in OrdinaryDiffEq.

ChrisRackauckas (approved):

avik-pal:
Can you do it? I am not sure what broke, this was a purely internal API change

oscardssmith:
will do. The problem was that OrdinaryDiffEq with NonlinearSolveAlg runs the `step!(cache)` loops of a nonlinear solver itself and was using `nf` to determine how many calls the `step!` had made to update the integrator's stats.

PR #446, opened 2024-06-07
Add forward mode to line search

tansongchen:
## Checklist

## Additional context

Added forward mode support to line searches. We prefer forward AD for better performance, however, reverse AD is also supported if user explicitly requests it.

1. If jvp is available, we use forward AD;
2. If reverse type is requested, we use reverse AD;
3. Otherwise, we use forward AD.

### DIFF
--- a/test/core/rootfind_tests.jl
+++ b/test/core/rootfind_tests.jl
@@ -55,7 +55,7 @@ end
 @testitem "NewtonRaphson" setup=[CoreRootfindTesting] tags=[:core] timeout=3600 begin
     @testset "LineSearch: $(_nameof(lsmethod)) LineSearch AD: $(_nameof(ad))" for lsmethod in (
             Static(), StrongWolfe(), BackTracking(), HagerZhang(), MoreThuente()),
-        ad in (AutoFiniteDiff(), AutoZygote())
+        ad in (AutoForwardDiff(), AutoZygote(), AutoFiniteDiff())
 
         linesearch = LineSearchesJL(; method = lsmethod, autodiff = ad)
         u0s = ([1.0, 1.0], @SVector[1.0, 1.0], 1.0)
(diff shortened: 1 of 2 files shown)


tansongchen:
Getting significant perf improvement for in-place functions like

```julia
quadratic_f!(du, u, p) = (du .= u .* u .- p)
u0 = fill(1.0, 1000)
p = 2.0
prob = NonlinearProblem(quadratic_f!, u0, p)
algfwd = NewtonRaphson(; linesearch=LineSearchesJL(; method=LineSearches.HagerZhang()))
algrev = NewtonRaphson(; linesearch=LineSearchesJL(; method=LineSearches.HagerZhang(), autodiff=AutoZygote())) # fallback to finitediff
```

```
julia> @btime sol = solve(prob, algfwd)
  27.795 ms (265 allocations: 15.62 MiB)
Julia> @btime sol = solve(prob, algrev)
  50.685 ms (15732 allocations: 133.12 MiB)
```

test/core/rootfind_tests.jl
```diff
@@ -55,7 +55,7 @@ end
 @testitem "NewtonRaphson" setup=[CoreRootfindTesting] tags=[:core] timeout=3600 begin
     @testset "LineSearch: $(_nameof(lsmethod)) LineSearch AD: $(_nameof(ad))" for lsmethod in (
             Static(), StrongWolfe(), BackTracking(), HagerZhang(), MoreThuente()),
-        ad in (AutoFiniteDiff(), AutoZygote())
```
oscardssmith, line 58: we should probably keep testing `AutoFiniteDiff`

avik-pal: +1 Keep FiniteDiff, just add ForwardDiff

tansongchen: @avik-pal  If using FiniteDiff, should we consider using it on $\phi(\alpha)$, instead of on the gradient? i.e.,

$$
\phi'(\alpha_0)=\frac{\phi(\alpha_0+\varepsilon)-\phi(\alpha_0-\varepsilon)}{2}
$$

This should be as accurate and much faster than FiniteDiff-ing the gradient

avik-pal: yeah makes sense

tansongchen:
Updated strategy:

0. If FiniteDiff is requested, we use finite diff on $\phi$;
1. If jvp is available, we use forward AD;
2. If vjp is available, we use reverse AD;
3. If reverse type is requested, we use reverse AD;
4. Finally, we use forward AD.

avik-pal (approved):

PR #453, opened 2024-07-17
add dispatch to avoid method ambiguity

m-bossart:
Adds a dispatch for forward diff for:  SciML/SimpleNonlinearSolve.jl#153


m-bossart:
Method ambiguity came from a mistake in SciML/SimpleNonlinearSolve.jl#153 , don't need this after all

PR #438, opened 2024-05-25
Add support for BigFloat

avik-pal:
Builds on top of https://github.com/SciML/NonlinearSolve.jl/pull/437. Unblocks stuff like https://github.com/SciML/BoundaryValueDiffEq.jl/issues/43.

Needs https://github.com/SciML/MaybeInplace.jl/pull/9

- [x] Needs some tests

PR #457, opened 2024-08-19
Better BigFloat support

ErikQQY:
Continue https://github.com/SciML/NonlinearSolve.jl/pull/438 and https://github.com/SciML/BoundaryValueDiffEq.jl/pull/202

When handling with BigFloat u0, instead of specializing when creating a new array, we directly fill zeros everywhere.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -163,13 +163,5 @@ end
 
 function __similar(x, args...; kwargs...)
     y = similar(x, args...; kwargs...)
-    return __init_bigfloat_array!!(y)
-end
-
-function __init_bigfloat_array!!(x)
-    if ArrayInterface.can_setindex(x)
-        eltype(x) <: BigFloat && fill!(x, BigFloat(0))
-        return x
-    end
-    return x
+    return zero(y)
 end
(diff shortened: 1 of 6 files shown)


ChrisRackauckas (approved):

src/utils.jl
avik-pal, line 1: This will double the allocations once for `similar` and again a new `zero` array.

Also performance-wise this is not a good idea:

```julia
julia> x = rand(100000);

julia> @btime similar($x);
  108.556 ns (3 allocations: 781.31 KiB)

julia> @btime zero($x);
  96.660 μs (3 allocations: 781.31 KiB)
```

avik-pal: https://github.com/SciML/SciMLStyle#always-default-to-constructs-which-initialize-data says that it is okay to use `undef` if we ensure that the data is initialized before being returned, so I don't think it applies in this context.

ChrisRackauckas: Where are you zeroing it?

avik-pal: The `u` in solution object will be initialized right? If the user doesn't assign a value to an element of `u` inside the in place function that is also an undefined behavior.

PR #460, opened 2024-09-23
feat: add `SciMLJacobianOperators` package

### LINKED ISSUE #442: Relax MINPACK.jl compat
https://github.com/JuliaPackaging/Yggdrasil/pull/8817 (or one of those related PRs) seems to be caused a regression in CMINPACK support. Currently it is compat bounded.

avik-pal:
Everyone knows how bad SciMLOperators function operator is (and I really want to avoid touching that package). This pulls out the JacobianOperator and related functionality from NonlinearSolve.jl so that I can use in SimpleNonlinearSolve and LineSearch

fixes #442 

## TODOs

- [x] docs
- [x] tests


avik-pal:
@ChrisRackauckas I am spinning out the jacobian operator out of NonlinearSolve. Need to use it for LineSearch. It also provides a solution to the long standing vjp/jvp interface for SciMLOperators

oscardssmith:
should WOperator also be split out?

avik-pal:
> should WOperator also be split out?

This particular split is to avoid a circular dep issue between LineSearch and NonlinearSolve.

avik-pal:
ILU is broken (segfaults) even on master. Skipping that to get the build passing

PR #458, opened 2024-08-23
feat: `NonlinearSolveBase` + `BracketingNonlinearSolve` + `SimpleNonlinearSolve`

avik-pal:
First step in cutting down on load times. This is mostly independent of #456 

1. Move away from `DiffEqBase` to `NonlinearSolveBase` in the `Simple` packages.
2. Split `SimpleNonlinearSolve` into (`SimpleNonlinearSolve` has longer load times than `Lux` which sounds extremely wrong)
	1. `SimpleNonlinearSolve` -- retains its current functionality but drops DiffEqBase and loads NonlinearSolveBase and BracketingNonlinearSolve
	2. `BracketingNonlinearSolve`
3. What does `NonlinearSolveBase` have?
    1. Termination Conditions that were kind of incorrectly put (by me) into DiffEqBase
    2. Forward AD overloads via extensions
    3. Helper functions for norm that is used across NonlinearSolve
    4. Moved the definition of ImmutableNonlinearProblem from SimpleNonlinearSolve
5. What are the other sources of bad load time?
    1. DocStringExtensions -- https://github.com/JuliaDocs/DocStringExtensions.jl/pull/167
    2. ~DifferentiationInterface -- SparseArrays https://github.com/gdalle/DifferentiationInterface.jl/issues/416~
    4. ~FiniteDiff -- again SparseArrays~
    6. ~MaybeInplace -- again SparseArrays (fixed by https://github.com/SciML/MaybeInplace.jl/pull/11)~
    7. MLStyle via SciMLBase -- I have been told this is being looked into.
6. Pending downstream changes:
    1. ~DiffEqBase loads NonlinearSolveBase and adds `get_concrete_problem`~

There's a bunch of stuff in NonlinearSolve that can eventually be moved into base, but that is something to do later on.

## Tasklist

- [x] NonlinearSolveBase
  - [x] Base Package
  - [x] Testing
  - [x] AD selection preferences
  - [x] ForwardAD Primitives
    - [x] Rootfinding
    - [x] NLLS
- [x] BracketingNonlinearSolve
  - [x] All core algorithms
  - [x] Add tests
  - [x] ForwardAD support
- [x] SimpleNonlinearSolve
  - [x] move package over from the other repo
    - [x] Needs https://github.com/SciML/LineSearch.jl/pull/11
  - [x] Tests
  - [x] AD Support
    - [x] NLS
    - [x] NLLS
- [x] CI scripts
  - [x] scripts for the new subpackages
    - [x] NonlinearSolveBase
[...]

PR #462, opened 2024-09-24
use LinearSolve precs setup

oscardssmith:
This updates NonlinearSolve to use the new `precs` interface in LinearSolve.


avik-pal:
This would require a major version bump.

mpf:
See also this Julia Discourse [post](https://discourse.julialang.org/t/warning-when-using-nonlinearsolve-jl-with-linearsolve-jl-and-minres/119781?u=mpf01).

ChrisRackauckas:
Is this needed?

oscardssmith:
it's not needed, but will be nice (as it will bring the interface inline with LinearSolve and OrdinaryDiffEq@V7

ChrisRackauckas:
okay, I thought it was covered by another PR already.

oscardssmith:
https://github.com/SciML/NonlinearSolve.jl/pull/465 was a simpler fix that fixed the bug that led me to create this PR, but all that did was switch NonlinearSolve from a cursed diagonal preconditioner to a real diagonal preconditioner, whereas this PR removes the preconditioner from the NonlinearSolve side entirely since the LinearSolve side can do all the work.

oscardssmith:
@avik-pal if I rebase this, do you think we're ready to do the version bump required? I think this is a good change, but I don't really want to go through the rebase until you think we're conceptually ready to merge.

avik-pal:
not yet, we need https://github.com/SciML/NonlinearSolve.jl/pull/458 to be merged before doing a bump

avik-pal:
@oscardssmith we can get this in now

PR #470, opened 2024-10-02
feat: use DI for structured Jacobians

avik-pal:
Seems to work beautifully. Haven't benchmarked yet, but the important part here is to ensure structure is preserved, since the main cost in these workloads is that Linear solving will be bottlnecked if a banded matrix is converted into a sparse matrix.

### DIFF
--- a/src/internal/jacobian.jl
+++ b/src/internal/jacobian.jl
@@ -37,7 +37,6 @@ Construct a cache for the Jacobian of `f` w.r.t. `u`.
     stats::NLStats
     autodiff
     di_extras
-    sdifft_extras
 end
 
 function reinit_cache!(cache::JacobianCache{iip}, args...; p = cache.p,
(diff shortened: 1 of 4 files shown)


gdalle:
Structure is indeed preserved for decompression (on the SparseMatrixColorings [PR branch](https://github.com/gdalle/SparseMatrixColorings.jl/pull/132) at least).
However, with the current implementation, a sparse matrix and its transpose _will be allocated_ from the structured Jacobian nonetheless, as a preparation step for decompression.
This seems futile when the decompression target still has the same structure, but when it doesn't, this sparse encoding is actually useful for fast decompression. The scenario here would be someone passing `sparsity_pattern::Tridiagonal` but then calling `jacobian` on a `jac_prototype::SparseMatrixCSC`.
Ultimately, coloring is a one-time cost, whereas decompression is paid at every Jacobian computation. So in SparseMatrixColorings we choose to allocate a little more during coloring, and that way we speed up more decompression scenarios down the road.

src/internal/jacobian.jl
```diff
@@ -126,20 +103,19 @@ function JacobianCache(prob, alg, f::F, fu_, u, p; stats, autodiff = nothing,
...
-        J, f, fu, u, p, stats, autodiff, di_extras, sdifft_extras)
+    return JacobianCache{iip}(J, f, fu, u, p, stats, autodiff, di_extras)
 end
 
 function JacobianCache(prob, alg, f::F, ::Number, u::Number, p; stats,
         autodiff = nothing, kwargs...) where {F}
     fu = f(u, p)
     if SciMLBase.has_jac(f) || SciMLBase.has_vjp(f) || SciMLBase.has_jvp(f)
-        return JacobianCache{false}(u, f, fu, u, p, stats, autodiff, nothing, nothing)
+        return JacobianCache{false}(u, f, fu, u, p, stats, autodiff, nothing)
```
gdalle, line 113: Same question about cases where DI preparation is skipped

src/internal/jacobian.jl
```diff
@@ -63,31 +62,13 @@ function JacobianCache(prob, alg, f::F, fu_, u, p; stats, autodiff = nothing,
...
-            di_extras = if iip
-                DI.prepare_jacobian(f, fu, autodiff, u, Constant(prob.p))
-            else
-                DI.prepare_jacobian(f, autodiff, u, Constant(prob.p))
-            end
+            DI.prepare_jacobian(f, autodiff, u, Constant(prob.p))
         end
     else
-        using_sparsedifftools = false
         di_extras = nothing
```
gdalle, line 71: In which cases do we not prepare the jacobian? With DI, if you forget to do that, you'll recompute the sparsity pattern, the coloring and the basis vectors every single time

avik-pal: accidental, removed

src/internal/jacobian.jl
avik-pal: Only where the user provides an analytic jacobian. This one is a special case for scalars when DI is skipped if jacobian/jvp/vjp is provided.

src/internal/jacobian.jl
oscardssmith: you don't prepare the jacobian if the solver doesn't need the jacobian and the user didn't provide a jacobian function. In these cases, the nonlinear solver is doing everything with vjps and jvps anyway so you don't need this info.

PR #472, opened 2024-10-04
Use DI preparation result when initializing Jacobian

### LINKED ISSUE #469: cache construction is no longer type stable
I will generate a MWE but the DI extras is not type stable, atleast in the bruss case I tested.

gdalle:
This should fix #469 for the example you gave.

### DIFF
--- a/src/internal/jacobian.jl
+++ b/src/internal/jacobian.jl
@@ -89,7 +89,7 @@ function JacobianCache(prob, alg, f::F, fu_, u, p; stats, autodiff = nothing,
                 if iip
                     DI.jacobian(f, fu, di_extras, autodiff, u, Constant(p))
                 else
-                    DI.jacobian(f, autodiff, u, Constant(p))
+                    DI.jacobian(f, di_extras, autodiff, u, Constant(p))
                 end
             end
         else


avik-pal (approved):
Ah my bad, completely missed it. Thanks!

gdalle:
It would be fair to expect `DI.jacobian(b, AutoForwardDiff(; chunksize=2), x)` to be type-stable. The problem is that this falls back directly onto `ForwardDiff.jacobian(f, x)`, which is itself type-unstable since I don't give it the static `Chunk` size it needs to create the `JacobianConfig`. I could give it the chunk size, but I would have to create a `JacobianConfig` myself for that, which is unnecessary work e.g. [for StaticArrays](https://github.com/JuliaDiff/ForwardDiff.jl/blob/ec74fbc32b10bbf60b3c527d8961666310733728/ext/ForwardDiffStaticArraysExt.jl#L66).

ChrisRackauckas:
We allowed for specifying chunk size as a Val to handle that. We probably need to ensure that's allowed in ADTypes implementations.

gdalle:
This should solve it: https://github.com/gdalle/DifferentiationInterface.jl/pull/541

PR #474, opened 2024-10-05
fix: don't use similar on prototype

avik-pal:


src/internal/jacobian.jl
```diff
@@ -92,13 +92,13 @@ function JacobianCache(prob, alg, f::F, fu_, u, p; stats, autodiff = nothing,
...
         else
-            jac_proto = if eltype(f.jac_prototype) <: Bool
-                similar(f.jac_prototype, promote_type(eltype(fu), eltype(u)))
+            if eltype(f.jac_prototype) <: Bool
+                jac_proto = similar(f.jac_prototype, promote_type(eltype(fu), eltype(u)))
+                fill!(jac_proto, false)
+                jac_proto
             else
-                similar(f.jac_prototype)
+                f.jac_prototype
```
ChrisRackauckas, line 100: for thread safety, it still needs to be copied?

gdalle:
I have opened https://github.com/gdalle/DifferentiationInterface.jl/issues/544 to discuss if the same should be done for DI

gdalle:
As shown in that issue, you take a possibly big performance hit by using your copy. I think it would make more sense to fix the similar semantics for banded arrays

avik-pal:
Specifically this is for FastAlmostBandedMatrices and not BandedMatrices

gdalle:
See, this is so specific I don't think it warrants a performance and generality decrease overall

ChrisRackauckas:
Whether this is done or not, FastAlmostBandedMatrices should fix its similar dispatch.

avik-pal:
https://github.com/SciML/FastAlmostBandedMatrices.jl/pull/16

PR #477, opened 2024-10-10
fix: allow `u0` to be duals

avik-pal:


src/internal/forward_diff.jl
```diff
@@ -2,16 +2,33 @@
...
-        prob::NonlinearProblem{<:Union{Number, <:AbstractArray}, iip,
-            <:Union{<:Dual{T, V, P}, <:AbstractArray{<:Dual{T, V, P}}}},
-        alg::Union{Nothing, AbstractNonlinearAlgorithm},
-        args...;
-        kwargs...) where {T, V, P, iip}
-    sol, partials = __nlsolve_ad(prob, alg, args...; kwargs...)
-    dual_soln = __nlsolve_dual_soln(sol.u, partials, prob.p)
-    return SciMLBase.build_solution(
-        prob, alg, dual_soln, sol.resid; sol.retcode, sol.stats, sol.original)
+for (uType, pType) in [
```
oscardssmith, line 5: Why is this necessary? This seems pretty ugly

avik-pal: ambiguity prevention

oscardssmith: is there a reason this doesn't overload `__init` and `__solve`  like `OrdinaryDiffEq does?

avik-pal: I was asking myself the same question today morning

ChrisRackauckas: i think you need to so that you promote first.

avik-pal:
This is now handled in #458 (via dispatches on __init & __solve)

PR #478, opened 2024-10-10
Sample enzyme extension

wsmoses:
See https://github.com/SciML/NonlinearSolve.jl/issues/476


oscardssmith:
How does performance compare?

avik-pal:
From our lab meeting discussion earlier this morning, we decided to have an Enzyme special case for SimpleNonlinearSolve at the very least (I will add that via https://github.com/SciML/NonlinearSolve.jl/pull/458).

We can see the overall performance benefits here (in the NonlinearSolve context) later, but for SimpleNonlinearSolve, I would argue we want to save every second we can, and if adding a minor maintenance burden is the way to go (especially with NonlinearSolveBase switching the default backend to Enzyme if loaded), then we should do it.

PR #480, opened 2024-10-21
fix: remove uses of LineSearches.jl backtracking

avik-pal:


avik-pal:
This is all we need to remove a direct `LineSearches` dep in v4

PR #482, opened 2024-10-25
feat: add `PETScSNES`

### LINKED ISSUE #160: An interface to PETSc SNES nonlinear solve would be nice

avik-pal:
`PETSc` 0.2 resolves old compatibility issues we were having with wrapping this library.

> [!WARNING]  
> I am still seeing occasional segfaults with PETSc. Very common in 1.11. And for some reason PETSc hijacks `Ctrl + C` and trying to use it leads to a segfault even when no code is running

fixes #160 

## TODOs

- [x] Sparsity Handling?
- [x] Testing
- [x] Rewrite one of the examples from `PETSc.jl`


avik-pal:
This 150 lines of code https://github.com/JuliaParallel/PETSc.jl/blob/main/examples/SNES_ex2.jl translates to 30 lines in nonlinearsolve with automatic sparsity detection

```julia
# This implements src/snes/examples/tutorials/ex2.c from PETSc using the PETSc.jl package, using SNES
#
# This solves the equations sequentially
# 
# Newton method to solve u'' + u^{2} = f, sequentially.
using NonlinearSolve, PETSc, LinearAlgebra, SparseConnectivityTracer

n = 21
u0 = fill(0.5, n)

function form_residual!(resid, x, _)
    n = length(x)
    xp = LinRange(0.0, 1.0, n)
    F = 6xp .+ (xp .+ 1e-12) .^ 6

    dx = 1 / (n - 1)
    resid[1] = x[1]
    for i in 2:(n - 1)
        resid[i] = (x[i - 1] - 2x[i] + x[i + 1]) / dx^2 + x[i] * x[i] - F[i]
    end
    resid[n] = x[n] - 1

    return
end

nlfunc = NonlinearFunction{true}(form_residual!; sparsity = TracerSparsityDetector())
nlprob = NonlinearProblem(nlfunc, u0)

solve(nlprob, NewtonRaphson())
solve(nlprob, PETScSNES())
```

avik-pal:
Looking at the profile all of the time goes in the UMFPACK

docs/src/api/petsc.md
```diff
@@ -0,0 +1,17 @@
+# PETSc.jl
+
+This is a extension for importing solvers from PETSc.jl SNES into the SciML interface. Note
+that these solvers do not come by default, and thus one needs to install the package before
+using these solvers:
+
+```julia
+using Pkg
+Pkg.add("PETSc")
+using PETSc, NonlinearSolve
```
ChrisRackauckas, line 10: Don't you also need to `using MPI`?

docs/src/tutorials/snes_ex2.md
```diff
@@ -0,0 +1,85 @@
...
+### Sparse Jacobian
+
+```@example snes_ex2
+@benchmark solve($(nlprob_sparse), $(NewtonRaphson()); abstol = 1e-8)
+nothing # hide
+```
+
+```@example snes_ex2
+@benchmark solve($(nlprob_sparse), $(PETScSNES()); abstol = 1e-8)
+nothing # hide
```
ChrisRackauckas, line 84: What's the result? Just curious

docs/src/api/petsc.md
avik-pal: No that's implicitly loaded inside PETSc

docs/src/tutorials/snes_ex2.md
avik-pal: https://html-preview.github.io/?url=https://raw.githubusercontent.com/SciML/NonlinearSolve.jl/refs/heads/gh-pages/previews/PR482/tutorials/snes_ex2/index.html#Sparse-Jacobian

not by much on CI. On my laptop it is 300us vs 260us (PETSc being faster).

avik-pal: How optimized is our sparse solver selection for LinearSolve?

ChrisRackauckas: Relatively optimal? What sparse linear solver is PETSc using here? Try all of UMFPACK, KLU, and MKLPardiso. Other thing to try is `using MKL` with Pardiso. Then note that there's a new sparse solver ParU that @rayegun is getting wrapped https://github.com/SciML/LinearSolve.jl/issues/394. 

One thing to double check is what the extra time is spent in. Is it only doing one symbolic factorization?

avik-pal: Let's get this in, since it is non-blocking. I anyways need to redo benchmarks for the paper, so I can revisit it after the splitting is done.

PR #507, opened 2024-11-18
Add Halley and Householder to SimpleNonlinearSolve

tansongchen:
## Checklist

## Additional context

Note that I didn't use `AutoTaylorDiff()` as the ADType, since TaylorDiff hasn't become a full-featured AD like ForwardDiff, therefore it's still faster to compute Jacobian with ForwardDiff or something else. Instead, an additional field `Taylor_mode` is added to control whether to use TaylorDiff or not.

### DIFF
--- a/lib/SimpleNonlinearSolve/src/halley.jl
+++ b/lib/SimpleNonlinearSolve/src/halley.jl
@@ -15,6 +15,7 @@ A low-overhead implementation of Halley's Method.
   - `autodiff`: determines the backend used for the Jacobian. Defaults to  `nothing` (i.e.
     automatic backend selection). Valid choices include jacobian backends from
     `DifferentiationInterface.jl`.
+    In addition, `AutoTaylorDiff` can be used to enable Taylor mode for computing the Hessian-vector-vector product more efficiently; in this case, the Jacobian would still be calculated using the default backend. You need to have `TaylorDiff.jl` loaded to use this option.
 """
 @kwdef @concrete struct SimpleHalley <: AbstractSimpleNonlinearSolveAlgorithm
     autodiff = nothing
--- a/lib/SimpleNonlinearSolve/Project.toml
+++ b/lib/SimpleNonlinearSolve/Project.toml
@@ -26,6 +26,7 @@ StaticArraysCore = "1e83bf80-4336-4d27-bf5d-d5a4f845583c"
 [weakdeps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
 ReverseDiff = "37e2e3b7-166d-5795-8a7a-e32c996b4267"
+TaylorDiff = "b36ab563-344f-407b-a36a-4f200bebf99c"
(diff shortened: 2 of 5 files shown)


lib/SimpleNonlinearSolve/src/halley.jl
```diff
@@ -15,16 +15,18 @@ A low-overhead implementation of Halley's Method.
   - `autodiff`: determines the backend used for the Jacobian. Defaults to  `nothing` (i.e.
     automatic backend selection). Valid choices include jacobian backends from
     `DifferentiationInterface.jl`.
+  - `taylor_mode`: whether to use Taylor mode automatic differentiation to compute the Hessian-vector-vector product. Defaults to `Val(false)`. If `Val(true)`, you must have `TaylorDiff.jl` loaded.
```
avik-pal, line 18: This is going to be extremely confusing and I don't see why can't there be a `AutoTaylorDiff()` even if it is being used for very restrictive usecases?

For example, what does it mean if someone specifies AutoForwardDiff + tyalor_mode = Val(true). In other parts of SciML, re OrdinaryDiffEq we are aggressively moves away from this kind of multi-arg AD specifications, it would be a bad idea to re-introduce that here.

avik-pal (changes requested):

lib/SimpleNonlinearSolve/src/halley.jl
avik-pal: A better way would be to use SecondOrder from DifferentiationInterface to specify using different backends. If outer is AutoTaylorDiff then you use the specialized version

ChrisRackauckas: I really think we should just make TaylorDiff use ADTypes. If we don't use DifferentiationInterface.jl here then that's a detail, but I don't think that should effect the user interface.

tansongchen: The reason is that currently TaylorDiff haven't got chunk size which makes it less efficient to compute the Jacobian. So, even if the user choose to use TaylorDiff for the Halley correction, they should still use ForwardDiff or something else to do the Jacobian.

If there is only one argument to be AutoTaylorDiff(), then I'm not sure what to do with Jacobian.

ChrisRackauckas: Just hardcode it then.

tansongchen: By saying "hardcode", did you mean when argument is "AutoTaylorDiff()", the behavior is to use TaylorDiff for Halley correction and ForwardDiff for Jacobian?

ChrisRackauckas: yes

tansongchen:
Would be fine once https://github.com/SciML/ADTypes.jl/pull/99 is ready

lib/SimpleNonlinearSolve/src/halley.jl
tansongchen: The taylor_mode arg is removed.

PR #523, opened 2024-12-26
feat: add `NonlinearSolveHomotopyContinuation.jl`

AayushSabharwal:
## Checklist


ChrisRackauckas:
What is left here?

ChrisRackauckas:
Add this to the CI

AayushSabharwal:
This is a bit stochastic since whether HC considers a root "real" depends on if the imaginary part is less than the accuracy of their solve. Sometimes their results are more accurate, sometimes less.

ChrisRackauckas:
That should be easily patchable on their end in like a day?

AayushSabharwal:
https://github.com/JuliaHomotopyContinuation/HomotopyContinuation.jl/issues/614

ChrisRackauckas:
It would be nice to have a short tutorial that showcases a place where Newton methods fail but this gives an accurate answer. Do you have one?

AayushSabharwal:
iirc bruss with some guesses can show this behavior. Will write something

PR #532, opened 2025-02-14
Use Sparspak when handling sparse system

ErikQQY:
Fix: https://github.com/SciML/LinearSolve.jl/issues/578
With the recent changes in LinearSolve, we need to use Sparspak when solving a sparse nonlinear problem


ErikQQY:
It seems using Sparspak when handling sparse nonlinear system indeed fix: https://github.com/SciML/LinearSolve.jl/issues/578

PR #534, opened 2025-02-20
Draft: `ImplicitDiscreteSolve`

vyudu:
Requires: 
- https://github.com/SciML/OrdinaryDiffEq.jl/pull/2607
- https://github.com/SciML/SciMLBase.jl/pull/937


vyudu:
@ChrisRackauckas This is barebones but working. Any important features I'm missing here? Also should this actually go in OrdinaryDiffEq?

ChrisRackauckas:
Can this to use SimpleNonlinearSolve.jl's Newton and make this be SimpleImplicitDiscreteSolve.jl? I think the complete version will want to use the Jacobian reuse aspect and such, so for now having a static version would be a good way to handle this.

lib/SimpleImplicitDiscreteSolve/src/SimpleImplicitDiscreteSolve.jl
```diff
@@ -0,0 +1,31 @@
...
+
+using SciMLBase: AbstractNonlinearProblem
+using SciMLBase
+using NonlinearSolveBase
+using SimpleNonlinearSolve
+using SymbolicIndexingInterface
+using LinearAlgebra
+using ADTypes
+using UnPack
+import OrdinaryDiffEqCore: OrdinaryDiffEqAlgorithm, alg_cache, OrdinaryDiffEqMutableCache, OrdinaryDiffEqConstantCache, get_fsalfirstlast, isfsal, initialize!, perform_step!, isdiscretecache, isdiscretealg, alg_order, beta2_default, beta1_default, dt_required
```
ChrisRackauckas, line 11: are these needed?

vyudu:
@ChrisRackauckas I think this is basically ready, initialization for problems without initializationproblems needs work but I think it's fine for the MTK purposes. How do I register it?

Project.toml
```diff
@@ -25,6 +25,7 @@ PrecompileTools = "aea7be01-6a6a-4083-8856-8a6e6704d82a"
 Preferences = "21216c6a-2e73-6563-6e65-726566657250"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 SciMLBase = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
+SimpleImplicitDiscreteSolve = "3263718b-31ed-49cf-8a0f-35a466e8af96"
```
ChrisRackauckas, line 28: this isn't necessary?

lib/SimpleImplicitDiscreteSolve/src/solve.jl
```diff
@@ -0,0 +1,81 @@
+# Remake the nonlinear problem, then update
+function perform_step!(integrator, cache::SimpleIDSolveCache, repeat_step = false)
+    @unpack alg, u, uprev, dt, t, f, p = integrator
+    @unpack state, prob = cache
+    state.u .= uprev
+    state.t_next = t
+    prob = remake(prob, p = state)
+
+    u = solve(prob, SimpleNewtonRaphson())
+    any(isnan, u) && (integrator.sol.retcode = SciMLBase.ReturnCode.Failure)
```
ChrisRackauckas, line 10: why not check the retcode of u here?

lib/SimpleImplicitDiscreteSolve/src/solve.jl
```diff
@@ -0,0 +1,81 @@
...
+#     if alg_extrapolates(integrator.alg)
+#         recursivecopy!(integrator.uprev2, integrator.uprev)
+#     end
+# 
+#     if nlsol.retcode != ReturnCode.Success
+#         integrator.sol = SciMLBase.solution_new_retcode(integrator.sol,
+#             ReturnCode.InitialFailure)
+#     end
+#     return
+# end
```
ChrisRackauckas, line 81:
```suggestion
```

That's not necessary.

vyudu:
Fixed these

ChrisRackauckas:
Okay so this implementation finds itself in a somewhat odd spot. It's not a "Simple" alg because it's not statically compile-able, which is a property that all others have. It would need to avoid the OrdinaryDiffEqCore part. But it's not a fully complete version either? But it's a rough start towards it. I guess seeing it at this stage, how about moving this to OrdinaryDiffEq and making this be ImplicitDiscreteSolve.jl, and we can use this as a base to do the required optimizations to make the full version. We can do the SimpleImplicitDiscreteSolve.jl as a somewhat separate version for static support.

vyudu:
I see, didn't realize I couldn't use OrdinaryDiffEqCore, will move. I'll keep this open too to keep iterating it toward a SimpleImplicitDiscreteSolve.

ChrisRackauckas:
See for example https://github.com/SciML/SimpleDiffEq.jl/blob/master/src/rk4/looprk4.jl which is the most basic RK4 implementation. For IDS, it would just be a loop over nonlinear solve calls. With SimpleNonlinearSolve being similar, the whole thing would statically compile since no mutable structs would be in the middle.

vyudu:
@ChrisRackauckas tried to refactor this to a simple solve

lib/SimpleImplicitDiscreteSolve/src/SimpleImplicitDiscreteSolve.jl
```diff
@@ -0,0 +1,95 @@
...
+    else
+        us = Vector{typeof(u0)}(undef, 1) # for interface compatibility
+    end
+
+    u = u0
+    convfail = false
+    for i in 2:length(ts)
+        uprev = u
+        t = ts[i]
+        nlf = isinplace(f) ? (out, u, p) -> f(out, u, uprev, p, t) : (u, p) -> f(u, uprev, p, t)
```
ChrisRackauckas, line 69: is the compiler always smart enough to optimize this closure? It probably is these days. I would've just made a callable type, but I think this got better. I guess we can just patch this if we find a case where it does allocate.

ChrisRackauckas (approved):

src/NonlinearSolve.jl
```diff
@@ -42,6 +42,7 @@ using NonlinearSolveFirstOrder: NonlinearSolveFirstOrder, GeneralizedFirstOrderA
 using NonlinearSolveQuasiNewton: NonlinearSolveQuasiNewton, QuasiNewtonAlgorithm
 using NonlinearSolveSpectralMethods: NonlinearSolveSpectralMethods, GeneralizedDFSane
 using SimpleNonlinearSolve: SimpleNonlinearSolve
+using SimpleImplicitDiscreteSolve: SimpleIDSolve
```
ChrisRackauckas, line 45:
```suggestion
```

src/NonlinearSolve.jl
```diff
@@ -109,7 +110,8 @@ end
 # Rexexports
 @reexport using SciMLBase, NonlinearSolveBase, LineSearch, ADTypes
 @reexport using NonlinearSolveFirstOrder, NonlinearSolveSpectralMethods,
-                NonlinearSolveQuasiNewton, SimpleNonlinearSolve, BracketingNonlinearSolve
+                NonlinearSolveQuasiNewton, SimpleNonlinearSolve, BracketingNonlinearSolve,
+                SimpleImplicitDiscreteSolve
```
ChrisRackauckas, line 114:
```suggestion
                NonlinearSolveQuasiNewton, SimpleNonlinearSolve, BracketingNonlinearSolve
```

Project.toml
ChrisRackauckas, line 28:
```suggestion
```

ChrisRackauckas:
This folder will need a license file in order to register, but it should be good to go now.

Project.toml
```diff
@@ -25,6 +25,7 @@ PrecompileTools = "aea7be01-6a6a-4083-8856-8a6e6704d82a"
 Preferences = "21216c6a-2e73-6563-6e65-726566657250"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 SciMLBase = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
+SimpleImplicitDiscreteSolve = "8b67ef88-54bd-43ff-aca0-8be02588656a"
```
ChrisRackauckas, line 28:
```suggestion
```

src/NonlinearSolve.jl
ChrisRackauckas, line 45:
```suggestion
```

src/NonlinearSolve.jl
ChrisRackauckas, line 114:
```suggestion
                NonlinearSolveQuasiNewton, SimpleNonlinearSolve, BracketingNonlinearSolve
```

src/NonlinearSolve.jl
ChrisRackauckas, line 45:
```suggestion
```

src/NonlinearSolve.jl
```diff
@@ -109,7 +109,8 @@ end
...
```
ChrisRackauckas, line 113:
```suggestion
                NonlinearSolveQuasiNewton, SimpleNonlinearSolve, BracketingNonlinearSolve
```

Project.toml
ChrisRackauckas, line 28:
```suggestion
```

lib/SimpleImplicitDiscreteSolve/src/SimpleImplicitDiscreteSolve.jl
```diff
@@ -0,0 +1,98 @@
...
+    if save_everystep && save_start
+        us = Vector{typeof(u0)}(undef, length(ts))
+        us[1] = u0
+    elseif save_everystep
+        us = Vector{typeof(u0)}(undef, length(ts) - 1)
+    elseif save_start
+        us = Vector{typeof(u0)}(undef, 2)
+        us[1] = u0
+    else
+        us = Vector{typeof(u0)}(undef, 1) # for interface compatibility
```
ChrisRackauckas, line 61: It would be good to make this possibly be an SArray to skip the allocation when possible. This is done for example in https://github.com/SciML/SimpleDiffEq.jl/blob/master/src/rk4/gpurk4.jl#L40

ChrisRackauckas:
Alright it looks like it's passing. Let's move this version over and register.

vyudu:
Oops I just added this staticarrays thing

lib/SimpleImplicitDiscreteSolve/src/SimpleImplicitDiscreteSolve.jl
```diff
@@ -49,16 +50,13 @@ function DiffEqBase.solve(prob::ImplicitDiscreteProblem, alg::SimpleIDSolve;
...
-        us = Vector{typeof(u0)}(undef, length(ts))
-        us[1] = u0
-    elseif save_everystep
-        us = Vector{typeof(u0)}(undef, length(ts) - 1)
-    elseif save_start
-        us = Vector{typeof(u0)}(undef, 2)
+    l = save_everystep ? length(ts) - 1 : 1
+    save_start && (l = l + 1)
+    u0type = typeof(u0)
+    us = u0type <: StaticArray ? MVector{l, u0type}(undef) : Vector{u0type}(undef, l)
```
ChrisRackauckas, line 56: This won't infer though

ChrisRackauckas: Maybe a switch in the alg that is type based? We can follow up with that though. Let's get a first version and then get that non-allocating.

vyudu: OK I'll open it in OrdinaryDiffEq

PR #536, opened 2025-03-03
fix type stability of SimpleNonlinearSolve

oscardssmith:
This fixes an issue where we were returning a solution with a different `alg.autodiff` or `prob` if we had an `InitialFailure` vs a successful solve. This removes the allocations we were seeing in https://docs.sciml.ai/NonlinearSolve/stable/tutorials/code_optimization/#Further-Optimizations-for-Small-Nonlinear-Solves-with-Static-Arrays-and-SimpleNonlinearSolve.

Also fixes https://github.com/SciML/NonlinearSolve.jl/pull/536


ChrisRackauckas (approved):

AayushSabharwal:
MTK CI tests this so it'd be good to wait for it

oscardssmith:
looks like I broke Halley a bit. Fixes incoming.

baggepinnen:
new release

PR #548, opened 2025-03-19
more ITP simplification

oscardssmith:
followup to https://github.com/SciML/NonlinearSolve.jl/pull/547. There is one minor change in that `diff` now is always positive (following the algorithm description in the paper and https://en.wikipedia.org/wiki/ITP_method). Otherwise, this should be nfc.


ChrisRackauckas:
Conflicts?

oscardssmith:
conflicts fixed. Possibly worth noting that this PR currently adds stats since stats are really useful for seeing if the alg works, but do add an allocation of the stats object to previously non-allocating code. Thoughts?

oscardssmith:
also, is the intended behavior different depending on whether tspan is forwards or backwards?

ChrisRackauckas:
We've normally tried to make all of the bracketing methods all non-allocating since they don't necessarily need any. I'd comment out the stats stuff until we get a better solution there then.

oscardssmith:
It's very hard for me to see how the test failures could be my fault, but otoh I know that theoretically they all should be working now.

ChrisRackauckas:
Downgrade and Windows are not yours. Those are just CI infra issues. @avik-pal you had something for the Windows stuff?

PR #550, opened 2025-03-23
More generic safe similar

ErikQQY:
Instead of only focusing on `similar` BigFloat throw `#undef` cases, we should use a more generic `similar` to cover more `#undef` cases.

### DIFF
--- a/lib/NonlinearSolveBase/src/utils.jl
+++ b/lib/NonlinearSolveBase/src/utils.jl
@@ -101,13 +101,13 @@ restructure(y, x) = ArrayInterface.restructure(y, x)
 
 function safe_similar(x, args...; kwargs...)
     y = similar(x, args...; kwargs...)
-    return init_bigfloat_array!!(y)
+    return init_similar_array!!(y)
 end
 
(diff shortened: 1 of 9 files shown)


ErikQQY:
Test failures seem not related

lib/NonlinearSolveBase/src/utils.jl
```diff
@@ -101,13 +101,13 @@ restructure(y, x) = ArrayInterface.restructure(y, x)
...
-    return init_bigfloat_array!!(y)
+    return init_similar_array!!(y)
 end
 
-init_bigfloat_array!!(x) = x
+init_similar_array!!(x) = x
 
-function init_bigfloat_array!!(x::AbstractArray{<:BigFloat})
-    ArrayInterface.can_setindex(x) && fill!(x, BigFloat(0))
+function init_similar_array!!(x::AbstractArray{<:T}) where {T}
```
ChrisRackauckas, line 109:
```suggestion
function init_similar_array!!(x::AbstractArray{<:T}) where {T <: Number}
```

PR #563, opened 2025-03-27
Fix ForwardDiff overload for user vjp

ChrisRackauckas:


ChrisRackauckas:
Inference issues look new, @oscardssmith take a look?

https://github.com/SciML/NonlinearSolve.jl/actions/runs/14109779969/job/39525170712?pr=563#step:5:630

oscardssmith:
this test is the inference test that I added 2 days ago. I think the problem is that the test is in OrdinaryDiffEq but the code is in one of the subpackages

PR #566, opened 2025-03-31
support Zygote@0.7

oscardssmith:
This is needed to support ForwardDiff@1.0. Testing separately to see how CI looks.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -115,7 +115,7 @@ StaticArraysCore = "1.4"
 Sundials = "4.23.1"
 SymbolicIndexingInterface = "0.3.31"
 Test = "1.10"
-Zygote = "0.6.69"
+Zygote = "0.6.69, 0.7"
 julia = "1.10"
 
 [extras]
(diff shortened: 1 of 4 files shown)


oscardssmith:
unfortunately it looks like there are real inference failures caused by making everything a thunk in Zygote@0.7.

ChrisRackauckas:
These are real failures from this? I don't see how it would be related. Run master tests?

oscardssmith:
From talking to Avik, Zygote@0.7 creates more thunks than 0.6 which can regress type stability.

ChrisRackauckas:
But that should only effect tests of reverse mode, since the standard only uses forward mode?

oscardssmith:
> But that should only effect tests of reverse mode, since the standard only uses forward mode?

The test that is failing is SimpleNonlinearSolve with `AutoReverseDiff`. https://github.com/SciML/NonlinearSolve.jl/pull/567 doesn't see the same failure.

Issue #403, opened 2024-04-08
Muller's method


fgittins:
**What kind of problems is it mostly used for? Please describe.**

Determining the roots of univariate complex functions. Muller's method is commonly used to determine the (quasi-normal) oscillation modes of neutron stars [1, 2].

**Note:** Given that this method only works for univariate functions, it might be too restrictive for `NonlinearSolve.jl`, but I thought I would suggest it anyway, since I frequently use it. Also, the algorithm is very simple so it should be easy to implement.

**Describe the algorithm you’d like**

The algorithm is quite simple and described well in Press et al. [3]. (See also Wikipedia [4].)

To summarise: Muller's method is a generalisation of the secant method, with the key difference being that it uses quadratic interpolation across three points (as opposed to linear interpolation among two). Solving for the roots of a quadratic is trivial and allows the method to work for complex roots.

**Other implementations to know about**

There is an existing Julia implementation in `Roots.jl` [5] and a Fortran routine in `IMSL` [6].

**References**

[1] [Kokkotas & Schutz (1992)](https://ui.adsabs.harvard.edu/abs/1992MNRAS.255..119K/abstract)
[2] [Kruger, PhD Thesis (2016)](https://ui.adsabs.harvard.edu/abs/2015PhDT.......604K/abstract)
[3] [Press et al. (2007), Sec. 9.5.2, p. 466](https://numerical.recipes/book.html)
[4] [Wikipedia, Muller's method](https://en.wikipedia.org/wiki/Muller%27s_method)
[5] [`Roots.muller`](https://juliamath.github.io/Roots.jl/stable/reference/#Roots.muller)
[6] [IMSL Math/Library Users Manual, `Zanly`](https://help.imsl.com/fortran/fnlmath/current/ZANLY.htm?Highlight=muller%27s%20method)

avik-pal:
>  Given that this method only works for univariate functions, it might be too restrictive for NonlinearSolve.jl

This would be great to have in SimpleNonlinearSolve

avik-pal:
yes open the pr

ChrisRackauckas:
It should be relatively the same. SimpleNonlinearSolve.jl mostly moved repos and changed from using DiffEqBase to NonlinearSolveBase, but its system isn't that different IIRC.

PR #568, opened 2025-04-05
Add Muller's method

fgittins:
## Checklist

## Additional context

This implements the algorithm discussed in #403.

An older pull request was opened in the since deprecated `SimpleNonlinearSolve.jl` package (see https://github.com/SciML/SimpleNonlinearSolve.jl/pull/139). This implementation uses the suggestions made in that pull request and implements the `Muller` algorithm as an `IntervalNonlinearProblem` under the `BracketingNonlinearSolve.jl` sub-library.

Moving forward, I think we need to
- Re-work the simple tests I have added to be more consistent with the other algorithms (see `muller_tests.jl`).
- Include some functionality to allow the user to instead specify the middle guess. To be consistent with the other algorithms, currently the middle guess is taken as a midpoint of the two (`left` and `right`) specified guesses.


ChrisRackauckas (approved):

ChrisRackauckas:
Add the algorithm into the other test sets: https://github.com/SciML/NonlinearSolve.jl/blob/master/lib/BracketingNonlinearSolve/test/rootfind_tests.jl#L10-L79

ChrisRackauckas:
This algorithm isn't bounds refinement, so that makes sense.

fgittins:
I've implemented a feature that allows the user to specify the `middle` guess, along with some unit tests.

The way this is implemented is at the algorithm level, where there is now an optional `middle` keyword argument,

```julia
Muller(; middle = nothing)
```

PR #570, opened 2025-04-05
Add contributing manual

ErikQQY:
Add some context for developing NonlinearSolve

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,36 @@
+  - This repository follows the [SciMLStyle](https://github.com/SciML/SciMLStyle) and the SciML [ColPrac](https://github.com/SciML/ColPrac).
+  - Please run `using JuliaFormatter, NonlinearSolve; format(joinpath(dirname(pathof(NonlinearSolve)), ".."))` before committing.
+  - Add tests for any new features.
+  - Additional help on contributing to the numerical solvers can be found at https://docs.sciml.ai/NonlinearSolve/stable/
+
+## Developing Locally
+
+NonlinearSolve is a large package and thus it uses a sublibrary approach to keep down
+the total number of dependencies per solver. As a consequence, it requires a bit of special
+handling compared to some other Julia packages.
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,36 @@
...
+using Pkg
+Pkg.develop(map(
+    path -> Pkg.PackageSpec.(; path = "$(@__DIR__)/lib/$(path)"), readdir("./lib")));
+```
+
+and then running tests accordingly.
+
+## Dependency Structure
+
+There is a tree dependency structure to the sublibraries of NonlinearSolve.jl. The core subpackage NonlinearSolveBase.jl is the hard dependency of all the other subpackages. SimpleNonlinearSolve.jl is a special one with BracketingNonlinearSolve.jl as dependency. NonlinearSolve.jl as the parent package contains all of the subpackages.
```
ChrisRackauckas, line 36: >SimpleNonlinearSolve.jl is a special one with BracketingNonlinearSolve.jl as dependency.

That should change.

avik-pal:
we should just use sources instead of having to manually dev https://github.com/LuxDL/Lux.jl/blob/f6cecd2ba03fd9f8d784b3fa9b5c8e667fe7314f/Project.toml#L56-L60

CONTRIBUTING.md
ErikQQY: Do you mean that we should no longer use and reexport BracketingNonlinearSolve in SimpleNonlinearSolve?

CONTRIBUTING.md
```diff
@@ -0,0 +1,36 @@
...
+  - Add tests for any new features.
+  - Additional help on contributing to the numerical solvers can be found at https://docs.sciml.ai/NonlinearSolve/stable/
+
+## Developing Locally
+
+NonlinearSolve is a large package and thus it uses a sublibrary approach to keep down
+the total number of dependencies per solver. As a consequence, it requires a bit of special
+handling compared to some other Julia packages.
+
+When running the subpackage test suite, it's recommended that one has dev'd its relevant packages in NonlinearSolve.jl. This can be done via(use `NonlinearSolveFirstOrder` as example):
```
fgittins, line 12: Some minor typos. Should probably be:

```via (with `NonlinearSolveFirstOrder` as an example)```

PR #574, opened 2025-04-06
Fix SimpleNonlinearSolve CI

ErikQQY:
Still need to investigate the inference error


avik-pal:
fixed all of these in https://github.com/SciML/NonlinearSolve.jl/pull/580

PR #580, opened 2025-04-18
fix: failing tests

avik-pal:


avik-pal:
finite diff failures are a problem upstream investigating that

avik-pal:
https://github.com/SciML/NonlinearSolve.jl/actions/runs/14544669588/job/40808434530?pr=580 is the only failing one atm and I am not sure why

lib/SimpleNonlinearSolve/src/lbroyden.jl
```diff
@@ -62,10 +62,11 @@ end
 
     # For scalar problems / if the threshold is larger than problem size just use Broyden
     if x isa Number || length(x) ≤ η
-        return SciMLBase.__solve(
+        sol = SciMLBase.__solve(
             prob, SimpleBroyden(; alg.linesearch), args...;
             abstol, reltol, maxiters, termination_condition, kwargs...
         )
+        return Utils.nonlinear_solution_new_alg(sol, alg)
```
ChrisRackauckas, line 69: won't this lead to type stability issues?

avik-pal: no this will fix it. we need to use simplebroyden for the solve process. But we ensure all branches mark the returned alg as simplelbroyden

PR #583, opened 2025-04-23
Simplify requirements from default polyalg

ChrisRackauckas:
One of the properties the default polyalg wants is not just robustness to the mathematical problem but also robustness to implementation. Certain issues like https://github.com/EnzymeAD/Enzyme.jl/issues/2360#issuecomment-2823335480 have highlighted recently that Backtracking linesearch has a vjp, and then in the scenario that `using Enzyme` is done globally somewhere in the system, the autodiff chooser will prefer Enzyme. That's not necessarily a bad thing, but this leads to odd failures / fixes like https://github.com/SciML/OrdinaryDiffEq.jl/pull/2683 where a seemingly random change to test order fixes things. A fairly common example of how a user can trigger this is that if they do `using Enzyme`, then the forward pass of the default nonlinear solver will now have the Backtracking line search change to using Enzyme reverse mode, which could make a previously solving example fail.

Another example is the Broyden with full Jacobian. This has machinery that require `pinv`, which for sparse matrices can be a not well-specified operation and can densify things. So that is not a general operation. It's not even clear that algorithm is a good choice in comparison to NR, since if you're making and `pinv`ing something then why not take a Newton step? Standard Broyden is motivated since it completely avoids the linear algebra but this intermediate step I don't think is really justfied as generally faster than NR or that much more robust, so I think it's an interesting option to maintain but I wouldn't put it in the default path.

This then makes the default path hit a few different trust region radius update schemes, which means it's globalized to things that do well, but also only requires forward mode across the board, making things stay simpler. This is a set of defaults that should be less requirements on user `f` functions. We can create other polyalgs that aren't restricted by this requirement, but I think the default should keep it simple in terms of autodiff support.

### DIFF
--- a/src/poly_algs.jl
+++ b/src/poly_algs.jl
@@ -40,7 +40,6 @@ function FastShortcutNonlinearPolyalg(
         else
             algs = (
                 NewtonRaphson(; common_kwargs...),
-                NewtonRaphson(; common_kwargs..., linesearch = BackTracking()),
                 TrustRegion(; common_kwargs...),
                 TrustRegion(; common_kwargs..., radius_update_scheme = RUS.Bastin)
             )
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
@avik-pal https://github.com/SciML/NonlinearSolve.jl/actions/runs/14612790929/job/40994078714?pr=583#step:5:1270 this is randomly showing up and not actually related right? I'm removing this method from the default method, but now it fails in the test 😅

ChrisRackauckas:
Noticed it was failing a bit ago in https://github.com/SciML/DiffEqBase.jl/pull/1141 so it's just not related.

ChrisRackauckas:
oh it was https://github.com/SciML/DiffEqProblemLibrary.jl/pull/138

PR #606, opened 2025-05-12
Fix return codes for nonlinear least squares

ChrisRackauckas:
Fixes https://github.com/SciML/NonlinearSolve.jl/issues/459. The crux of the issue is that `f(x) = residual` only applies in the NonlinearProblem and SteadyStateProblem cases. When `f(x)` is a nonlinear least squares problem, finding a local minima is a solution, not a failure of the algorithm. Thus this reclassifies Stalled in NLLSQ to StalledSuccess, which makes it a successful return.

Algorithms which require the NonlinearLeastSquares solution to have `||resid|| < tol` thus need to be careful with the return handling, as is done in the PR that introduces this return code https://github.com/SciML/SciMLBase.jl/pull/1016. However, that's a fairly odd case because it's feasibility checking, while the normal use case for NLLSQ is for optimization, and in an optimization case there's no reason to believe you should always have a solution close to zero.


lib/NonlinearSolveBase/src/termination_conditions.jl
```diff
@@ -209,7 +219,7 @@ function (cache::NonlinearTerminationModeCache)(
         end
         du_norm = L2_NORM(cache.u_diff_cache)
         cache.step_norm_trace[mod1(cache.nsteps, length(cache.step_norm_trace))] = du_norm
-        if cache.nsteps > mode.max_stalled_steps
+        if cache.nsteps > mode.max_stalled_steps || iszero(du_norm)
```
ChrisRackauckas, line 222:
```suggestion
        if cache.nsteps > mode.max_stalled_steps
```

test/core_tests.jl
```diff
@@ -427,3 +427,13 @@ end
...
 end
+
+@testitem "NonlinearLeastSquares ReturnCode" tags=[:core] begin
+    f(u,p) = [1.0]
+    nlf = NonlinearFunction(f; resid_prototype=zeros(1))
+    prob = NonlinearLeastSquaresProblem(nlf, [1.0])
+    sol = solve(prob)
+    @test SciMLBase.successful_retcode(sol)
+    @test sol.retcode == ReturnCode.StalledSuccess
+    @test sol.stats.nf == 3
```
ChrisRackauckas, line 438:
```suggestion
```

lib/NonlinearSolveBase/Project.toml
```diff
@@ -68,7 +68,7 @@ MaybeInplace = "0.1.4"
 Preferences = "1.4"
 Printf = "1.10"
 RecursiveArrayTools = "3"
-SciMLBase = "2.69"
+SciMLBase = "2.89"
```
ChrisRackauckas, line 71:
```suggestion
SciMLBase = "2.89.1"
```

lib/NonlinearSolveBase/src/termination_conditions.jl
```diff
@@ -177,7 +181,7 @@ function (cache::NonlinearTerminationModeCache)(
     end
 
     # Main Termination Criteria
-    if objective ≤ criteria
+    if !cache.leastsq && objective ≤ criteria
```
ChrisRackauckas, line 184:
```suggestion
    if objective ≤ criteria
```

PR #619, opened 2025-05-21
Bump to use SciMLBase ImmutableNonlinearProblem

ChrisRackauckas:


lib/NonlinearSolveBase/Project.toml
```diff
@@ -68,7 +68,7 @@ MaybeInplace = "0.1.4"
 Preferences = "1.4"
 Printf = "1.10"
 RecursiveArrayTools = "3"
-SciMLBase = "2.89.1"
+SciMLBase = "2.91"
```
ChrisRackauckas:
```suggestion
SciMLBase = "2.92"
```

PR #622, opened 2025-05-21
Continue on with the polyalg if a stalledsuccess

ChrisRackauckas:
Only exit if less than tolerance

### DIFF
--- a/lib/NonlinearSolveBase/src/solve.jl
+++ b/lib/NonlinearSolveBase/src/solve.jl
@@ -174,7 +174,7 @@ end
                         $(prob_syms[i]), alg.algs[$(i)], args...;
                         stats, alias_u0, verbose, kwargs...
                     )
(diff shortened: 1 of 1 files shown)


lib/NonlinearSolveBase/src/solve.jl
```diff
@@ -174,7 +174,7 @@ end
                         $(prob_syms[i]), alg.algs[$(i)], args...;
                         stats, alias_u0, verbose, kwargs...
                     )
-                    if SciMLBase.successful_retcode($(cur_sol))
+                    if SciMLBase.successful_retcode($(cur_sol)) && cur_sol.retcode !== ReturnCode.StalledSuccess
```
ChrisRackauckas, line 177:
```suggestion
                    if SciMLBase.successful_retcode($(cur_sol)) && $(cur_sol).retcode !== ReturnCode.StalledSuccess
```

PR #640, opened 2025-07-18
Remove copy in termination conditions reinit!

ChrisRackauckas:


lib/NonlinearSolveBase/src/termination_conditions.jl
```diff
@@ -98,9 +98,11 @@ function SciMLBase.reinit!(
...
 
     mode = cache.mode
-    u_unaliased = mode isa AbstractSafeBestNonlinearTerminationMode ?
-                  (ArrayInterface.can_setindex(u) ? copy(u) : u) : nothing
-    cache.u = u_unaliased
+    if ArrayInterface.can_setindex(u)
+        cache.u .= u
+    else
+        cache.u = u
+    end
```
ChrisRackauckas:
```suggestion
    if u isa Number ||  !ArrayInterface.can_setindex(u)
        cache.u = u
    else
        cache.u .= u
    end
```

PR #647, opened 2025-07-24
Draft: use SciMLVerbosity verbosity system

jClugstor:
## Checklist

## Additional context

I'm experimenting with using a ScopedValue to hold the NonlinearVerbosity object. There are a couple of reasons for this. Unlike LinearSolve, presently none of the caches have a `verbose` field that can hold the verbosity specifier. We could add these fields, but then only functions that explicitly depend on the cache can use the verbosity settings. Alternatively, we could thread the verbosity through every function that should be able to use it. Currently, many of the utility functions in NonlinearSolve don't explicitly depend on the cache, but do include warning messages, and we need to use the verbosity system in these functions if we want to be able to turn the warnings off while still having them available if needed. It seems unreasonable to have every function that might be used have to have the verbosity or the cache as one of it's arguments. 

From what I can tell it seems like ScopedValues might be good for this, apparently avoiding some of the performance issues of normal globals. For now I have a global ScopedValue `const nonlinear_verbose = ScopedValue{Union{NonlinearVerbosity{true}, NonlinearVerbosity{false}}}()`. Then `__init` and `solve!` are called in the context of the `@with` macro, where `nonlinear_verbose` is set to the value of the `verbose` keyword. This allows any function that's called by `__init` or `solve!` to use the verbosity settings from the `verbose` keyword, without having to explicitly depend on either the verbosity object or the nonlinear cache. 

I think my biggest concern here would be the performance of accessing the ScopedValue with `nonlinear_verbose[]`, since that will need to be done in any place the `@SciMLMessage` macro would be used, which could be in parts of the code that is run very often. 

I've been doing some profiling and some benchmarking and I do see a slowdown in some cases, but it's hard to tell if it's related to using ScopedValues or if it's other changes related to the verbosity system. 

@oscardssmith do you have a sense of whether this is a good idea or not? If there are going to be really obvious performance issues then it's probably not worth it to do this


oscardssmith:
do we have a new profile of this to confirm it's cheap?

jClugstor:
It seems like most of the failures are from downgrade...
I'm getting a lot of 
```julia
ERROR: LoadError: KeyError: key UUID("be0214bd-f91f-a760-ac4e-3421ce2b2da0") not found
Stacktrace:
 [1] getindex
   @ ./dict.jl:477 [inlined]
 [2] Resolver.SAT(info::Dict{UUID, Resolver.PkgInfo{UUID, VersionNumber}})
   @ Resolver /tmp/resolver/src/SAT.jl:79
 [3] resolve(info::Dict{UUID, Resolver.PkgInfo{UUID, VersionNumber}}, reqs::Vector{UUID}; max::Int64, by::Function)
   @ Resolver /tmp/resolver/src/Resolve.jl:200
 [4] top-level scope
   @ /tmp/resolver/bin/resolve.jl:335
in expression starting at /tmp/resolver/bin/resolve.jl:335
```
does anyone know what that means?

jClugstor:
As far as I can tell all of the test failures that are different than what's failing on master are the downgrade issues

docs/src/basics/solve.md
```diff
@@ -36,3 +36,32 @@ These are exclusively available for native `NonlinearSolve.jl` solvers.
...
     stored in the solution object. (Defaults to `Val(false)`)
+
+## Verbosity Controls
+
+NonlinearSolve.jl provides fine-grained control over diagnostic messages, warnings, and errors
+through the `verbose` keyword argument. The verbosity system allows you to control what
+information is displayed during the solve process. See [SciMLLogging.jl](https://docs.sciml.ai/SciMLLogging/dev/) for more details. 
+
+```@docs
+NonlinearVerbosity
```
ChrisRackauckas, line 47: This goes in a manual page. Can you make a follow up PR? Just the quick introduction goes here, the docstring with more information on all of the controls needs a manual page.

lib/BracketingNonlinearSolve/src/brent.jl
```diff
@@ -33,9 +43,9 @@ function SciMLBase.__solve(
     end
 
     if sign(fl) == sign(fr)
-        verbose &&
-            @warn "The interval is not an enclosing interval, opposite signs at the \
-                   boundaries are required."
+        @SciMLMessage("The interval is not an enclosing interval, opposite signs at the \
```
ChrisRackauckas, line 46: Does this support Bool? Or do all of the downstream packages need to lower bound to the new NonlinearSolveBase?

ChrisRackauckas: Oh I see the lower bounds, perfect.

docs/src/basics/solve.md
jClugstor: Yes, will do

jClugstor: https://github.com/SciML/NonlinearSolve.jl/pull/722

lib/NonlinearSolveBase/ext/NonlinearSolveBaseLinearSolveExt.jl
```diff
@@ -5,21 +5,22 @@ using ArrayInterface: ArrayInterface
 using CommonSolve: CommonSolve, init, solve!
 using LinearSolve: LinearSolve, QRFactorization, SciMLLinearSolveAlgorithm
 using SciMLBase: ReturnCode, LinearProblem, LinearAliasSpecifier
+using SciMLLogging: @SciMLMessage
 
 using LinearAlgebra: ColumnNorm
 
-using NonlinearSolveBase: NonlinearSolveBase, LinearSolveJLCache, LinearSolveResult, Utils
+using NonlinearSolveBase: NonlinearSolveBase, LinearSolveJLCache, LinearSolveResult, Utils, NonlinearVerbosity
```
ChrisRackauckas, line 12: Is this not passing down the linearverbosity? Seems incorrect.

PR #648, opened 2025-07-26
Hardcode `get_tolerance` for `Float64` for trimming

RomeoV:
This is definitely a bug somewhere, but without this `@report_opt` or `@report_trim` reports errors here.

## Checklist

### DIFF
--- a/lib/NonlinearSolveBase/src/common_defaults.jl
+++ b/lib/NonlinearSolveBase/src/common_defaults.jl
@@ -39,6 +39,11 @@ function get_tolerance(::Nothing, ::Type{T}) where {T}
     η = real(oneunit(T)) * (eps(real(one(T)))^(4 // 5))
     return get_tolerance(η, T)
 end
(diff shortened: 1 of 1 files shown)


oscardssmith (changes requested):
I think we can probably do better here. this works, but there's really no reason we should be doing a rational number power for a heuristic like this.

ChrisRackauckas:
Can you give a concrete suggestion? I think hardcoding Float64 and Float32 is reasonable.

oscardssmith:
what is this number actually used for? Default nonlinear solve tolerance?

ChrisRackauckas:
Yes. And JET really dislikes it. So, let's do it and then if you have concrete improvements let's follow up, but it's strictly better to just add this for now.

oscardssmith:
right, but is there a reason we aren't just using `100*eps()`? What's special about x^0.8?

ChrisRackauckas:
I think @avik-pal had a reason for that.

PR #665, opened 2025-08-02
Add `trim` tests

RomeoV:
Add a `trim` directory with basic tests for trimming.
I tried to imitate the code/package loading structure of SciMLBase.

There are a few quirks here, most of which hopefully will resolve themselves soon:
1. We need to dev seven packages to appease JET and juliac. Each package's reason is documented in the new `test/trim/Project.toml` file. Also, when the upstream branches are merged and deleted, this will fail. But that's good so we can bump the compat accordingly.
2. The testing code includes `clean_optimization.jl` and `trimmable_optimization.jl`. One day hopefully both will be trimmable. At the moment both versions pass JET (impressive!) but only the heavily preallocating version passes `juliac`.
3. In order to run `juliac` we have to put the code to trim into a package. It seems a module will not do. This is documented in the `src/TrimTest.jl` file.
4. This is all Julia 1.12.0-rc1 only. But the `runtests.jl` should take care of that.

### DIFF
--- a/test/trim/Project.toml
+++ b/test/trim/Project.toml
@@ -30,6 +30,9 @@ SciMLBase = {url = "https://github.com/AayushSabharwal/SciMLBase.jl", rev="as/fi
 [compat]
 ADTypes = "1.15.0"
 DiffEqBase = "6.179.0"
+ForwardDiff = "1.0.1"
 LinearAlgebra = "1.12.0"
 NonlinearSolveFirstOrder = "1.6.0"
+Polyester = "0.7.18"
+PolyesterWeave = "0.2.2"
 StaticArrays = "1.9.0"
--- a/test/trim/Project.toml
+++ b/test/trim/Project.toml
@@ -13,12 +13,19 @@ SciMLBase = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 
 [sources]
-ForwardDiff = {rev = "rv/remove-quote-assert-string-interpolation", url = "https://github.com/RomeoV/ForwardDiff.jl"}
-LinearSolve = {rev = "rv/remove-linsolve-forwarddiff-special-path", url = "https://github.com/RomeoV/LinearSolve.jl"}
+# Remove assert that triggers false positive for JET. Tracked at https://github.com/aviatesk/JET.jl/issues/736.
+ForwardDiff = {url = "https://github.com/RomeoV/ForwardDiff.jl", rev="rv/remove-quote-assert-string-interpolation"}
+# Remove special path code path for ForwarDiff of LinearSolve, which uses untrimmable `deepcopy`.
+# Use of `deepcopy` is tracked at https://github.com/SciML/LinearSolve.jl/issues/648.
+LinearSolve = {url = "https://github.com/RomeoV/LinearSolve.jl", rev="rv/remove-linsolve-forwarddiff-special-path"}
 NonlinearSolveFirstOrder = {path = "../../lib/NonlinearSolveFirstOrder"}
-Polyester = {rev = "master", url = "https://github.com/RomeoV/Polyester.jl"}
-PolyesterWeave = {rev = "main", url = "https://github.com/RomeoV/PolyesterWeave.jl"}
-SciMLBase = {rev = "as/fix-jet-opt", url = "https://github.com/AayushSabharwal/SciMLBase.jl"}
+# Remove use of CPUSummary which segfaults trimmed binary. Tracked at https://github.com/JuliaSIMD/Polyester.jl/pull/163.
+Polyester = {url = "https://github.com/RomeoV/Polyester.jl", rev="master"}
+# Remove use of CPUSummary which segfaults trimmed binary. Tracked at https://github.com/JuliaSIMD/PolyesterWeave.jl/pull/28
+PolyesterWeave = {url = "https://github.com/RomeoV/PolyesterWeave.jl", rev="main"}
+# Fix a type instability. Tracked at https://github.com/SciML/SciMLBase.jl/pull/1074.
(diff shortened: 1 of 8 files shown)


RomeoV:
Ready for review.

RomeoV:
I think the two main tests to look here are:
- [pre, ubuntu-latest, GROUP=trim](https://github.com/SciML/NonlinearSolve.jl/actions/runs/16689079835/job/47243822466?pr=665)
- [pre, macos-latest, GROUP=trim](https://github.com/SciML/NonlinearSolve.jl/actions/runs/16689079835/job/47243822471?pr=665)

I don't think anything else should have been impacted...

RomeoV:
The tests are looking good. Checking the CI results from the previous results, everything that passes on master also passes here. Most importantly, [Ubuntu w/ julia-pre](https://github.com/SciML/NonlinearSolve.jl/actions/runs/16689079835/job/47243822466), which is the only one that should actually be affected (besides MacOS julia-pre, which crashes for an unrelated reason).
I just pushed one more commit fixing the formatting, so the whole CI has to run again, but otherwise I'd say it's ready to merge.

ChrisRackauckas:
We shouldn't merge this with stuff still on branches. Need to get the upstreams fixed first.

test/trim/src/TrimTest.jl
```diff
@@ -0,0 +1,27 @@
...
+```julia
+mod MyMod
+    include("optimization_trimmable.jl")
+end
+function (@main)(argv::Vector{String})::Cint
+    MyMod.minimize(1.0)
+    return 0
+end
+```
+segfaults `juliac`. Looking at the segfault stacktrace it seems the culprit is
```
topolarity, line 21: This looks like it's worth a bug report upstream

ChrisRackauckas: Claude wasn't able to reproduce it with an MWE, so @RomeoV can you take this?

test/trim/Project.toml
```diff
@@ -0,0 +1,41 @@
...
+LinearSolve = "7ed4a6bd-45f5-4d41-b270-4a48e9bafcae"
+NonlinearSolveFirstOrder = "5959db7a-ea39-4486-b5fe-2dd0bf03d60d"
+Polyester = "f517fe37-dbe3-4b94-8317-1923a5111588"
+PolyesterWeave = "1d0040c9-8b98-4ee7-8388-3f51789ca0ad"
+SafeTestsets = "1bc83da4-3b8d-516f-aca4-4fe02f6d838f"
+SciMLBase = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
+StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
+
+[sources]
+# Remove assert that triggers false positive for JET. Tracked at https://github.com/aviatesk/JET.jl/issues/736.
```
ChrisRackauckas, line 20: @aviatesk can you look into this?

test/trim/Project.toml
```diff
@@ -0,0 +1,41 @@
...
+# Remove special path code path for ForwarDiff of LinearSolve, which uses untrimmable `deepcopy`.
+# Use of `deepcopy` is tracked at https://github.com/SciML/LinearSolve.jl/issues/648.
+LinearSolve = {url = "https://github.com/RomeoV/LinearSolve.jl", rev="rv/remove-linsolve-forwarddiff-special-path"}
+NonlinearSolveFirstOrder = {path = "../../lib/NonlinearSolveFirstOrder"}
+# Remove use of CPUSummary which segfaults trimmed binary. Tracked at https://github.com/JuliaSIMD/Polyester.jl/pull/163.
+Polyester = {url = "https://github.com/RomeoV/Polyester.jl", rev="master"}
+# Remove use of CPUSummary which segfaults trimmed binary. Tracked at https://github.com/JuliaSIMD/PolyesterWeave.jl/pull/28
+PolyesterWeave = {url = "https://github.com/RomeoV/PolyesterWeave.jl", rev="main"}
+# Fix a type instability. Tracked at https://github.com/SciML/SciMLBase.jl/pull/1074.
+SciMLBase = {url = "https://github.com/AayushSabharwal/SciMLBase.jl", rev="as/fix-jet-opt"}
```
ChrisRackauckas, line 24: @AayushSabharwal was this completed?

AayushSabharwal: No, this is running into the JET failures in LinearAlgebra/`@assert` in ForwardDiff I mentioned in slack. It fixes the type instability relevant here, we just can't test it outside of 1.11.

test/trim/Project.toml
```diff
@@ -0,0 +1,39 @@
...
+NonlinearSolveFirstOrder = "5959db7a-ea39-4486-b5fe-2dd0bf03d60d"
+Polyester = "f517fe37-dbe3-4b94-8317-1923a5111588"
+PolyesterWeave = "1d0040c9-8b98-4ee7-8388-3f51789ca0ad"
+SafeTestsets = "1bc83da4-3b8d-516f-aca4-4fe02f6d838f"
+SciMLBase = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
+StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
+
+[sources]
+# Remove assert that triggers false positive for JET. Tracked at https://github.com/aviatesk/JET.jl/issues/736.
+ForwardDiff = {url = "https://github.com/RomeoV/ForwardDiff.jl", rev="rv/remove-quote-assert-string-interpolation"}
```
topolarity, line 21: This should no longer be necessary for `--trim` w/ https://github.com/JuliaLang/julia/pull/59298 (although JET will probably still false-positive)

RomeoV:
When the SciMLBase PR is closed I'll refactor this using https://github.com/JuliaLang/JuliaC.jl instead of my custom build steps.

ChrisRackauckas:
Rebased to finish in https://github.com/SciML/NonlinearSolve.jl/pull/734

PR #669, opened 2025-08-08
Migrate `solve` dispatches from DiffEqBase to NonlinearSolveBase

jClugstor:
## Checklist


lib/NonlinearSolveFirstOrder/src/NonlinearSolveFirstOrder.jl
```diff
@@ -12,7 +12,7 @@ using LineSearch: BackTracking
 using StaticArraysCore: SArray
 
 using CommonSolve: CommonSolve
-using DiffEqBase: DiffEqBase    # Needed for `init` / `solve` dispatches
+#using DiffEqBase: DiffEqBase    # Needed for `init` / `solve` dispatches
```
ChrisRackauckas: not a comment, just delete.

lib/NonlinearSolveBase/src/solve.jl
```diff
@@ -1,3 +1,449 @@
+const allowedkeywords = (:dense,
```
ChrisRackauckas: The utilities at the top here are reused by DiffEqBase, NonlinearSolveBase, OptimizationBase (in the near future), etc. Let's move these to SciMLBase and share them across the different repos.

ChrisRackauckas: The allowedkwargs handling, the special error types, etc.

jClugstor: I thought so too, but I figured it might be nice for each to have a separate list of acceptable keywords, so if you tried to pass e.g. `dtmin` in to a NonlinearSolve it would give a nice error. Also the messages right now are specific to DiffEqs, they point to the DiffEq docs. I made it so that if there's a mismatch kwarg for a nonlinear solve it points to the NonlinearSolve docs. 

But yeah if these are better in SciMLBase I can just put them there.

jClugstor: There's also a lot of small utility functions that live in DiffEqBase right now, e.g. 
`get_concrete_problem`, `extract_alg`, `promote_u0`, `anyeltypedual` etc. that get used in the solve functions. I'm thinking those should also be moved to SciMLBase?

ChrisRackauckas: The list of acceptable kwargs, yes that one should be problem specific, that would be a nice improvement.

But the others yes.

jClugstor: Currently the definitions for functions like `promote_u0` and `anyeltypedual` etc. are in different extensions in DiffEqBase. I'm thinking I should create corresponding extensions in SciMLBase for those packages and move the functions to those. But I'll keep the parts of the extensions specific to DiffEqs in the extensions in DiffEqBase.

ChrisRackauckas:
sources should handle that on v1

Project.toml
```diff
@@ -137,6 +119,7 @@ julia = "1.10"
 Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
 BandedMatrices = "aae01518-5342-5314-be14-df237901396f"
 BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
+DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e"
```
ChrisRackauckas: why is this needed in tests?

lib/NonlinearSolveBase/test/runtests.jl
```diff
@@ -7,12 +7,14 @@ using InteractiveUtils, Test
...
     @testset "Aqua" begin
         using Aqua, NonlinearSolveBase
+        using NonlinearSolveBase: AbstractNonlinearProblem, NonlinearProblem
 
         Aqua.test_all(
             NonlinearSolveBase; piracies = false, ambiguities = false, stale_deps = false
         )
         Aqua.test_stale_deps(NonlinearSolveBase; ignore = [:TimerOutputs])
-        Aqua.test_piracies(NonlinearSolveBase)
+        #ENSEMBLE PROBLEM SHOULD BE REMOVED, THIS IS TEMPORARY FOR TESTS
```
ChrisRackauckas: ?

Project.toml
jClugstor: Ah yeah, it's not used anywhere in the tests, so not needed anymore.

jClugstor:
Alright, there is now absolutely zero dependency on DiffEqBase.

PR #680, opened 2025-08-15
Rm `kwargs`

### LINKED ISSUE #679: Extra `kwargs` is invalid code
Check
https://github.com/SciML/NonlinearSolve.jl/blob/b37b31b22a6650c773a4fdef8ca220f4e8815051/lib/NonlinearSolveFirstOrder/src/solve.jl#L269

There is an extra `kwargs` that's never defined anywhere and thus invalid code. I guess this line is not triggered by CI? Either way, JET complains (rightfully so).

RomeoV:
This is nevery defined anywhere and thus invalid code and trips up JET.

Fixes #679 

## Checklist

### DIFF
--- a/lib/NonlinearSolveFirstOrder/src/solve.jl
+++ b/lib/NonlinearSolveFirstOrder/src/solve.jl
@@ -266,7 +266,7 @@ function InternalAPI.step!(
             end
             # In the 2nd call the `new_jacobian` is guaranteed to be `true`.
             cache.make_new_jacobian = true
-            InternalAPI.step!(cache; recompute_jacobian = true, kwargs...)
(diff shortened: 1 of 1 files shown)


lib/NonlinearSolveFirstOrder/src/solve.jl
```diff
@@ -266,7 +266,7 @@ function InternalAPI.step!(
             end
             # In the 2nd call the `new_jacobian` is guaranteed to be `true`.
             cache.make_new_jacobian = true
-            InternalAPI.step!(cache; recompute_jacobian = true, kwargs...)
```
ChrisRackauckas, line 269: should probably be `cache.kwargs...`

PR #683, opened 2025-08-20
remove aliasing from caches

oscardssmith:
Attempt to drastically reduce the amount of aliasing between different caches to reduce the possibility of getting them out of sync (e.g. in `reinit!`). Things aren't totally deduplicated yet, but the goal is to only have `u`, `J`, `fu`, and `du` in a single place each.

### DIFF
--- a/lib/NonlinearSolveBase/src/jacobian.jl
+++ b/lib/NonlinearSolveBase/src/jacobian.jl
@@ -107,7 +107,7 @@ function construct_jacobian_cache(
     @assert !(autodiff isa AutoSparse) "`autodiff` cannot be `AutoSparse` for scalar \
                                         nonlinear problems."
     di_extras = DI.prepare_derivative(f, autodiff, u, Constant(prob.p))
-    return JacobianCache(fu, f, fu, p, stats, autodiff, di_extras)
+    return JacobianCache(u, f, fu, p, stats, autodiff, di_extras)
 end
 
 @concrete mutable struct JacobianCache <: AbstractJacobianCache


oscardssmith:
agreed.

oscardssmith:
@avik-pal

oscardssmith:
CI seems happy (other than downgrade). @avik-pal this look ready to you? The release strategy here is to release a breaking NonlinearSolveBase change and minor changes for the rest of the libraries. Should this PR include those bumps?

lib/NonlinearSolveFirstOrder/src/solve.jl
```diff
@@ -165,7 +170,7 @@ function SciMLBase.__init(
             prob, alg, prob.f, fu, u, prob.p;
             stats, alg.autodiff, linsolve, alg.jvp_autodiff, alg.vjp_autodiff
         )
-        J = jac_cache(nothing)
+        J = jac_cache(u)
```
ChrisRackauckas:
```suggestion
        J = jac_cache.J
```

It doesn't need to actually compute here?

lib/NonlinearSolveBase/src/wrappers.jl
```diff
@@ -101,13 +101,10 @@ function construct_extension_jac(
...
-    else
-        J_no_scalar
-    end
+    J_final(J, u) = copyto!(J, J_no_scalar(u))
+    J_final(u) = J_no_scalar(u)
 
     initial_jacobian isa Val{false} && return J_final
 
-    return J_final, Jₚ(nothing)
+    return J_final, Jₚ(u0)
```
ChrisRackauckas:
```suggestion
    return J_final, Jₚ.J
```

This also wasn't computing before?

lib/NonlinearSolveFirstOrder/src/solve.jl
oscardssmith: unfortunately, these do slightly different things (primarily in the case of JacobianOperator)

lib/NonlinearSolveBase/src/wrappers.jl
```diff
@@ -101,13 +101,10 @@ function construct_extension_jac(
...
-    else
-        J_no_scalar
-    end
+    J_final(J, u) = copyto!(J, J_no_scalar(u))
+    J_final(u) = J_no_scalar(u)
 
     initial_jacobian isa Val{false} && return J_final
 
-    return J_final, Jₚ(nothing)
+    return J_final, J = reused_jacobian(Jₚ, u0)
```
ChrisRackauckas:
```suggestion
    return J_final, reused_jacobian(Jₚ, u0)
```

oscardssmith:
@ChrisRackauckas I think this is finally ready! The CI error is gone locally.

ChrisRackauckas:
That test failure is now just lts because lts doesn't use sources.

oscardssmith:
I'll take that as a no, and make a followup for the bumps.

ChrisRackauckas:
Will these need lower bounds bumps between them?

oscardssmith:
yes. This is a major bump of NonlinearSolveBase and (I'm pretty sure) a minor bump of everything else (and enforcement of the new NonlinearSolveBase version)

PR #684, opened 2025-08-20
Specialize on internalnorm function.

RomeoV:
See https://github.com/JuliaLang/julia/issues/59280#issuecomment-3197802085
and https://github.com/JuliaLang/julia/issues/59215#issuecomment-3201263939.


ChrisRackauckas:
we should probably do this everywehere then

RomeoV:
Actually it looks like most cases of `internalnorm` already have the type parameter, see [here](https://github.com/search?q=repo%3ASciML%2FNonlinearSolve.jl%20internalnorm&type=code). I can go through later and see where it's missing still.

RomeoV:
Looks like it's just one other location. Fixed now.

Issue #706, opened 2025-09-24
SCC does not transfer residuals to solution object


mx-p9a:
**Describe the bug 🐞**

When using the SCC algorithm to solve a ModelingToolkit initialization system, the residuals are not transferred to the solution object.

```julia
julia> iprob = InitializationProblem(sys, 0.0, u0, guesses = guesses(sys))
SCCNonlinearProblem with uType Vector{Float64}. In-place: false
u0: 6-element Vector{Float64}:
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0

julia> isol.alg
SCCNonlinearSolve.SCCAlg{Nothing, Nothing}(nothing, nothing)

julia> isol.retcode
ReturnCode.Success = 1

julia> isol.resid
5-element Vector{Nothing}:
 nothing
 nothing
 nothing
 nothing
 nothing

julia> isol.stats

julia>
```

**Expected behavior**

Output like described in [here](https://docs.sciml.ai/ModelingToolkit/dev/tutorials/initialization/#Numerical-Isolation:-InitializationProblem).

  - Output of `versioninfo()`

```julia
Julia Version 1.11.7
Commit f2b3dbda30a (2025-09-08 12:10 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: macOS (arm64-apple-darwin24.0.0)
  CPU: 12 × Apple M2 Max
  WORD_SIZE: 64
  LLVM: libLLVM-16.0.6 (ORCJIT, apple-m2)
Threads: 1 default, 0 interactive, 1 GC (on 8 virtual cores)
Environment:
  JULIA_PKG_USE_CLI_GIT = true
```

PR #707, opened 2025-09-28
Fix SCC residuals transfer for linear problems

ChrisRackauckas-Claude:
## Summary
- Fixes issue #706 where residuals were not being transferred to solution object when using SCC algorithm
- Computes residuals for linear problems as `resid = A*u - b` 
- Ensures all residuals are properly computed and transferred to the solution object

## Details
Previously, when solving SCC problems that contained LinearProblem components, the residuals were being passed as `nothing` in the `build_linear_solution` call. This resulted in a solution object with `nothing` values in the residuals array.

The fix computes the residuals for linear problems using the formula `resid = A*u - b` after solving, and passes them to `build_linear_solution`.

## Test plan
- [x] Added a new test case `"SCC Residuals Transfer"` that verifies:
  - Residuals are not `nothing`
  - No `nothing` values exist in the residuals array
  - Residuals have the correct length
  - Residuals are near zero for converged solutions
  - Residuals match manually computed values
- [x] Verified existing tests still pass
- [x] Tested locally with sample SCC problem containing both linear and nonlinear components


ChrisRackauckas-Claude:
Updated the fix based on feedback. The LinearSolution object does have a  field, but it's often  by default. The updated fix now checks if  is  and only computes the residuals manually when needed using . This is more robust and handles both cases where the linear solver might or might not provide residuals.

ChrisRackauckas-Claude:
Updated the fix based on feedback. The LinearSolution object does have a `resid` field, but it's often `nothing` by default. The updated fix now checks if `sol.resid` is `nothing` and only computes the residuals manually when needed using `resid = A*u - b`. This is more robust and handles both cases where the linear solver might or might not provide residuals.

Project.toml
```diff
@@ -23,6 +23,7 @@ NonlinearSolveSpectralMethods = "26075421-4e9a-44e1-8bd1-420ed7ad02b2"
 PrecompileTools = "aea7be01-6a6a-4083-8856-8a6e6704d82a"
 Preferences = "21216c6a-2e73-6563-6e65-726566657250"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
+SCCNonlinearSolve = "9dfe8606-65a1-4bb3-9748-cb89d1561431"
```
ChrisRackauckas:
```suggestion
```

PR #714, opened 2025-10-08
Add explicit error for Simple algorithms iterator interface

### LINKED ISSUE #712: Proper error message for "Simple" implementations
**Is your feature request related to a problem? Please describe.**

From my understanding the Simple algorithms do not support the iterator interface. However, they do not fail with an error message indicating this. Instead, Julia throws an error that some field in the cache cannot be accessed (`force_stop`).

**Describe the solution you’d like**

A clear and concise error message could help users to understand they hit some unimplemented code path.

**Describe alternatives you’ve considered**

I think the alternative, to implement the iterator interface, defeats the purpose of the simple implementation. If not, then implementing it would be an alternative.

**Additional context**

-

ChrisRackauckas-Claude:
## Summary

Fixes #712 

This PR adds a clear error message when users attempt to use the iterator interface (`init`/`step!`) with Simple algorithms from SimpleNonlinearSolve.jl.

## Problem

Previously, when users tried to use Simple algorithms with the iterator interface, they received a confusing error about a missing `force_stop` field. This made it unclear what the actual problem was.

## Solution

Added an override of `SciMLBase.init()` for `AbstractSimpleNonlinearSolveAlgorithm` that throws a descriptive error message explaining:
- Simple algorithms don't support the iterator interface
- The iterator interface is only available for full-featured algorithms from NonlinearSolve.jl
- Users should either use algorithms like `NewtonRaphson()`, `TrustRegion()`, etc., or use `solve(prob, alg)` directly

## Changes

- `lib/SimpleNonlinearSolve/src/SimpleNonlinearSolve.jl`: Added `SciMLBase.init()` override with clear error message
- `lib/SimpleNonlinearSolve/test/core/iterator_interface_tests.jl`: Added tests to verify error behavior

## Testing

All Simple algorithms tested:
- ✅ SimpleNewtonRaphson
- ✅ SimpleBroyden
- ✅ SimpleTrustRegion
- ✅ SimpleDFSane
- ✅ SimpleKlement
- ✅ SimpleLimitedMemoryBroyden

Tests verify:
- Error is thrown when calling `init(prob, alg)`
- Error message contains helpful guidance
- `solve(prob, alg)` still works correctly

## Example Error Message

```julia
julia> init(prob, SimpleNewtonRaphson())
ERROR: The Simple algorithms from SimpleNonlinearSolve.jl do not support the iterator interface (init/step!).

The iterator interface is only available for the full-featured algorithms from NonlinearSolve.jl.

If you need the iterator interface, please use one of the following algorithms instead:
- NewtonRaphson()
- TrustRegion()
- LevenbergMarquardt()
- And other algorithms from NonlinearSolve.jl

If you want to solve the problem directly, use `solve(prob, alg)` instead of `init(prob, alg)`.

Algorithm attempted: SimpleNewtonRaphson{Nothing}
```

### DIFF
--- a/lib/SimpleNonlinearSolve/src/SimpleNonlinearSolve.jl
+++ b/lib/SimpleNonlinearSolve/src/SimpleNonlinearSolve.jl
@@ -124,6 +124,29 @@ function simplenonlinearsolve_solve_up(
     return SciMLBase.__solve(prob, alg, args...; kwargs...)
 end
 
+# Iterator Interface Error Handling
+# Simple algorithms do not support the iterator interface (init/step!)
+function SciMLBase.init(
+        prob::Union{NonlinearProblem, NonlinearLeastSquaresProblem},
+        alg::AbstractSimpleNonlinearSolveAlgorithm, args...; kwargs...
+)
+    error("""
+    The Simple algorithms from SimpleNonlinearSolve.jl do not support the iterator interface (init/step!).
+
+    The iterator interface is only available for the full-featured algorithms from NonlinearSolve.jl.
+
+    If you need the iterator interface, please use one of the following algorithms instead:
+    - NewtonRaphson()
+    - TrustRegion()
+    - LevenbergMarquardt()
+    - And other algorithms from NonlinearSolve.jl
+
+    If you want to solve the problem directly, use `solve(prob, alg)` instead of `init(prob, alg)`.
+
+    Algorithm attempted: $(typeof(alg))
+    """)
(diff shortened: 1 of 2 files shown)


ChrisRackauckas-Claude:
Updated the test to use `@test_throws` pattern instead of try-catch blocks, as requested. The test is automatically discovered by `@run_package_tests` in runtests.jl via the `@testitem` macro with `tags=[:core]`.

PR #716, opened 2025-10-22
Make string interpolations `lazy`

RomeoV:
This is to appease JET and get all the paths ready for trimming.

### DIFF
--- a/ext/NonlinearSolvePETScExt.jl
+++ b/ext/NonlinearSolvePETScExt.jl
@@ -35,7 +35,7 @@ function SciMLBase.__solve(
         end
 
         if petsclibidx === nothing
-            error("No compatible PETSc library found for element type $(T). Pass in a \
+            error(lazy"No compatible PETSc library found for element type $(T). Pass in a \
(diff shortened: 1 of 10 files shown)


lib/NonlinearSolveBase/src/linear_solve.jl
```diff
@@ -100,7 +100,7 @@ fix_incompatible_linsolve_arguments(A, b, u) = u
 fix_incompatible_linsolve_arguments(::SArray, ::SArray, u::SArray) = u
 function fix_incompatible_linsolve_arguments(A, b, u::SArray)
     (Core.Compiler.return_type(\, Tuple{typeof(A), typeof(b)}) <: typeof(u)) && return u
-    @warn "Solving Linear System A::$(typeof(A)) x::$(typeof(u)) = b::$(typeof(u)) is not \
+    @warn lazy"Solving Linear System A::$(typeof(A)) x::$(typeof(u)) = b::$(typeof(u)) is not \
```
RomeoV: FYI, this is actually the only instance I care about. But I changed all the other occurences since I was already at it.

PR #737, opened 2025-11-17
fix Non-Numeric-u0

oscardssmith:
fixes https://github.com/SciML/NonlinearSolve.jl/issues/736, introduced by https://github.com/SciML/NonlinearSolve.jl/pull/669


oscardssmith:
possibly? I still would expect Aqua to find this.

ChrisRackauckas (approved):

Issue #319, opened 2023-12-12
Inexact Newton Steps in Levenberg Marquardt


avik-pal:
See the last part of http://ceres-solver.org/nnls_solving.html#levenberg-marquardt. This is basically already there and this is mostly a internal solver selection mechanism:

1. Add a forcing function to LM
2. Add a `newton_step = Val(:exact) / Val(:inexact)` option. Set it to nothing by default
3. Default to exact for factorization, else inexact
4. Inexact Newton is just setting the tolerance correctly in `dolinsolve`.

avik-pal:
https://www.sciencedirect.com/science/article/pii/S0377042705007958

avik-pal:
With the redesign this is very easy to do now, we need to add forcing terms to the newton-style descent algorithms and all algorithms will be able to use it

PR #738, opened 2025-11-17
Eisenstat-Walker Newton-Krylov solver

termi-official:
Needs a new LinearSolve.jl release first.

## Checklist

## Additional context

Related issue is here https://github.com/SciML/NonlinearSolve.jl/issues/319 . I left the default to not add any forcing.

### DIFF
--- a/lib/NonlinearSolveBase/Project.toml
+++ b/lib/NonlinearSolveBase/Project.toml
@@ -1,7 +1,7 @@
 name = "NonlinearSolveBase"
 uuid = "be0214bd-f91f-a760-ac4e-3421ce2b2da0"
 authors = ["Avik Pal <avikpal@mit.edu> and contributors"]
-version = "2.4.0"
+version = "2.5.0"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
--- a/lib/NonlinearSolveBase/src/verbosity.jl
+++ b/lib/NonlinearSolveBase/src/verbosity.jl
@@ -133,8 +133,10 @@ function NonlinearVerbosity(;
     end
 
     # Build arguments using NamedTuple for type stability
+    # Use None() for linear_verbosity by default since BLAS errors are not fatal
+    # in the nonlinear solver context (the solver handles singular matrices gracefully)
     default_args = (
-        linear_verbosity = linear_verbosity === nothing ? Minimal() : linear_verbosity,
+        linear_verbosity = linear_verbosity === nothing ? None() : linear_verbosity,
         non_enclosing_interval = WarnLevel(),
         alias_u0_immutable = WarnLevel(),
         linsolve_failed_noncurrent = WarnLevel(),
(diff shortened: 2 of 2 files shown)


lib/NonlinearSolveBase/src/NonlinearSolveBase.jl
```diff
@@ -95,6 +95,8 @@ export RelTerminationMode, AbsTerminationMode,
 export DescentResult, SteepestDescent, NewtonDescent, DampedNewtonDescent, Dogleg,
        GeodesicAcceleration
 
+export EisenstatWalkerForcing2
```
oscardssmith: ?

termi-official: This is "choice 2" from the original paper (to get things started).

lib/NonlinearSolveFirstOrder/src/eisenstat_walker.jl
```diff
@@ -0,0 +1,99 @@
...
+    # Safeguard 2 to prevent over-solving
+    if cache.p.safeguard
+        ηsg = γ*ηprev^α
+        if ηsg > cache.p.safeguard_threshold && ηsg > cache.η
+            cache.η = ηsg
+        end
+    end
+
+    # Far away from the root we also need to respect η ∈ [0,1)
+    cache.η = clamp(cache.η, 0.0, cache.p.ηₘₐₓ)
```
oscardssmith, line 72: how does this enforce "far away from the root"?

termi-official: If you are outside of the convergence radius, you might encounter the scenario that your residual grows. In this case your eta grows accordingly and you might end up with some weird relative tolerance for your linear solver of 1e-3, which will definitely mess up your nonlinear solve. However, if you limit eta, then the solver at least has a chance to recover and climb towards the root.

oscardssmith: right, but isn't that just `clamp(cache.η, 0.0, Inf)`? why do we need to upper bound the convergance rate?

termi-official: The central equation for the analysis is $$||R(u_k) + R'(u_k) \delta u_k|| \leq \eta ||R(u_k)||$$ . If you allow $\eta \geq 1$ then you essentially allow the case where the residual grows. Which essentially translates to "we move away from the solution". Does that make sense?

oscardssmith: oh, I got mixed up with which direction η went. I was thinking of it as the inverse where η>1 was superlinear convergance.

ChrisRackauckas-Claude:
Rebased version of this PR is now at #754 with conflicts resolved against the new `@verbosity_specifier` macro-based verbosity system in master.

PR #749, opened 2025-12-17
Handle edge case values for abstol and span with ITP

dcourteville:
## Checklist

## Additional context

Allows ITP to handle very small `abstol` (and in particular `abstol=0.0`) and very big intervals. This is required to use it in DiffEqBase.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -22,8 +22,10 @@ NonlinearSolveQuasiNewton = "9a2c21bd-3a47-402d-9113-8faf9a0ee114"
 NonlinearSolveSpectralMethods = "26075421-4e9a-44e1-8bd1-420ed7ad02b2"
 PrecompileTools = "aea7be01-6a6a-4083-8856-8a6e6704d82a"
 Preferences = "21216c6a-2e73-6563-6e65-726566657250"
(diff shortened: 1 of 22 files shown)


lib/BracketingNonlinearSolve/src/itp.jl
```diff
@@ -96,8 +96,13 @@ function SciMLBase.__solve(
...
     k1 = alg.scaled_k1 * span^(1 - k2) # k1 > 0
     n0 = alg.n0
-    n_h = exponent(span / (2 * ϵ))
-    ϵ_s = ϵ * exp2(n_h + n0)
+    if span / 2 > ϵ * floatmax(typeof(span))
+        # Workaround for when span / (2 * ϵ) == Inf
+        ϵ_s = span / 2 * exp2(n0)
+    else
+        n_h = exponent(span / (2 * ϵ))
+        ϵ_s = ϵ * exp2(n_h) * exp2(n0)
```
oscardssmith, line 104: why isn't this still `exp2(n_h+n0)`? I would expect that to be faster.

oscardssmith:
yeah this makes sense. It is somewhat annoying that this is parameterized a little bit inefficiently, but realistically this won't cost any real performance.

PR #764, opened 2025-12-30
Ensure consistent left/right solution brackets for bracketing solvers

dcourteville:
## Checklist

## Additional context

cf. https://github.com/SciML/DiffEqBase.jl/pull/1234 & https://github.com/SciML/NonlinearSolve.jl/issues/765

### DIFF
--- a/lib/BracketingNonlinearSolve/test/rootfind_tests.jl
+++ b/lib/BracketingNonlinearSolve/test/rootfind_tests.jl
@@ -81,7 +81,7 @@ end
     prob=IntervalNonlinearProblem(quadratic_f, (1.0, 20.0), 2.0)
     prob_lin=IntervalNonlinearProblem(linear_f, (-1.0, 1.0), 0.0)
 
-    @testset for alg in (ITP(),)
+    @testset for alg in (Alefeld(), Bisection(), Brent(), ITP(), Ridder(), nothing)
         sol = solve(prob, alg; abstol = 0.0)
         # Test that solution is to floating point precision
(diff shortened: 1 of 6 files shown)


lib/BracketingNonlinearSolve/test/rootfind_tests.jl
```diff
@@ -74,15 +75,27 @@ end
...
     end
+end
+
+@testitem "Zero-Tolerance Tests Interval Methods" setup=[RootfindingTestSnippet] tags=[:core] begin
+    prob=IntervalNonlinearProblem(quadratic_f, (1.0, 20.0), 2.0)
+    prob_lin=IntervalNonlinearProblem(linear_f, (-1.0, 1.0), 0.0)
 
-    # Some solvers support abstol=0.0 and converge to floating point precision
-    @testset for alg in (Bisection(), ITP(), Brent())
+    @testset for alg in (ITP(),)
```
ChrisRackauckas: Why were the other methods removed from this? I would assume you'd expand this now?

dcourteville: Bisection and Brent have weird behavior when they step on the exact root, so some of the new tests fails. This is not a difficult fix but I would rather open a separate issue and PR for supporting `abstol=0.0` with all solvers.

dcourteville:
Now all solvers pass the zero-tolerance tests except Muller (because it's not an actual bracketing solver) and Falsi (because it's not numerically stable at very low tolerance).

PR #785, opened 2026-01-16
Restore EnzymeRules on 1.12

wsmoses:
In the future these shouldn't be version gated as rule defns shouldnt ever cause issues


ChrisRackauckas:
I guess we can play it by ear

PR #790, opened 2026-01-20
Pass u and p from nonlinear solve to inner linear solve

termi-official:
## Checklist

## Additional context

For some preconditioners (e.g. GMG, ORAS, ...) we require access to $u$ and potentially the problem parameters like the current time point in a Runge-Kutta stage. This PR adds a simple way to pass this information down.

### DIFF
--- a/lib/NonlinearSolveBase/ext/NonlinearSolveBaseLinearSolveExt.jl
+++ b/lib/NonlinearSolveBase/ext/NonlinearSolveBaseLinearSolveExt.jl
@@ -82,8 +82,8 @@ function LinearSolve.update_tolerances!(cache::LinearSolveJLCache; kwargs...)
     return LinearSolve.update_tolerances!(cache.lincache; kwargs...)
 end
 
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
that seems new

termi-official:
So, I really have trouble reproducing this locally. The only difference I can spot between the red and green CI runs is the order of test execution. The failing tests have the single failing test executed last. Looks like if the test "No PolyesterForwardDiff for SArray" precedes the "Complex Valued Problems: Single-Shooting" (only when using ReTestItems) then some different code path into `DifferentiationInterfacePolyesterForwardDiffExt` of `DifferentiationInterface`is taken, which is not taken if PolyesterForwardDiff is not loaded.

However, I cannot reproduce this locally as I do not know how to enforce that the test item "Complex Valued Problems: Single-Shooting" to be last. Just executing the tests one by one does not trigger the issue - the erroring code path is simply never taken.

termi-official:
I got an MWE in https://github.com/SciML/NonlinearSolve.jl/issues/791 but no exact reproducer of the CI failure. In fact, I can run the CI locally on two different linux machines without triggering the failure in Julia 1.11.8 .

PR #794, opened 2026-01-22
Fix linear solver reinit

termi-official:
Fixes parts of the failing CI in OrdinaryDiffEq (e.g. https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/21210163979/job/61016380710?pr=2899)

### DIFF
--- a/lib/NonlinearSolveBase/src/polyalg.jl
+++ b/lib/NonlinearSolveBase/src/polyalg.jl
@@ -104,11 +104,17 @@ function Base.show(io::IO, ::MIME"text/plain", cache::NonlinearSolvePolyAlgorith
     return NonlinearSolveBase.show_nonlinearsolve_cache(io, cache.caches[cache.current], 4)
 end
 
+function SciMLBase.reinit!(cache::NonlinearSolvePolyAlgorithmCache; kwargs...)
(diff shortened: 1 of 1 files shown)


lib/NonlinearSolveBase/src/linear_solve.jl
```diff
@@ -77,7 +77,7 @@ function construct_linear_solver(alg, linsolve, A, b, u, p; stats, kwargs...)
 
     u_fixed = fix_incompatible_linsolve_arguments(A, b, u)
     @bb u_cache = copy(u_fixed)
-    linprob = LinearProblem(A, b, LinearSolveParameters(u, p); u0 = u_cache)
+    linprob = LinearProblem(A, b, LinearSolveParameters(u_fixed, p); u0 = u_cache)
```
termi-official, line 80: Just to make sure I get it one right this time: Does this (and the reinit above) guarantee, for in-place solvers, that the current solution of the nonlinear solver is aliased into the LinearSolveParameters object and hence into the preconditioner constructor? Not having this guarantee will yield very hard to track convergence issues in the future.

ChrisRackauckas: It should

PR #814, opened 2026-02-13
Add support for transforming bounded parameters into unbounded ones

JamesWrigley:
This adds support for bounds for any algorithm that doesn't natively support it. As suggested in https://github.com/SciML/CurveFit.jl/issues/84. I didn't use TransformVariables.jl since the transform seems like a fairly straightforward thing to implement. I did add a dependency on LogExpFunction since that has robust versions of `logit` and `logistic`, but that's already a transitive dependency through ForwardDiff so it shouldn't increase load times. PreallocationTools is likewise also a dependency of SciMLBase.

Written with a lot of help from Claude :robot: 

## Checklist

### DIFF
--- a/lib/NonlinearSolveBase/src/bounds_transform.jl
+++ b/lib/NonlinearSolveBase/src/bounds_transform.jl
@@ -39,7 +39,7 @@ end
 function _clamp_to_bounds(u, lb, ub)
     has_lb = isfinite(lb)
     has_ub = isfinite(ub)
-    eps_frac = eps(typeof(u))^(3/4)
+    eps_frac = eps(typeof(u))^(3 / 4)
     if has_lb && has_ub
         # Margin scales with interval width
         margin = (ub - lb) * eps_frac
--- a/test/bounds_tests.jl
+++ b/test/bounds_tests.jl
@@ -95,11 +95,11 @@ end
         unbounded_prob = transform_bounded_problem(prob, LevenbergMarquardt())
         @test unbounded_prob.f.f.u_cache isa FixedSizeDiffCache
 
-        unbounded_prob = transform_bounded_problem(prob, LevenbergMarquardt(; autodiff=AutoForwardDiff()))
+        unbounded_prob = transform_bounded_problem(prob, LevenbergMarquardt(; autodiff = AutoForwardDiff()))
         @test unbounded_prob.f.f.u_cache isa FixedSizeDiffCache
 
         # Test Enzyme
-        unbounded_prob = transform_bounded_problem(prob, LevenbergMarquardt(; autodiff=AutoEnzyme()))
+        unbounded_prob = transform_bounded_problem(prob, LevenbergMarquardt(; autodiff = AutoEnzyme()))
         @test unbounded_prob.f.f.u_cache isa typeof(prob.u0)
     end
 


lib/NonlinearSolveBase/src/bounds_transform.jl
```diff
@@ -0,0 +1,124 @@
+# Element-wise scalar transforms between bounded and unbounded spaces.
+# Each function handles 4 cases based on finiteness of lb/ub.
+
+function _to_unbounded(p, lb, ub)
```
ChrisRackauckas: u?

lib/NonlinearSolveBase/src/bounds_transform.jl
```diff
@@ -0,0 +1,124 @@
...
+end
+
+# Clamp a value into the strict interior of [lb, ub] so that _to_unbounded (logit)
+# doesn't receive 0 or 1, which would give ±Inf. Only applied once to u0 before
+# the initial transform — it's a no-op if u0 is already in the interval.
+#
+# We use sqrt(eps) (~1.5e-8 for Float64) as the relative nudge factor. Plain eps
+# (~2.2e-16) is so small that nudged values can round back to the boundary, while
+# sqrt(eps) gives comfortable room without meaningfully changing the starting point.
+function _clamp_to_bounds(p, lb, ub)
```
ChrisRackauckas: u?

lib/NonlinearSolveBase/src/bounds_transform.jl
```diff
@@ -0,0 +1,124 @@
...
+    transformed_u .= _from_unbounded.(u, w.lb, w.ub)
+    return w.f(transformed_u, p)
+end
+
+function (w::BoundedWrapper{true})(resid, u, p)
+    transformed_u = get_tmp(w.u_cache, u)
+    transformed_u .= _from_unbounded.(u, w.lb, w.ub)
+    w.f(resid, transformed_u, p)
+    return resid
+end
```
ChrisRackauckas, line 100: should get algorithm gated, since it shouldn't use a preallocationtools cache if enzyme

test/bounds_tests.jl
```diff
@@ -0,0 +1,120 @@
...
+        sol = solve(prob, alg)
+        @test SciMLBase.successful_retcode(sol)
+        @test sol.u ≈ [true_a, true_b] atol = 1.0e-6
+    end
+
+    @testset "bounded finds true params when in range" begin
+        prob = NonlinearLeastSquaresProblem(
+            nf, u0, x; lb = [0.0, -2.0], ub = [5.0, 0.0]
+        )
+        sol = solve(prob, alg)
```
ChrisRackauckas: add tests where enzyme is used as the ad?

lib/NonlinearSolveBase/src/bounds_transform.jl
JamesWrigley: Fixed in 1f9e07e74.

lib/NonlinearSolveBase/src/bounds_transform.jl
JamesWrigley: Fixed in 1f9e07e74.

lib/NonlinearSolveBase/src/bounds_transform.jl
JamesWrigley: I tried to fix this in 1f9e07e74 but I'm not entirely confident about it. Basically I assume that if `alg` is not passed it'll default to something that defaults to using ForwardDiff for autodiff.

I also enabled the Enzyme tests for 1.12 since Enzyme now officially supports it.

test/bounds_tests.jl
JamesWrigley: Added in 1f9e07e74.

lib/NonlinearSolveBase/src/bounds_transform.jl
```diff
@@ -0,0 +1,139 @@
...
+    else
+        return t
+    end
+end
+
+# Clamp a value into the strict interior of [lb, ub] so that _to_unbounded (logit)
+# doesn't receive 0 or 1, which would give ±Inf. Only applied once to u0 before
+# the initial transform — it's a no-op if u0 is already in the interval.
+#
+# We use sqrt(eps) (~1.5e-8 for Float64) as the relative nudge factor. Plain eps
```
ChrisRackauckas: that still seems large? Is  eps()^(3/4) not more reasaonable?

lib/NonlinearSolveBase/src/bounds_transform.jl
```diff
@@ -0,0 +1,139 @@
...
+@concrete struct BoundedWrapper{isinplace}
+    f
+    lb
+    ub
+    u_cache
+end
+
+function _transform_u(w::BoundedWrapper, u)
+    return if isnothing(w.u_cache)
+        _from_unbounded.(u, w.lb, w.ub)
```
ChrisRackauckas: this allocates. If it's enzyme, the cache should be in the `w` function and be written to.

lib/NonlinearSolveBase/src/bounds_transform.jl
JamesWrigley: Yeah that's fair, changed in 8167dcdb0.

lib/NonlinearSolveBase/src/bounds_transform.jl
JamesWrigley: Ah right I didn't realize Enzyme doesn't use Dual's. Fixed in 8167dcdb0 and added some more tests for the cache type. Do we also need to check Mooncake/ReverseDiff?

PR #818, opened 2026-02-15
Fixed bisection typo to bisecting

oscardssmith:


oscardssmith:
@Proektsoftbg looks like there's still a bug here (looks like there's a division by 2 somewhere there shouldn't be). Do you see the fix? if not, I'll investigate eventually.

PR #821, opened 2026-02-15
Revert "Move `FastShortcutNLLSPolyalg` into NonlinearSolveFirstOrder"

devmotion:
Reverts SciML/NonlinearSolve.jl#813

This requires a new release of NonlinearSolveFirstOrder. Afterwards one can reland the NonlinearSolveFirstOrder changes in a breaking release of NonlinearSolveFirstOrder. And that can be followed up with a new release of NonlinearSolve that supports the breaking release of NonlinearSolveFirstOrder.

Releasing these changes in a non-breaking release of NonlinearSolveFirstOrder breaks any but the latest NonlinearSolve release.


devmotion:
The non-breaking bugfix release of NonlinearSolveFirstOrder will break the latest NonlinearSolve release, though. Maybe the safest option would be to just yank the latest releases of NonlinearSolve and NonlinearSolveFirstOrder. And then release the NonlinearSolveFirstOrder changes in a breaking release - which should have been done in the first place...

devmotion:
I opened JuliaRegistries/General#148468

ChrisRackauckas:
What is the break from this? It just moves the algorithm?

devmotion:
The breaking change in the original PR and releases? Or this PR?

#813 broke precompilation of old releases of NonlinearSolve since if these old releases are used with the latest NonlinearSolveFirstOrder release the same methods are defined both in NonlinearSolve and NonlinearSolveFirstOrder.

Similarly, releasing this PR as a new non-breaking release of NonlinearSolveFirstOrder would break the latest NonlinearSolve release as then these methods wouldn't be defined anymore at all (neither in NonlinearSolve nor in NonlinearSolveFirstOrder).

The main error was merging and releasing #813 - it's not possible to move around algorithms in non-breaking releases without breaking existing releases. One has to make a breaking release of NonlinearSolveFirstOrder with the moved algorithms. Support for this breaking release of NonlinearSolveFirstOrder can then be added in a non-breaking release of NonlinearSolve, in which the algorithm is removed from NonlinearSolve and older versions of NonlinearSolveFirstOrder are disallowed.

The clean way forward here seems to be to yank the latest releases of NonlinearSolve and NonlinearSolveFirstOrder, and then follow the steps (breaking release of NonlinearSolveFirstOrder and a subsequent non-breaking release of NonlinearSolve with adjusted compat bounds) that should have been done in #813.

PR #826, opened 2026-02-16
fix modAB typos

oscardssmith:
built on top of https://github.com/SciML/NonlinearSolve.jl/pull/818 @Proektsoftbg can you take a look to make sure there are no issues? with this, I see better than ITP performance with the first benchmark from https://docs.sciml.ai/SciMLBenchmarksOutput/stable/IntervalNonlinearProblem/suite/

### DIFF
--- a/lib/BracketingNonlinearSolve/src/modAB.jl
+++ b/lib/BracketingNonlinearSolve/src/modAB.jl
@@ -59,17 +59,16 @@ function SciMLBase.__solve(
             # Ordinate of chord at midpoint
             ym = (y1 + y2) / 2
             if 4abs(ym - y3) < abs(ym) + abs(y3)
-                bisection = false
+                bisecting = false
             end
         else
             # Falsi
             x3 = (x1 * y2 - y1 * x2) / (y2 - y1)
             y3 = f(x3)
         end
-
         if iszero(y3)
             return build_exact_solution(prob, alg, x3, y3, ReturnCode.Success)
-        elseif abs(y3) < ϵ
+        elseif (x2-x1) < 2ϵ
             return build_bracketing_solution(prob, alg, x3, y3, x1, x2, ReturnCode.Success)
         end
         x0 = x3
@@ -92,7 +91,7 @@ function SciMLBase.__solve(
             x2, y2 = x3, y3
         end
         if nextfloat(x1) == x2
-            return build_bracketing_solution(prob, alg, x2, y2, x1, x2, ReturnCode.FloatingPointLimit)
+            return build_bracketing_solution(prob, alg, x2, f(x2), x1, x2, ReturnCode.FloatingPointLimit)
         end
         i += 1
         if i >= N #taking longer than expected
--- a/lib/BracketingNonlinearSolve/test/rootfind_tests.jl
+++ b/lib/BracketingNonlinearSolve/test/rootfind_tests.jl
@@ -55,7 +55,7 @@ end
     prob = IntervalNonlinearProblem(quadratic_f, (1.0, 20.0), 2.0)
     ϵ = eps(Float64) # least possible tol for all methods
 
-    @testset for alg in (Bisection(), Falsi(), ITP(), Muller(), ModAB(), nothing)
+    @testset for alg in (Bisection(), Falsi(), ITP(), Muller(), nothing)
         @testset for abstol in [0.1, 0.01, 0.001, 0.0001, 1.0e-5, 1.0e-6]
             sol = solve(prob, alg; abstol)
             result_tol = abs(sol.u - sqrt(2))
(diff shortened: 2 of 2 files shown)


oscardssmith:
The results I'm seeing from SciMLBenchmark are
```
================================================================================
COMPREHENSIVE BENCHMARK SUMMARY
================================================================================
Function                 Roots.jl       Alefeld (BNS)  Bisection (BNS Brent (BNS)    Falsi (BNS)    ITP (BNS)      Ridder (BNS)   ModAB (BNS)
-------------------------------------------------------------------------------------------------------------------------------------------------
Wilkinson-like polynomia 11.9ms         6.5ms          6.2ms          2.6ms          34.4ms         3.5ms          3.4ms          2.1ms
sin(x) - 0.5x            18.3ms         13.3ms         9.5ms          6.2ms          14.7ms         4.4ms          3.9ms          2.9ms
exp(x) - 1 - x - x²/     18.7ms         24.6ms         9.1ms          6.2ms          478.0ms        4.4ms          4.3ms          3.2ms
1/(x-0.5) - 2            12.0ms         4.1ms          6.0ms          2.4ms          75.7ms         3.7ms          2.3ms          1.7ms
log(x) - x + 2           18.8ms         16.8ms         11.4ms         4.7ms          25.5ms         5.3ms          4.8ms          3.9ms
sin(20x) + 0.1x          FAIL           FAIL           273.2ms        261.6ms        256.8ms        249.5ms        255.6ms        263.1ms
x³ - 2x² +               13.2ms         5.5ms          5.5ms          2.4ms          84.0ms         3.3ms          2.4ms          1.7ms
x·sin(1/x) - 0.          19.4ms         FAIL           10.7ms         9.6ms          17.4ms         4.4ms          3.4ms          4.0ms
```
So looks like we're faster than ITP pretty much across the board!

Proektsoftbg:
Hi!

I built the package from your latest code and tested it versus the 91 functions set. It seems to work correctly for all of them and the number of evaluations match the theoretical predictions. I published the source code and the results here:
https://github.com/Proektsoftbg/Numerical/tree/main/Numerical-Julia-NonlinearSolve.jl

Here is a summary of the results:
### Count of function evaluations for 91 function benchmark
Func | bisect | brent | ridder | A42 | ITP | modAB | modAB_CS
-- | --: | --: | --: | --: | --: | --: | --:
SUM | 4366 | 5410 | 3752 | 93976 | 2298 | 1863 | 1783
AVE | 48,0 | 59,5 | 41,2 | 1032,7 | 25,3 | 20,5 | 19,6
REL | 245% | 303% | 210% | 5271% | 129% | 104% | 100%
Func | bisect | brent | ridder | A42 | ITP | modAB | modAB_CS

I also made a rough Julia translation of my latest C# implementation as **modAB_CS** with Claude. It slightly optimized by about 5%, compared to the officially published algorithm. I am going to publish it this year. I will certainly cite NonlinearSolve.jl and your Julia implementation. You can use the new version in future to further tune and optimize the NonlinearSolve.jl version, if you like. You are also free to use my test functions. More details with plots, source references and links is provided here:
https://github.com/Proektsoftbg/Numerical/blob/main/Numerical.Benchmark/doc/Root/Root-test-functions.pdf

Finally, I think that the current implementation is correct, very efficient and ready for publishing. You did a really fantastic job on this! Thank you very much!

oscardssmith:
One followup question: do you think  it be worth having a branch to switch to bisection mode when falsi either fails to do one of shrinking the interval or lowering `y3` by more than bisection would have (in expectation)?

Proektsoftbg:
Hi! I did some experiments with the Weierstrass function, and modAB works here surprisingly well:

image

Back-switching, based on not overshooting the midpoint does not help here as well. Even makes the things worse (151 vs 100).

{7FE8DC01-A0BB-4281-B122-DCD6FBB0422C}

 I suppose that Anderson-Bjork's endpoint shaking does the job better. Howerever, I will dig this little-by-little with time and as soon as I arrive at something, I will let you know.


Here is the source code for the benchmark:
https://gist.github.com/Proektsoftbg/7982df123b6f6e17d92cfc9d73b006e7

Proektsoftbg:
But as you can see by the benchmarks, Anderson-Bjork alone cannot achieve such performance alone without our modification, especially for poorly behaved functions. So, before it was not much better than ITP or Ridder's methods. Now modAB is like AB on steroids. ;)

https://github.com/Proektsoftbg/Numerical/blob/main/Numerical.Benchmark/doc/Root/Root-tests.pdf

Proektsoftbg:
I agree. Can you please fix it in the PR i posted, or I should post another PR.

Proektsoftbg:
sqrt is too sub-optimal. cbrt() is quite fine. OK will run some benchmarks and will report the results.

Proektsoftbg:
OK, Thank you! I will post a pull request with this version.

PR #827, opened 2026-02-17
fix: fix splatting of vector-form `SCCNonlinearProblem`, excessive compilation in `Tuple` form

AayushSabharwal:


lib/SCCNonlinearSolve/src/SCCNonlinearSolve.jl
```diff
@@ -67,6 +67,50 @@ probvec(prob::LinearProblem) = prob.b
 
 iteratively_build_sols(alg, sols; kwargs...) = sols
 
+function solve_single_scc(alg, prob, explicitfun; kwargs...)
+    explicitfun(
+        SymbolicIndexingInterface.parameter_values(prob), sols
```
ChrisRackauckas:
```suggestion
        prob.p, sols
```

PR #828, opened 2026-02-18
switch default bracketing solver to ModAB

oscardssmith:
Tests seem to show it as ~20-30% faster across the board.


oscardssmith:
yeah. The test that failed was testing that the default solver converges slowly (which is no longer true)

PR #859, opened 2026-03-02
Add underdetermined Levenberg-Marquardt support (minimum-norm mode)

ChrisRackauckas-Claude:
# Add underdetermined Levenberg-Marquardt support (minimum-norm mode)

> [!NOTE]

Closes #851.

## Summary

Adds a `:minimum_norm` mode to `DampedNewtonDescent` for underdetermined nonlinear least-squares problems (`length(u) > length(fu)`). Instead of the least-squares/normal-form formulations on the n-sized system, it solves the m×m dual system

1. `(JJᵀ + λD̃ᵀD̃) z = -fu`
2. `δu = Jᵀz`

which, as λ → 0, yields the minimum-norm step solving the linearized equations (Chen et al. (2026), appendix B; suggested by @stevengj in #851).

## Motivation is efficiency, not correctness

Master already solves underdetermined NLLS problems correctly: the default `:least_squares` mode solves `min ‖[J; √(λDᵀD)]δ + [fu; 0]‖` by QR, and the stacked matrix has full column rank for λ > 0, so the LM step is well-defined for m < n. All of the test problems in this PR pass on unmodified master.

The value of the dual formulation is the m×m system when n ≫ m: on a dense problem with n = 4000, m = 10, this PR solves in ~2s vs ~10s on master (QR of a 4010×4000 matrix per iteration vs Cholesky of 10×10 plus matvecs).

## Design: reuse of the existing machinery

The minimum-norm formulation is exactly the `:normal_form` mode of the *dual* system, so the implementation reuses the existing infrastructure with substitutions `JᵀJ → JJᵀ`, `Jᵀfu → fu`, `u → z` (fu-sized):

- The damping function cache is initialized with `fu` in the `u` slot, so the LM damping diagonal `D̃ᵀD̃` is m×m and tracks `diag(JJᵀ)` — Marquardt scaling on the dual system. The `Val(true)` damping `solve!` receives the JJᵀ matrix exactly like `:normal_form` passes JᵀJ.
- `dampen_jacobian!!` is reused unchanged; no new damping helpers.
- The JJᵀ product lives in the existing `JᵀJ_cache` slot; the only new cache field is `z_cache` (m-sized intermediate).
- The linear solver goes through the same `construct_linear_solver`/`lincache` path; A is square m×m symmetric so any user `linsolve` works.

Mode selection: `min_norm_mode = :auto` (default) engages the dual formulation when the system is underdetermined and the damping function returns normal-form damping (i.e. LevenbergMarquardt; PseudoTransient's damping is not normal-form and is unaffected). `:disabled` restores the previous behavior; `:minimum_norm` forces it.

## Verification (all run locally)

- New tests in `lib/NonlinearSolveFirstOrder/test/least_squares_tests__item3.jl` (wired into the SafeTestsets runner): linear/nonlinear/in-place/large underdetermined systems, minimum-norm solution check, mode-engagement check, `:disabled` fallback check, StaticArrays, badly scaled system, overdetermined/square regression tests, the wide-matrix `A \ b` minimum-norm comparison suggested by @stevengj, a rank-deficient wide system, and a CholeskyFactorization run. 27/27 pass.
- Full `Pkg.test()` of NonlinearSolveBase and NonlinearSolveFirstOrder (core group) pass locally on Julia 1.11.
- On a badly scaled underdetermined problem (`f = [1e4 u₁ + u₂ - 1, u₃ + u₄ - 2]`), the damping diagonal now matches `diag(JJᵀ)` exactly and LM converges in 7 iterations — the same count as master's least-squares path — while solving only m×m systems.

## History

The original version of this PR (`8eebe40`) had its own `dampen_jacobian_minimum_norm!!`/`_extract_scalar_damping` helpers which passed the raw m×n Jacobian into the `Val(true)` damping update; LM's `DᵀD` then read `J[i,i]` as if it were `diag(JᵀJ)` and a mean-of-diagonal produced an arbitrary damping scale (2× the iterations of master on badly scaled problems). Its test file also used `@testitem`/ReTestItems style, which the current SafeTestsets-based runner never included, so CI never exercised it. The second commit reworks the implementation as described above; see the commit message for details.

## Semver

`DampedNewtonDescent` gains a documented `min_norm_mode` kwarg and underdetermined LM defaults to the new formulation: NonlinearSolveBase is bumped 2.31.3 → 2.32.0.

https://claude.ai/code/session_01KksSwVBbmEjchzWgbjS5eW


stevengj:
Thanks Chris, I didn't realize that you had worked on this.

cc @aristeidis-karalis and @mochen4

mochen4:
Thanks! This looks great!

stevengj:
A simple test case would be to put in a linear function $f(x) = Ax - b$ for a "wide" matrix $A$ and check that the result of minimizing $\Vert f(x) \Vert_2^2$ starting at $x=0$ matches the minimum-norm solution `A \ b`.

aristeidis-karalis:
@ChrisRackauckas-Claude, it's awesome you implemented this!

Just a comment: 
I could not figure out (from the code) what `D` is in your original (m>n) formulation. Usually, it's best if it's something that scales with `J²`, so that the result is independent from the scale of the problem. Some commonly used options are `D = diag(JᵀJ)` or `D ~ ||J||²∙I` (where `||J||` can be the Frobenius norm).

The same holds for the minimum-norm solution, it's often preferable if it is independent from the scale of `f, J`. Therefore, instead of only the `λI` option, it would be good to give the user the ability to choose `λD` also with, e.g., `D = diag(JJᵀ)` or `D ~ ||J||²∙I`. We used the latter, in particular `D ~ ||J||²/m∙I`

@stevengj

lib/NonlinearSolveBase/src/descent/damped_newton.jl
```diff
@@ -262,3 +385,44 @@ function dampen_jacobian!!(J_cache, J::AbstractMatrix, D::Union{AbstractMatrix,
...
+end
+function _extract_scalar_damping(D::AbstractMatrix)
+    # Extract diagonal and take mean
+    n = minimum(size(D))
+    s = zero(eltype(D))
+    for i in 1:n
+        s += D[i, i]
+    end
+    return s / n
+end
```
stevengj: Couldn't all three of these methods be replaced by
```jl
_extract_scalar_damping(D) = tr(D) / size(D,1)
```
assuming `D` is always a square matrix as seems to be the case?

ChrisRackauckas-Claude:
Rebased onto master (was 160 commits behind; clean rebase) and reworked the implementation in the second commit. Summary of what the review found and changed:

**Findings on the original implementation:**

1. **The damping computation was incorrect.** `LevenbergMarquardtDampingFunction` has `requires_normal_form_jacobian() = false`, so the minimum-norm branch passed the raw m×n Jacobian into the `Val(true)` (normal form) damping `solve!`. That update reads `J[i,i]` as if it were `diag(JᵀJ)`, corrupting `DᵀD`, and the `_extract_scalar_damping` mean-of-diagonal then produced an arbitrary scale. Reproduced on `f = [1e4·u₁ + u₂ - 1, u₃ + u₄ - 2]`: `DᵀD = [1e4, 1e-8, 1e-8, 1e-8]` instead of the correct `[1e8, 1, 1, 1]`, taking 14 iterations where master's least-squares path takes 7.
2. **The tests were never run in CI.** `underdetermined_tests.jl` used `@testitem` (ReTestItems style), but the test runner is SafeTestsets with explicit `include`s and never referenced the file.
3. **Master already solves underdetermined NLLS correctly** via the `:least_squares` mode (QR of the full-column-rank stacked `[J; √(λDᵀD)]`), including all of this PR's test problems. So the feature is an *efficiency* win for n ≫ m (measured ~5x at n=4000, m=10, and asymptotically much more), plus the explicit minimum-norm step bias — not a correctness fix.

**Rework (second commit):** the minimum-norm formulation is treated as the `:normal_form` mode of the dual system, so it now reuses the existing machinery — damping cache initialized with `fu` in the `u` slot (m×m Marquardt scaling tracking `diag(JJᵀ)`), `dampen_jacobian!!` unchanged, `JJᵀ` in the existing `JᵀJ_cache` slot, one new cache field (`z_cache`) instead of three, and no new damping helpers. Also fixed a StaticArrays crash (the `@bb similar`/zero pattern collapses to a scalar for immutable matrices — latent in `:normal_form` on master, exposed here because underdetermined static problems now route to `:minimum_norm`) and added ArgumentError guards for invalid `min_norm_mode` values.

**Verified locally (Julia 1.11):** new 18-test underdetermined suite passes; full `Pkg.test()` of NonlinearSolveBase and NonlinearSolveFirstOrder (core group) pass; damping trajectory on the badly scaled problem now matches master's iteration count while solving only m×m systems.

Note for a follow-up: the same immutable-matrix crash exists latently in the `:normal_form` branch on master (`@bb J_cache = similar(JᵀJ); @bb @. J_cache = 0` → `J_cache = 0::Int` → `copyto!(0, ::SMatrix)`); it is unreachable today only because static-array NLLS problems default to `:least_squares`.

ChrisRackauckas-Claude:
Both review suggestions are now addressed in 509de20, along with a performance and numerics pass over the mode. All numbers below are from local runs (Julia 1.11).

**@stevengj's linear test** — added exactly as suggested: for `f(x) = Ax - b` with wide `A` started at `x = 0`, the converged solution is checked against `A \ b` (at 5×20 and 10×40, StableRNGs-seeded, `atol = 1e-8`; observed agreement is ~1e-13). This holds structurally, not just numerically: every minimum-norm step `δu = Jᵀz` lies in the row space of `A`, so iterates never leave it. Also added: a rank-deficient wide system (duplicated row, consistent rhs) which still recovers `pinv(A)*b`, and a `linsolve = CholeskyFactorization()` run since `JJᵀ + λD̃ᵀD̃` is SPD by construction.

**@aristeidis-karalis on the damping scale** — the reworked implementation uses your first suggested option: the damping applied to the dual system is `λ·D̃ᵀD̃` with `D̃ᵀD̃ = ` running elementwise max of `diag(JJᵀ)` (the same Marquardt scaling the overdetermined path applies to `diag(JᵀJ)`, from Transtrum & Sethna), so the damping scales with `‖J‖²` and the step is scale-invariant. Verified directly: scaling `f` by `1e8` or `1e-8` produces identical iteration counts (108/109 and 19/19 vs the least-squares path). This is now documented in the `DampedNewtonDescent` docstring. A `‖J‖²/m·I`-style option would slot in as an alternative `AbstractDampingFunction` — the descent is generic over that interface — but that's left for a follow-up rather than adding a second knob here.

**Performance** — the descent code allocates zero bytes per iteration; all buffers (`JJᵀ`, `z`, damping diagonal) live in the init-time cache. Measured per-iteration allocation slope on n=2000, m=10 (slope between `maxiters=10` and `maxiters=30` runs, geodesic acceleration on): **2 B/iter** with `linsolve = CholeskyFactorization()`. The default `linsolve = nothing` path was allocating ~13 KB/iter, which traced to a pre-existing one-line routing bug — `Symmetric` dense matrices match the `WrappedArray{<:Any, <:SMatrix}` static-array pattern and fall into the naive `A \ b` fallback, refactorizing every iteration. That's fixed separately in #1000 (it equally affects the `:normal_form` mode on master). With #1000 applied, the remaining default-path overhead (~10 KB/iter) is inside LinearSolve's `DefaultLinearSolver` handling of `Symmetric` and is upstream of this PR; `CholeskyFactorization` avoids it entirely and is the documented recommendation.

**Numerics stress battery** (wide linear systems, vs the `:disabled` QR-based least-squares path):
- cond(J) ≈ 1: converges to the min-norm solution to 4e-15 (QR path converges but to a non-min-norm solution, as expected).
- badly scaled columns (1e6 spread): min-norm solution to 3e-16; QR path returns a solution with ‖u‖ 5000× larger.
- cond(J) ≈ 1e6: still reaches the min-norm solution (rel. err 4e-11); the QR path hits maxiters without finding it.
- cond(J) ≈ 1e10: the dual formulation fails (squared conditioning, cond(JJᵀ) ≈ 1e20) while QR still drives the residual down. This is the known tradeoff of any normal-form approach; it's now stated in the docstring with `min_norm_mode = :disabled` as the escape hatch.
- complex problems: the dual system is built with `transpose`, which is only correct for real arithmetic, so `:auto` now restricts to real eltypes and forcing `:minimum_norm` on a complex problem throws. (LevenbergMarquardt errors on complex input upstream of the descent on master anyway, so this is future-proofing.)

Full `Pkg.test()` of NonlinearSolveBase and NonlinearSolveFirstOrder pass locally with these changes (the underdetermined suite is now 27 tests).

ChrisRackauckas-Claude:
CI triage for the final board (89 pass / 3 fail):

- `ModelingToolkit.jl/InterfaceI` and `InterfaceII`: pre-existing master-borne downstream failures (also failing on master-equivalent branches; MTK-side tracking in [ModelingToolkit.jl#4706](https://github.com/SciML/ModelingToolkit.jl/issues/4706) and runner infrastructure).
- `tests / Adjoint (julia lts, macos-latest)`: segfault inside Julia 1.10's GC (`gc_mark_obj8` during compiler SSA `compact!`) while running `test/Adjoint/adjoint_tests__item3.jl` — which is itself the regression test for the known Enzyme GC-corruption failure mode on macOS LTS (EnzymeAD/Enzyme.jl#3130, #939). Evidence this is the known flake and not this PR:
  - The test drives `Enzyme.autodiff` through `NewtonRaphson()` on a square 1-equation `NonlinearProblem`; `DampedNewtonDescent` (the only code this PR touches) is never on that path.
  - The identical job (same platform, same Julia LTS, same **Enzyme v0.13.173**) passed on #1000's CI run a few hours earlier, and Adjoint passes on this PR on Linux LTS and macOS Julia 1.
  - The crash signature (GC corruption during compilation, nondeterministic) matches exactly what the test's own comment documents for Enzyme.jl#3130.

I don't have rerun rights on this repo, so the job would need a maintainer rerun to go green — or it can be read as the known Enzyme macOS-LTS flake recurring at v0.13.173.

ChrisRackauckas-Claude:
## Scaling benchmarks vs the other least-squares methods

Two problem families, all run locally (Julia 1.11, linux, dense ForwardDiff Jacobians, best of 3 after warmup):

**A. Underdetermined Brusselator steady state** — unknowns are the full (u,v) state on an Ng×Ng periodic grid (n = 2Ng²); the steady-state residuals are enforced on 1/3 of the grid points (m ≈ n/3), classical Brusselator initial guess:

| Ng | n / m | LM min-norm (new default) | LM min-norm + Cholesky | LM `:disabled` (old path) | GaussNewton | TrustRegion |
|---|---|---|---|---|---|---|
| 8 | 128 / 44 | 0.002s (8) | 0.002s (8) | 0.016s (7) | **0.002s (3)** | 0.002s (3) |
| 16 | 512 / 170 | 0.027s (8) | 0.028s (8) | 0.248s (7) | 0.037s (2) | 0.038s (2) |
| 24 | 1152 / 384 | 0.579s (38) | 0.153s (8) | 1.616s (7) | 0.191s (2) | 0.188s (2) |
| 32 | 2048 / 684 | 4.32s (102)† | 2.69s (102)† | 58.8s (102)† | **1.00s (2)** | 1.02s (2) |

(iterations in parens; † = StalledSuccess, LM's trust-region tail stalls at ~2e-8 vs abstol 1e-8 on both old and new paths)

**B. Dense random wide system** (`f = Ax + 0.05x₁:ₘ² − b`), m = 20 fixed, growing n — the n ≫ m regime from #851:

| n | LM min-norm | LM `:disabled` | GaussNewton | TrustRegion |
|---|---|---|---|---|
| 200 | 0.008s | 0.032s | 0.004s | 0.007s |
| 800 | 0.112s | 0.338s | 0.060s | 0.090s |
| 3200 | 1.79s | 5.87s | **0.73s** | 1.49s |

**C. Robustness check** — same Brusselator but started far away (`x0 = 8·rand` and `20·rand`): GaussNewton/TrustRegion still converge in 3–4 iterations; LM needs 9–105. Underdetermined systems are structurally easy for undamped Gauss-Newton: the linearized system is always consistent (no residual-vs-curvature tension as in overdetermined fitting), so damping is rarely needed for convergence.

### Takeaways

1. **This PR fixes LM's underdetermined scaling penalty**: 3–22× vs the old stacked-QR path, growing with size (58.8s → 2.7s at n=2048). This is the PR working as intended — the old path's per-iteration QR of the (m+n)×n stacked matrix is O((m+n)n²) vs O(m²n + m³) for the dual system.
2. **GaussNewton is generically the fastest least-squares method for underdetermined problems**, including from bad initial guesses — its QR solve of the wide m×n Jacobian is also O(m²n) and also produces minimum-norm steps. On these benchmarks LM's damping + geodesic acceleration only add iterations. LM min-norm mode is for users who specifically want LM (its damping dynamics on genuinely nasty problems, or its API), now without the scaling cliff.
3. The Ng=24 default-vs-Cholesky gap (38 vs 8 iterations, same math) is trajectory divergence through the `Symmetric` default-solver fallback that #1000 fixes; with an explicit `CholeskyFactorization` the default-guess iteration counts match the old path. One more reason to land #1000 alongside this.
4. The LM stall at Ng≥24 tight tolerances (StalledSuccess at ~102 iterations, resid ~2e-8) is identical on the old and new paths — pre-existing LM trust-region tail behavior, not introduced here.

ChrisRackauckas-Claude:
Follow-up to the benchmarks (a4e6991): the minimum-norm mode now **defaults to a Cholesky factorization**.

The damped dual system `JJᵀ + λD̃ᵀD̃` is SPD by construction (`JJᵀ` PSD, LM damping adds a strictly positive diagonal), but that knowledge only exists at the descent level — the generic linear-solver default sees just `Symmetric` and must assume possibly-indefinite. New hook `NonlinearSolveBase.default_spd_linsolve(A)`: returns `CholeskyFactorization()` from the LinearSolve extension for real `Symmetric` A, generic default otherwise; user-specified `linsolve` always wins. This also makes the mode take the cached-factorization LinearSolve path regardless of #1000 (an explicit algorithm skips the misrouted `linsolve === nothing` branch).

Benchmark deltas (underdetermined Brusselator, default settings):

| size | generic default (before) | Cholesky default (now) | old `:disabled` path |
|---|---|---|---|
| n=1152 / m=384 | 0.58s (38 it) | **0.16s (8 it)** | 1.6s |
| n=2048 / m=684 | 4.3s | **2.7s** | 47–59s |

The 38-vs-8 iteration anomaly at n=1152 is gone (it was trajectory divergence through the fallback solver). Edge cases re-verified with the Cholesky default: rank-deficient Jacobian converges to `pinv(A)b` at 2e-13, and cond(J)=1e10 behaves identically to the generic default (no hard Cholesky failure — the damping floor keeps the matrix numerically SPD). Full NonlinearSolveBase + NonlinearSolveFirstOrder suites pass; the underdetermined suite (28 tests) now asserts the default lincache is `CholeskyFactorization`.

Issue #535, opened 2025-02-27
GaussNewton algorithm fails with sparse Jacobian


afossa:
**Describe the bug 🐞**

I am trying to solve a nonlinear least squares problem using the `GaussNewton` algorithm. For this, I first build a `NonlinearFunction` object and provide it a function that evaluates the analytical Jacobian via the keyword `jac`, as well as the sparsity pattern of the Jacobian as a `SparseMatrixCSC` via the keyword `jac_prototype`.

With these settings, the algorithm fails to find a solution. However, the same algorithm succeeds when using a dense Jacobian, i.e. `jac_prototype=nothing`. In both cases, the function passed to `jac` is the same.

The problem is also correctly solved with both sparse and dense Jacobians by `LevenbergMarquardt`, and by a simple implementation of the Gauss-Newton algorithm that I wrote myself. 

**Expected behavior**

The algorithm should find the same solution using either the sparse or dense Jacobian.

**Minimal Reproducible Example 👇**

The code is quite involved and I cannot easily share a MWE. Here is how I invoke `NonlinearSolve`:

```julia
# define the nonlinear function
nlf = NonlinearFunction(
	(y, u, p) -> compute_residuals!(y, u, f, p);
	jac=(j, u, p) -> compute_jacobian!(j, u, f, p),
	resid_prototype=zeros(eltype(z), 7 * (length(t) - 1)),
	jac_prototype=sparse ? build_jacobian_prototype(length(t), eltype(z)) : nothing,
)

# setup the nonlinear least squares problem
nlls = NonlinearLeastSquaresProblem(nlf, z, p)

# solve the problem
sol = solve(
	nlls, alg; abstol=abstol, reltol=reltol, maxiters=maxiters, store_trace=Val(true)
)
```

And this is my implementation of the Gauss-Newton algorithm that works as expected:

```julia
# preallocate the residuals vector and the Jacobian matrix
y = Vector{eltype(z)}(undef, 7 * (n - 1))
J = sparse ? build_jacobian_prototype(n, eltype(z)) : zeros(eltype(z), 7 * (n - 1), 8 * n - 1)

# perform differential correction
@printf "\n%-6s %-12s\n" "Iter" "||F(X)||"
iter = 1
while iter <= maxiters

	# compute the residuals and the Jacobian
	compute_residuals!(y, z, f, p)
	compute_jacobian!(J, z, f, p)

	# compute the error
	err = norm(y)
[...]

PR #889, opened 2026-04-02
fix issue #535

Iskaban10:
…or issue #535

## Checklist

## Additional context
The problem lied with the fact that in newton.jl the multiplication with the transpose was happening in-place due to `@bb` and that hampered the sparsity pattern for sparse matrices. Hence illogical step values were observed. The fix is to replace the `@bb` in-place multiplication with a fresh allocation `cache.JᵀJ_cache = transpose(J) * J` when J is sparse.
Would appreciate any advice or corrections.

### DIFF
--- a/lib/NonlinearSolveBase/src/descent/newton.jl
+++ b/lib/NonlinearSolveBase/src/descent/newton.jl
@@ -105,7 +105,15 @@ function InternalAPI.solve!(
         if normal_form(cache)
             @assert !preinverted_jacobian(cache)
             if idx === Val(1)
-                @bb cache.JᵀJ_cache = transpose(J) × J
+                # For sparse J, @bb cannot update entries in-place
++               # It will update entries absent from the initial sparsity pattern of
++               # cache.JᵀJ_cache, corrupting the normal equations.
++               # Fresh memory is therefore allocated in case of sparse matrix
++               if ArrayInterface.isstructured(J)
++                   cache.JᵀJ_cache = transpose(J) * J
++               else
++                   @bb cache.JᵀJ_cache = transpose(J) × J
++               end
             end
             @bb cache.Jᵀfu_cache = transpose(J) × vec(fu)
             @static_timeit cache.timer "linear solve" begin
--- a/lib/NonlinearSolveBase/src/descent/newton.jl
+++ b/lib/NonlinearSolveBase/src/descent/newton.jl
@@ -106,7 +106,7 @@ function InternalAPI.solve!(
             @assert !preinverted_jacobian(cache)
             if idx === Val(1)
                 # For sparse J, @bb cannot update entries in-place
-+               # It will update entries absent from the initial sparsity pattern of
++               # It cannot update entries absent from the initial sparsity pattern of
 +               # cache.JᵀJ_cache, corrupting the normal equations.
 +               # Fresh memory is therefore allocated in case of sparse matrix
 +               if ArrayInterface.isstructured(J)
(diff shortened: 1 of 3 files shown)


lib/NonlinearSolveBase/src/descent/newton.jl
```diff
@@ -105,7 +105,15 @@ function InternalAPI.solve!(
         if normal_form(cache)
             @assert !preinverted_jacobian(cache)
             if idx === Val(1)
-                @bb cache.JᵀJ_cache = transpose(J) × J
```
oscardssmith, line 108: why not `mul!`?

Iskaban10: `mul!` can only update or write in the positions that exist in the cache. For a `SparseMatrixCSC`, those positions were fixed when the matrix was first created. If `JᵀJ` has a nonzero at position (i,j) but the initial matrix has no stored entry there, mul! has nowhere to put that value and will skip it silently.

oscardssmith: Are you sure about that? I'm pretty sure that's not how SparseMatrix works in Julia.

Iskaban10: My apologies for the mistake. My understanding of `SparseMatrix` was wrong. I am still learning and appreciate your patience and guidance.
Will my previous approach of trying to use BackTracking() as the default linesearch in [gauss-newton.jl](https://github.com/SciML/NonlinearSolve.jl/blob/master/lib/NonlinearSolveFirstOrder/src/gauss_newton.jl#L12) be a good fix then?

lib/NonlinearSolveFirstOrder/src/gauss_newton.jl
```diff
@@ -9,7 +9,7 @@ matrices via colored automatic differentiation and preconditioned linear solvers
 for large-scale and numerically-difficult nonlinear systems.
 """
 function GaussNewton(;
-        concrete_jac = nothing, linsolve = nothing, linesearch = missing,
+        concrete_jac = nothing, linsolve = nothing, linesearch = BackTracking(),
```
ChrisRackauckas, line 12: no, the default doesn't have a line search on purpose. If you want a line search, it's documented that you choose one. Putting a default here would have lots of downstream implications. We could make an aliased name if necessary, but then also, you probably don't want this line search.

ChrisRackauckas:
Nothing changed in the source code?

lib/NonlinearSolveFirstOrder/src/gauss_newton.jl
Iskaban10: I was also very skeptical about this. But then what can be the solution to the issue? As you mentioned about aliasing, will implementing a `RobustGaussNewton` with a default LM linesearch be a good option then?

ChrisRackauckas: I mean, a good option for what? It's just a shorthand to a different method. If the answer is the user chose the wrong method then 🤷sure a shorthand could be nice but it's not actually doing anything. The real question is, why is someone choosing a method without a line search when they want a line search? Is it documented as having a line search? Is it documented as robust? It shouldn't be. Everything should be pointing people already to TrustRegion, so how did this even show up in a user code?

Iskaban10: Yupp agreed, a simple GaussNewton wasn't a very good choice in the user code.

Issue #445, opened 2024-06-05
Implementing ForwardDiff.hessian through NonlinearLeastSquaresProblem


IJ05:
I wanted to find the partial double and mixed derivatives of the variables c1_length, c2_length, c3_length with respect to the variables px and py. The first order derivatives of the same, obtained using ForwardDiff.gradient or ForwardDiff.jacobian are working fine, but ForwardDiff.hessian returns a method error. 

Below is a MWE of the concerned functions:
**Minimum working example**

`using NonlinearSolve, ForwardDiff, SciMLSensitivity`
```
function objfn(F,init,params)
    th1,th2 = init
    px,py,l1,l2 = params
    F[1] = l1*cos(th1)+l2*cos(th1+th2)-px
    F[2] = l1*sin(th1)+l2*sin(th1+th2)-py
    return F
end
```

```
function solve_nlprob(pxpy)
    px,py = pxpy
theta1 = pi/4;
theta2 = pi/4
initial_guess = [theta1;theta2];
l1 = 60; l2 = 60;
p = [px.value;py.value;l1;l2];
prob = NonlinearLeastSquaresProblem(NonlinearFunction(objfn, resid_prototype = zeros(2)), initial_guess,p)
resu = solve(prob,reltol = 1e-12, abstol = 1e-12);
th1,th2 = resu.u
cable1_base = [-90;0;0];
cable2_base = [-150;0;0];
cable3_base = [150;0;0];
cable1_top = [l1*cos(th1)/2;l1*sin(th1)/2;0];
cable23_top = [l1*cos(th1)+l2*cos(th1+th2)/2;l1*sin(th1)+l2*sin(th1+th2)/2;0];
c1_length=sqrt((cable1_top[1]-cable1_base[1])^2 + (cable1_top[2]-cable1_base[2])^2)
c2_length=sqrt((cable23_top[1]-cable2_base[1])^2 + (cable23_top[2]-cable2_base[2])^2)
c3_length=sqrt((cable23_top[1]-cable3_base[1])^2 + (cable23_top[2]-cable3_base[2])^2)
    return c1_length
end
```
The output for `grad1 = ForwardDiff.gradient(solve_nlprob,[34.0,87.0])` is
```
2-element Vector{Float64}:
 0.0
 0.0
```
The output error for `hess1 = ForwardDiff.hessian(solve_nlprob,[34.0,87.0])` is a method error which is: 

**Error and Stacktrace**
```
[...]

PR #910, opened 2026-04-11
Drop Vector{Float64} specialization in NonlinearSolveBase tag stamping

ChrisRackauckas-Claude:
## Summary

- Addresses the NonlinearSolveBase side of SciML/OrdinaryDiffEq.jl#3381. A nested ForwardDiff call graph (NonlinearLeastSquares over an ODE solve, the MWE in the issue) crashes with `FirstAutodiffJacError(MethodError(Float64, (::nested_dual,)))` inside the inner Rosenbrock Jacobian. The proximate cause on the NonlinearSolve side is `standardize_forwarddiff_tag` unconditionally substituting a canonical `Tag{NonlinearSolveTag, Float64}` into any `Vector{Float64}` user problem — including FullSpecialize problems — which drags in the precompile-baked `@generated tagcount` literal for that tag and inverts ForwardDiff's nesting precedence against tags generated lazily at runtime for inner solves.
- Removes the `isa Vector{Float64}` gates from both `standardize_forwarddiff_tag` dispatches (AutoForwardDiff and AutoPolyesterForwardDiff) and from `maybe_wrap_nonlinear_f`. The canonical tag is now stamped if and only if the user function was actually wrapped via AutoSpecialize; under FullSpecialize the AD backend is returned unchanged so DifferentiationInterface generates a fresh runtime tag from the function type, which orders correctly relative to inner-solve tags.
- Generalizes `_wrapped_forwarddiff_ad()` to take the problem eltype as a type parameter, so the stamped tag is `Tag{NonlinearSolveTag, eltype(prob.u0)}` rather than always `Tag{NonlinearSolveTag, Float64}`. Applies to both the AutoForwardDiff and AutoPolyesterForwardDiff paths.
- **Precompile workload kept as-is**: `const dualT = Dual{Tag{NonlinearSolveTag, Float64}, Float64, 1}` and its `@compile_workload` usage are unchanged. The workload is only reached via the actually-wrapped AutoSpecialize path, and the remaining nested-AD failure mode for that path is addressed on the inner-solver side by widening `UJacobianWrapper.p` to the nested Dual type in SciML/OrdinaryDiffEq.jl#3389 — so the precompile-baked tagcount literal is no longer load-bearing for correctness and the TTFX win from exercising the real `NonlinearSolveTag` Dual operations at precompile time is preserved.
- Adds a regression test (`lib/NonlinearSolveBase/test/runtests.jl`) that builds a FullSpecialize `NonlinearLeastSquaresProblem` with `Vector{Float64}` state and asserts `standardize_forwarddiff_tag(ad, prob) === ad` for both `AutoForwardDiff` and `AutoPolyesterForwardDiff`. Confirmed to fail on unpatched master (1/18 fail) and pass post-fix (18/18 pass).
- Bumps `NonlinearSolveBase` patch version 2.24.0 → 2.25.0.

## Root cause chain (abbreviated)

ForwardDiff's tag precedence `≺` is `tagcount(T1) < tagcount(T2)`, where `tagcount` is a `@generated` function that bakes a literal value at first compile. Each `(F, V)` pair that's exercised during a package's `@compile_workload` gets a tagcount literal frozen into that package's precompile cache. At runtime, tags that were not precompile-baked have their tagcount generated lazily against ForwardDiff's session-local `TAGCOUNT` atomic, which starts at 0.

The broken situation: the user's FullSpecialize MWE went through `standardize_forwarddiff_tag`, which substituted `Tag{NonlinearSolveTag, Float64}` (precompile-baked by NSB's workload) onto the outer AD backend even though the function was not wrapped. Inside the residual, `OrdinaryDiffEqDifferentiation.prepare_ADType` creates `Tag(OrdinaryDiffEqTag(), eltype(u0))` where `eltype(u0) = Dual{Tag{NonlinearSolveTag, Float64}, Float64, 2}` — a compound type neither NSB nor OrdinaryDiffEq precompiles individually, so it's generated lazily at runtime with tagcount `0`. The cross-tag `p[i] * u[i]` dispatch inside `ode` then checks `Ty ≺ Tx` = `inner ≺ outer` = `(0 < N)` = true, takes the wrong branch of `*`, and produces a triple-nested Dual that crashes `setindex!(du, ...)` with `Float64(::nested_dual)` / MethodError.

Removing the `isa Vector{Float64}` gate eliminates this substitution for FullSpecialize problems entirely — DI generates its own runtime tag from the user function, whose tagcount is lower than the inner compound tag (generated later), and `≺` orders them correctly.

For the remaining AutoSpecialize-wrapped case, where NSB genuinely needs the stamped tag to match the precompiled `FunctionWrappersWrapper` signatures, the precompile-baked tagcount literal remains — but the `_widen_uf_p_for_jac` patch in SciML/OrdinaryDiffEq.jl#3389 pre-lifts `uf.p` to the inner nested-Dual type before the inner Jacobian runs, so the cross-tag multiplication never happens. The two patches together cover the full failure class.

## Notes

- The `NORECOMPILE_ARGUMENT_MESSAGE` / `NoRecompileArgumentError` in `autospecialize.jl:9-22` still mention the `Vector{Float64}` / `NullParameters` restriction, but they are unused internally (no call sites). Left alone to minimize scope.
- No behavior change for users who were already wrapping their function via `AutoSpecialize` — they still get the stamped tag, now parameterized by `eltype(prob.u0)` rather than hardcoded Float64.

## Test plan

- [x] `lib/NonlinearSolveBase/test/runtests.jl` regression test reproduces the bug on unpatched master (fails) and passes post-fix (18/18).
- [x] User's original MWE from SciML/OrdinaryDiffEq.jl#3381 passes under the patched NSB with unpatched `OrdinaryDiffEqDifferentiation`.
- [x] Aqua / ExplicitImports / Termination Conditions / BandedMatrix existing tests still pass.
- [ ] Full NonlinearSolve.jl CI test matrix.

### DIFF
--- a/lib/NonlinearSolveBase/Project.toml
+++ b/lib/NonlinearSolveBase/Project.toml
@@ -1,6 +1,6 @@
 name = "NonlinearSolveBase"
 uuid = "be0214bd-f91f-a760-ac4e-3421ce2b2da0"
-version = "2.24.0"
+version = "2.25.0"
 authors = ["Avik Pal <avikpal@mit.edu> and contributors"]
 
 [deps]
--- a/lib/NonlinearSolveBase/ext/NonlinearSolveBaseForwardDiffExt.jl
+++ b/lib/NonlinearSolveBase/ext/NonlinearSolveBaseForwardDiffExt.jl
@@ -30,36 +30,37 @@ dualgen(::Type{T}) where {T} = ForwardDiff.Dual{
     ForwardDiff.Tag{NonlinearSolveTag, T}, T, 1,
 }
 
-# Helper: build the canonical AutoForwardDiff for wrapped functions (chunksize=1 + tag).
-function _wrapped_forwarddiff_ad()
-    tag = ForwardDiff.Tag(NonlinearSolveTag(), Float64)
+# Helper: build the canonical AutoForwardDiff for wrapped functions
+# (chunksize=1 + NonlinearSolveTag). The tag's `V` parameter takes the
+# actual problem eltype rather than being hardcoded to `Float64`, so the
+# stamped AD backend properly reflects the user's problem type.
+function _wrapped_forwarddiff_ad(::Type{T}) where {T}
+    tag = ForwardDiff.Tag(NonlinearSolveTag(), T)
     return AutoForwardDiff{1, typeof(tag)}(tag)
 end
 
-# Stamp AutoForwardDiff with NonlinearSolveTag so duals match FunctionWrapper signatures.
-# When the function has been wrapped via AutoSpecialize, also force chunksize=1 to match
-# the precompiled N=1 dual type in the wrappers.
+# Stamp AutoForwardDiff with NonlinearSolveTag so duals match the wrapped
+# `FunctionWrappersWrapper` signatures. Only stamps when the user function was
+# actually wrapped via AutoSpecialize — otherwise leaves `ad` untouched so
+# DifferentiationInterface generates a fresh runtime tag from the function type.
+# Substituting the canonical tag in the non-wrapped path would otherwise drag in
+# a precompile-time `@generated tagcount` literal that can `≺`-reverse against
+# tags created later for nested ForwardDiff over an inner solve.
 function standardize_forwarddiff_tag(
         ad::AutoForwardDiff{CS, Nothing}, prob::AbstractNonlinearProblem
     ) where {CS}
-    prob.u0 isa Vector{Float64} || return ad
-    if is_fw_wrapped(prob.f.f)
-        return _wrapped_forwarddiff_ad()
-    end
-    tag = ForwardDiff.Tag(NonlinearSolveTag(), Float64)
-    return AutoForwardDiff{CS, typeof(tag)}(tag)
+    is_fw_wrapped(prob.f.f) || return ad
+    return _wrapped_forwarddiff_ad(eltype(prob.u0))
 end
 
-# AutoPolyesterForwardDiff doesn't support custom tags. When the function is wrapped,
-# replace it with AutoForwardDiff (chunksize=1, NonlinearSolveTag) so duals match wrappers.
+# AutoPolyesterForwardDiff doesn't support custom tags. When the function is
+# wrapped, replace it with AutoForwardDiff (chunksize=1, NonlinearSolveTag) so
+# duals match wrappers. Otherwise leave it alone.
 function standardize_forwarddiff_tag(
         ad::AutoPolyesterForwardDiff, prob::AbstractNonlinearProblem
     )
-    prob.u0 isa Vector{Float64} || return ad
-    if is_fw_wrapped(prob.f.f)
-        return _wrapped_forwarddiff_ad()
-    end
-    return ad
+    is_fw_wrapped(prob.f.f) || return ad
+    return _wrapped_forwarddiff_ad(eltype(prob.u0))
 end
 
 # IIP wrapfun: wraps f(du, u, p) with dual-aware type combinations.
--- a/lib/NonlinearSolveBase/ext/NonlinearSolveBaseForwardDiffExt.jl
+++ b/lib/NonlinearSolveBase/ext/NonlinearSolveBaseForwardDiffExt.jl
@@ -64,16 +64,16 @@ function standardize_forwarddiff_tag(
 end
 
 # IIP wrapfun: wraps f(du, u, p) with dual-aware type combinations.
-# Works for any `AbstractArray` state; the Dual-eltype array type `VdT` is
-# derived from `similar(u0, dT)` so the signatures follow the user's concrete
-# array kind (plain `Vector{Float64}` → `Vector{Dual}`, `Array{Float64, 3}` →
-# `Array{Dual, 3}`, etc.).
+# Works for any `AbstractArray` state; `ArrayInterface.promote_eltype` produces
+# the Dual-eltype array type at the type level (no allocation) so signatures
+# follow the user's concrete array kind (plain `Vector{Float64}` →
+# `Vector{Dual}`, `Array{Float64, 3}` → `Array{Dual, 3}`, etc.).
 @inline function wrapfun_iip(
         ff, inputs::Tuple{T1, T2, T3}
     ) where {T1 <: AbstractArray, T2 <: AbstractArray, T3}
     T = eltype(T1)
     dT = dualgen(T)
-    VdT = typeof(similar(inputs[1], dT))
+    VdT = ArrayInterface.promote_eltype(T1, dT)
     iip_arglists = (
         Tuple{T1, T2, T3},
         Tuple{VdT, VdT, T3},
--- a/lib/NonlinearSolveBase/ext/NonlinearSolveBaseForwardDiffExt.jl
+++ b/lib/NonlinearSolveBase/ext/NonlinearSolveBaseForwardDiffExt.jl
@@ -64,16 +64,20 @@ function standardize_forwarddiff_tag(
 end
 
 # IIP wrapfun: wraps f(du, u, p) with dual-aware type combinations.
-# Works for any `AbstractArray` state; `ArrayInterface.promote_eltype` produces
-# the Dual-eltype array type at the type level (no allocation) so signatures
-# follow the user's concrete array kind (plain `Vector{Float64}` →
-# `Vector{Dual}`, `Array{Float64, 3}` → `Array{Dual, 3}`, etc.).
+# Works for any `AbstractArray` state; the Dual-eltype array type `VdT` is
+# derived via `typeof(similar(u0, dT))` so signatures follow the user's
+# concrete array kind (plain `Vector{Float64}` → `Vector{Dual}`,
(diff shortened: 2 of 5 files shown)


ChrisRackauckas-Claude:
Pushed `8b8a8de` to address the CI failures from the prior commit.

**Root cause of the regression:** the prior commit removed the `u0 isa Vector{Float64}` gate from `maybe_wrap_nonlinear_f` in addition to the equivalent gate in `standardize_forwarddiff_tag`. Removing it from the latter is what actually fixes #3381 (gate now keys on `is_fw_wrapped(prob.f.f)`), but removing it from `maybe_wrap_nonlinear_f` was unnecessary and broke two paths:

1. `forward_ad_tests.jl "NLLS Hessian SciML/NonlinearSolve.jl#445"` — `get_concrete_problem` calls `promote_u0(u0, p, nothing)` before `maybe_wrap_f`. With outer `ForwardDiff.gradient` over an NLLS, `p::Vector{Dual{outer,...,2}}` promotes `u0::Vector{Float64}` to `Vector{Dual{outer,...,2}}`. Without the gate, the FWW gets built keyed off Dual eltype. `nonlinearsolve_forwarddiff_solve` then `remake`'s the prob to Float64 u0/p but reuses that Dual-keyed wrapper, so the inner Jacobian call `f(du::Vector{Float64}, u::Vector{Float64}, p::Vector{Float64})` crashes inside `FunctionWrappersWrappers._call` with "No matching function wrapper was found!".

2. `sparsity_tests.jl "Brusselator 2D"` — `u0::Array{Float64, 3}`. The default (non-ForwardDiff) `wrapfun_iip` froze the wrapper's only signature to that exact 3D array type, and the inner ForwardDiff Jacobian call with `Array{Dual,3}` args couldn't dispatch.

**Fix:** restore the `u0 isa Vector{Float64}` gate in `maybe_wrap_nonlinear_f`. The `standardize_forwarddiff_tag` change (the actual #3381 fix, gating on `is_fw_wrapped`) is preserved.

For a FullSpecialize Vector{Float64} prob (the #3381 MWE):
- `maybe_wrap_nonlinear_f`: gate passes (u0 is Vector{Float64}), then the `specialization === AutoSpecialize` check fails → returns raw `prob.f.f`. Not wrapped. ✓
- `standardize_forwarddiff_tag`: `is_fw_wrapped(prob.f.f)` is false → returns `ad` unchanged. DI generates a runtime tag from the function type. ✓ (#3381 fixed.)

Added a `maybe_wrap_nonlinear_f` regression test in `lib/NonlinearSolveBase/test/runtests.jl` covering the three cases (Vector{Float64} → wraps, Vector{Dual} → does not wrap, Array{Float64,3} → does not wrap). NonlinearSolveBase test suite passes locally (22/22).

ChrisRackauckas-Claude:
Pushed `c196f61` to drop the `Vector{Float64}` gate per review. The regressions are instead addressed at the source:

1. **Broaden `wrapfun_iip` (ForwardDiff ext) from `AbstractVector` to `AbstractArray`**, and derive the Dual-eltype array type as `typeof(similar(u0, dT))` instead of hardcoding `Vector{dT}`. This produces correctly-shaped Dual signatures for `Array{Float64, 3}` (Brusselator 2D), any `AbstractVector` eltype (including `Vector{Dual}` from a nested-AD promotion), and any other array kind where `similar` returns a sensible concrete type.

2. **Unwrap the FWW in `nonlinearsolve_forwarddiff_solve` and `__init(::DualAbstractNonlinearProblem)`** when building the value-typed inner problem via `remake(prob; u0=Utils.value, p=Utils.value)`. `get_concrete_problem` runs `promote_u0` before `maybe_wrap_f`, so a nested-AD NLLS outer prob is wrapped under a Dual u0 eltype; `remake` reuses that wrapper, and without unwrapping the inner Float64 solve would miss dispatch. Letting the inner `maybe_wrap_f` rebuild at the value eltype keeps the two worlds consistent.

3. **`maybe_wrap_nonlinear_f` now gates only on `AbstractArray` state, `AutoSpecialize`, and `isinplace`** — no Float64.

Updated the added regression test to assert the new positive behavior: `Vector{Float64}`, `Vector{Dual}`, and `Array{Float64, 3}` all wrap, and the wrapper works under the respective AD paths.

NonlinearSolveBase test suite passes locally (21/21).

ChrisRackauckas-Claude:
Reproduced the HC "no real solutions" failures locally and they are **not** caused by this PR:

- Checked out `8b8a8de` (which `CI (NonlinearSolveHomotopyContinuation)` showed as green on 2026-04-13) into the working tree and ran `lib/NonlinearSolveHomotopyContinuation` tests. Got **the same 6 failures** at `allroots.jl:156` (`length(_sol) == 1` evaluating `2 == 1`).
- HC's `solve` dispatch (`CommonSolve.solve(::NonlinearProblem, ::HomotopyContinuationJL{true})` in `lib/NonlinearSolveHomotopyContinuation/src/solve.jl:59`) does not call `get_concrete_problem`, `maybe_wrap_nonlinear_f`, `standardize_forwarddiff_tag`, or `wrapfun_iip` — the functions this PR touches are structurally unreachable from this codepath.
- `HomotopyContinuation.jl` shipped v2.18.2 on 2026-04-12 with a "patch decompose for dimension 0" change. The failing test asserts that at `p = zeros(2)` the solver returns a length-1 sentinel ensemble (no real roots), but under the new HC release `HC.results(orig_sol; only_real = true)` returns 2 near-real solutions that survive the imaginary-part threshold, so the HC wrapper enters the valid-solutions branch and returns a length-2 ensemble instead. This is a numerical boundary issue in HC v2.18.2 on the "no real roots" case, unrelated to NonlinearSolveBase.

`NonlinearSolveBase` tests still pass locally (21/21) on `bffb6cd`. Leaving this PR's implementation as-is; the HC regression is a separate ecosystem issue and should be filed against HomotopyContinuation.jl (or the test's tolerance relaxed in `NonlinearSolveHomotopyContinuation`).

Remaining CI failures that persist:
- 6× "no real solutions" in `lib/NonlinearSolveHomotopyContinuation/test/allroots.jl:156` — HC v2.18.2 ecosystem issue (above).
- `Documentation` — pre-existing `modAB`/`ModAB` cross-reference in `docs/src/solvers/bracketing_solvers.md:11`, failing on master for days.
- 3× `ModelingToolkit.jl/*` — pre-existing `UndefVarError: handle_callback_modifiers! not defined in StochasticDiffEq`, also failing on master.

ChrisRackauckas-Claude:
Pushed `d3fbe2c` to fix the "nopre" `test (..., ubuntu-latest) / test` / `test (..., macos-latest) / test` failures. They were real, not flaky — separate from the HC v2.18.2 issue.

**Root cause:** `forward_ad_tests.jl` Scalar AD / Jacobian testsets iterate over every `(u0, p)` combination inside a `ForwardDiff.derivative(solve, p)` closure, for every concrete algorithm (`NewtonRaphson`, `TrustRegion`, `Broyden`, …). The dual path goes:

1. `init(prob, alg)` → `get_concrete_problem` → `promote_u0` upgrades `u0` to a `Dual`-eltype array when `p` carries an outer-AD Dual.
2. `maybe_wrap_f` then wrapped the promoted prob (AbstractArray gate from `c196f61`), producing a `FunctionWrappersWrapper` whose signatures were keyed off the outer AD tag.
3. Dispatch landed on `SciMLBase.__init(::DualAbstractNonlinearProblem, alg::<concrete>)`. The forward-diff ext's `__init` override I added was scoped to `GENERAL_SOLVER_TYPES = [Nothing, PolyAlg]`, so for concrete algs there was no unwrap step.
4. The Jacobian cache then called `f(du::Vector{Float64}, u::Vector{Float64}, p::Float64)` (value-typed inner solve) on the outer-tagged FWW — "No matching function wrapper".

**Fix:** skip `maybe_wrap_nonlinear_f` entirely when `eltype(u0)` is a `Dual` (via `SciMLBase.isdualtype`). Outer-AD already introduced the duals, the forward-diff ext's dual dispatch handles the rest via `Utils.value` / `nodual_value`, and the raw user function specializes fine at each eltype combination without a FWW in between. No Float64 requirement — the gate is "has the outer AD already promoted this?", which is the load-bearing invariant.

Remaining 6 "no real solutions" HC failures are the HomotopyContinuation.jl v2.18.2 `decompose-patch` numerical issue documented in the previous comment. Documentation + MTK failures are pre-existing master failures.

ChrisRackauckas-Claude:
Confirmed that the 6 `CI (NonlinearSolveHomotopyContinuation)` "no real solutions" failures are **not caused by this PR**. #915 is a diagnostic branch off plain master with no changes except a single no-op comment in `lib/NonlinearSolveHomotopyContinuation/src/interface_types.jl` (just to trigger the path-filtered workflow). On that branch the HC CI reports the exact same pattern on all 3 Ubuntu jobs (1.10, 1.11, 1):

```
no real solutions: Test Failed at lib/NonlinearSolveHomotopyContinuation/test/allroots.jl:156
 Evaluated: 2 == 1
```

Same system, same assertion, no NonlinearSolveBase changes — so the flake lives on master. #914 has the actual fix (seed the HC path tracker + relax the assertion to the wrapper's true invariant `all(!SciMLBase.successful_retcode, _sol.u)`). #915 will be closed without merging once the macOS jobs finish.

PR #952, opened 2026-06-04
ci: revert unsafe sublibrary CI consolidation (restore bespoke per-sublibrary CI; keep faithful main conversions)

ChrisRackauckas-Claude:
## Summary

This PR corrects a **CI coverage regression** introduced by #951 (CI centralization). An audit of #951 found that consolidating the 10 per-sublibrary `CI_*.yml` workflows into a single `SublibraryCI.yml` (a caller of `sublibrary-tests.yml@v1`) **silently dropped test coverage**, because `sublibrary-tests.yml@v1` defaults to `GROUP=Core`/`GROUP=QA` and does not run the real group/tag axes these sublibraries dispatch on.

This PR **reverts only the unsafe part** of #951 and **keeps the faithful parts**.

## What #951 dropped (the regression)

- **NonlinearSolveSciPy**: the entire `:wrappers` suite ran zero items.
- **SimpleNonlinearSolve**: the `:adjoint` and `:alloc_check` groups were no longer run.
- **macOS** (`macos-latest`) legs for **all 10** sublibraries.
- Per-suite **Julia version matrices** (e.g. `1.10`/`1.11`/`1`).
- The main-package `CI_NonlinearSolve.yml` was converted to `tests.yml@v1`, which **cannot express** its `local_dependencies` (`Pkg.develop` of the `lib/*` path deps) or its `GROUP` `test_args`.

## What this PR does

**Reverts (restores bespoke):**
- Restores all 10 per-sublibrary `CI_*.yml` verbatim from before #951, preserving their `adjoint`/`alloc_check`/`wrappers` groups, their `ubuntu-latest`+`macos-latest` OS matrices, and their per-suite version matrices.
- Restores `CI_NonlinearSolve.yml` to its bespoke form (preserving `local_dependencies` and `GROUP` `test_args`).
- Removes the added `SublibraryCI.yml`.

**Keeps (faithful main-package conversions from #951 — these were NOT flagged by the audit):**
- `Documentation.yml` → `documentation.yml@v1`
- `Downstream.yml` → `downstream.yml@v1` (same 7-package matrix)
- `FormatCheck.yml` → `runic.yml@v1` (Runic preserved)
- `SpellCheck.yml` → `spellcheck.yml@v1`
- `Downgrade.yml` was already a reusable caller (untouched).

**Untouched:** `DowngradeSublibraries.yml` (from #950), `GPU.yml`, `TagBot.yml`, `CompatHelper.yml`.

## Why the sublibrary CI was intentionally LEFT BESPOKE

`sublibrary-tests.yml@v1` auto-discovers `lib/*` but sends only the default `GROUP=Core`/`QA` values. The sublibraries here dispatch their `test/runtests.jl` on **different** group/tag names (`:wrappers`, `:adjoint`, `:alloc_check`, plus per-suite OS and version matrices). Consolidating onto the reusable defaults therefore runs **zero items** for those axes — a silent coverage loss. Per the centralization spec's "MONOREPO SUBLIBRARY CONSOLIDATION — DANGER" rule, the correct action is to leave the per-sublibrary `CI_*.yml` bespoke rather than convert-and-lose.

## Proper follow-up (separate, maintainer-reviewed step)

Adopting `sublibrary-tests.yml@v1` safely requires, per sublibrary:
1. Add `lib/<name>/test/test_groups.toml` encoding the real groups (e.g. `core`, `adjoint`, `alloc_check`, `wrappers`) **plus** the OS and Julia-version matrices.
2. Align each `lib/<name>/test/runtests.jl` `GROUP` parsing to honor the `GROUP` value the reusable workflow sends.

Only once both are in place (and verified to run the same item count) should the bespoke `CI_*.yml` be replaced by `SublibraryCI.yml`. That work is out of scope here and should be done as its own reviewed PR.

## Validation

- All workflow YAML files parse (`python3 yaml.safe_load`).
- Every remaining reusable caller references a real `SciML/.github/.github/workflows/*.yml@v1` workflow, and every input passed is declared by that reusable workflow.
- The net effect on `master`: the 10 sublibrary `CI_*.yml` + bespoke `CI_NonlinearSolve.yml` are restored, `SublibraryCI.yml` is removed, and the 4 faithful main-package conversions are kept.

---


ChrisRackauckas-Claude:
Superseded by https://github.com/SciML/NonlinearSolve.jl/pull/953: keeping the centralized sublibrary-tests@v1 form and fixing coverage via test_groups.toml + runtests GROUP parsing, instead of reverting to bespoke CI_*.yml.

PR #955, opened 2026-06-05
Fix bounds transform for problems with initialization data and quasi-Newton algorithms

ChrisRackauckas-Claude:
> [!NOTE]

Two bugs in the box-constraint bounds transform (`lib/NonlinearSolveBase/src/bounds_transform.jl`), both surfaced while wiring ModelingToolkit's `bounds` metadata through to nonlinear solvers (SciML/ModelingToolkit.jl#4308).

## 1. `alg.autodiff` `FieldError` on quasi-Newton algorithms

`transform_bounded_problem` unconditionally read `alg.autodiff` to choose the cache type. `QuasiNewtonAlgorithm` (Broyden/Klement, reached as part of the default `FastShortcutNonlinearPolyalg`) has no `autodiff` field, so solving a bounded problem with the default polyalgorithm threw:

```
FieldError: type QuasiNewtonAlgorithm has no field `autodiff`
```

Fixed by guarding the access with `hasproperty`.

## 2. Initialization data corrupts the transformed solution

The wrapped problem function carried over the original `initialization_data`, which is defined in the **bounded** coordinates. The transform reparametrizes the problem into an unbounded variable `t`; running the bounded-space initialization on the transformed problem interprets the iterate `t` as a bounded state and corrupts the result.

Concretely, every ModelingToolkit `NonlinearProblem` attaches initialization data, so a bounded MTK problem converged to `from_unbounded(root)` instead of `root`:

```julia
# x^2 - 4x + 3, root in [0,2] is x=1
solve(prob).u   # => [1.4621...]  ==  2*logistic(1.0)  ==  from_unbounded(1.0)   (wrong)
```

Initialization is now run **once in the original bounded space** (`run_bounded_initialization`) before the transform, and stripped from the transformed function (`build_initializeprob = Val{false}`) so it isn't re-run in `t`-space. Raw problems without initialization data are unaffected.

## Tests

- Added a `:bounds` testitem covering the default polyalgorithm and quasi-Newton algorithms (`Broyden`, `Klement`) on a bounded `NonlinearProblem` (exercises fix #1 and confirms the correct root is found).
- The initialization interaction (fix #2) is exercised end-to-end by the new bounds tests in SciML/ModelingToolkit.jl#4308.
- Existing `test/bounds_tests.jl` scenarios verified locally against this branch (no regression).

## Dependencies

Builds on SciML/SciMLBase.jl#1372 (so `remake` preserves `lb`/`ub`, which the transform's solution-reconstruction relies on).

### DIFF
--- a/lib/NonlinearSolveBase/Project.toml
+++ b/lib/NonlinearSolveBase/Project.toml
@@ -1,6 +1,6 @@
 name = "NonlinearSolveBase"
 uuid = "be0214bd-f91f-a760-ac4e-3421ce2b2da0"
-version = "2.30.2"
+version = "2.30.3"
 authors = ["Avik Pal <avikpal@mit.edu> and contributors"]
 
 [deps]
--- a/lib/NonlinearSolveBase/src/bounds_transform.jl
+++ b/lib/NonlinearSolveBase/src/bounds_transform.jl
@@ -117,6 +117,13 @@ end
 # unbounded parameter `t` using the logistic function to map all values of `t`
 # into the interval (lb, ub).
 function transform_bounded_problem(prob, alg)
+    # `initialization_data` (e.g. from ModelingToolkit) is defined in the original,
+    # bounded coordinates. It must run *before* the transform so the resulting `u0`/`p`
+    # are consistent; running it on the transformed problem would interpret the unbounded
+    # iterate `t` as a bounded state and corrupt the solution. We run it here and strip it
+    # from the transformed function below so it isn't re-run in `t`-space.
+    prob = run_bounded_initialization(prob, alg)
+
     lb, ub = _normalize_bounds(prob.lb, prob.ub, prob.u0)
 
     # Clamp u0 into the interior of the bounds so that _to_unbounded doesn't hit log(0)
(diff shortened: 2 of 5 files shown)


ChrisRackauckas-Claude:
Pushed a follow-up addressing the `Tests (core, lts)` failure.

**Cause:** the umbrella `test/bounds_tests.jl` integration test exercised the fix through a full solve, but on Julia 1.10 (`lts`) the `[sources]` path redirect to the in-repo `lib/NonlinearSolveBase` is **ignored**, so the umbrella resolved the *registered* (unpatched) NonlinearSolveBase and hit the very `alg.autodiff` `FieldError` this PR fixes.

**Fix:** moved the regression coverage into NonlinearSolveBase's own test suite (`lib/NonlinearSolveBase/test/bounds_transform.jl`), which always runs against the in-repo code (Sublibrary CI, incl. lts) — it calls `transform_bounded_problem` with a dummy algorithm lacking an `autodiff` field. The umbrella full-solve test is now gated on `pkgversion(NonlinearSolveBase) >= v"2.30.3"` so it only runs where the fix is actually loaded.

**Unrelated pre-existing failure:** `Tests (nopre, lts)` fails on the "Out-of-place/Inplace Matrix Resizing" testitems (`test/core_tests.jl`), which assert `vec(solve(matrix_prob).u) == solve(vec_prob).u` with exact `==`. The two solves differ by ~1 ULP on Julia 1.10 (e.g. `1.414213562373095` vs `…51`). This is a fragile exact-equality comparison across the matrix/vector reshape path, independent of this PR (it touches neither `core_tests.jl` nor the solver internals, and on lts this PR's code isn't even loaded).

PR #962, opened 2026-06-08
Add `HomotopySweep`: natural-parameter continuation solver for `HomotopyProblem`

BatchZero:
## Summary

Adds `HomotopySweep`, a natural-parameter continuation solver for `SciMLBase.HomotopyProblem`. Given a homotopy family `0 = H(u, p, λ)`, it sweeps `λ` across the problem's `λspan` (default `0 → 1`), warm-starting an inner nonlinear solve at each step from the previous step's solution, with step-size bisection on a non-converging step. This robustly solves systems whose target form (`λ=1`) is hard to solve cold but whose `λ=0` form is easy.

It is MTK-independent — any user with a hand-built `HomotopyProblem` can `solve(prob, HomotopySweep())`. It is unrelated to the polynomial `HomotopyContinuationJL` (which tracks complex polynomial paths via HomotopyContinuation.jl); `HomotopySweep` is a real-arithmetic embedding-homotopy continuation for general nonlinear systems.

## API

```julia
HomotopySweep(; inner = nothing, nsteps = 10, adaptive = true, min_dλ = 1.0e-3)
```

- `inner` — inner nonlinear algorithm used at each λ-step; `nothing` selects the default polyalgorithm (not a hardcoded Newton).
- `nsteps` — number of λ steps across `λspan`.
- `adaptive` / `min_dλ` — on a non-converging inner solve, halve the λ increment and retry, down to `min_dλ`; otherwise surface the inner solver's failure retcode.

A `HomotopyProblem` solved with no algorithm (`solve(prob, nothing)`) defaults to `HomotopySweep()`.

The algorithm lives in `NonlinearSolveBase` (beside the sibling meta-algorithm `NonlinearSolvePolyAlgorithm`) and is re-exported by NonlinearSolve.

## Tests

`test/Core/homotopy_sweep_tests__item*.jl` (Core group): construction/defaults; monotone-sweep convergence; an **out-of-basin rescue** acceptance test (a guess from which a cold Newton diverges — `atan(x-3)` saturation — rescued to the actual root by continuation); failure retcode when the family has no real root past a fold; composable inner solver (default polyalg / NewtonRaphson / SimpleNewtonRaphson); integer-eltype `p`; and the `nothing`-algorithm default route.

## Depends on

SciML/SciMLBase.jl#1375 — adds `HomotopyProblem`, which this solver dispatches on. This PR should merge **after** that one releases; the `[compat] SciMLBase` lower bound will be bumped to the release carrying `HomotopyProblem`.

## Part of a series

Second of three coordinated PRs for the Modelica `homotopy()` operator:
SciMLBase `HomotopyProblem` (SciML/SciMLBase.jl#1375) → **(this PR) NonlinearSolve `HomotopySweep`** → ModelingToolkit lowering (to follow).


lib/NonlinearSolveBase/src/homotopy_sweep.jl
```diff
@@ -0,0 +1,89 @@
...
+    λ = float(λ0)
+    local last_sol
+
+    while true
+        next_λ = abs(λ1 - λ) <= abs(dλ) ? λ1 : λ + dλ
+        inner_prob = NonlinearProblem(prob.f, u, set_λ(prob.p, next_λ))
+        last_sol = solve(inner_prob, alg.inner, args...; kwargs...)
+
+        if SciMLBase.successful_retcode(last_sol)
+            u = last_sol.u
```
ChrisRackauckas: This is a pointer assignment, so won't it just alias and start overwriting?

lib/NonlinearSolveBase/src/homotopy_sweep.jl
```diff
@@ -0,0 +1,89 @@
...
+            λ = next_λ
+            λ == λ1 && break
+        elseif alg.adaptive && abs(dλ) / 2 >= alg.min_dλ
+            # conservative: step size is not restored after a bisection
+            dλ = dλ / 2          # bisect; retry from the same λ (do not advance)
+        else
+            # on failure: u is the last CONVERGED iterate (λ<λ1); resid is from the failed step (advisory)
```
ChrisRackauckas: This is not true if pointer assignment.

lib/NonlinearSolveBase/src/homotopy_sweep.jl
```diff
@@ -0,0 +1,89 @@
...
+"""
+@concrete struct HomotopySweep <: AbstractNonlinearSolveAlgorithm
+    inner
+    nsteps::Int
+    adaptive::Bool
+    min_dλ
+end
+
+function HomotopySweep(; inner = nothing, nsteps = 10, adaptive = true, min_dλ = 1.0e-3)
+    nsteps >= 1 || throw(ArgumentError("HomotopySweep `nsteps` must be ≥ 1, got $nsteps"))
```
ChrisRackauckas: nsteps magic number?

lib/NonlinearSolveBase/src/homotopy_sweep.jl
```diff
@@ -0,0 +1,89 @@
...
+        inner_prob = NonlinearProblem(prob.f, u, set_λ(prob.p, next_λ))
+        last_sol = solve(inner_prob, alg.inner, args...; kwargs...)
+
+        if SciMLBase.successful_retcode(last_sol)
+            u = last_sol.u
+            λ = next_λ
+            λ == λ1 && break
+        elseif alg.adaptive && abs(dλ) / 2 >= alg.min_dλ
+            # conservative: step size is not restored after a bisection
+            dλ = dλ / 2          # bisect; retry from the same λ (do not advance)
```
ChrisRackauckas: This is a bit simple, but can be improved in a future PR.

lib/NonlinearSolveBase/src/homotopy_sweep.jl
ChrisRackauckas: I'm not sure about having a magic number here. `initial_step_factor = 0.1`, `nsteps= nothing`, `adaptive = true`. If `adaptive = false`, `nsteps` must be set. `initial_step_factor` is a default initial step, but not necessarily a default number of steps, which IMO would be dangerous.

ChrisRackauckas: `min_dλ = 1.0e-3` this is not floating point dependent. I would make it `nothing`, and in the algorithm instead use a value of `sqrt(eps(lambda))`

ChrisRackauckas: That way if you change precision, the minimum changes with it. 1e-3 is more for Float32 than Float64.

ChrisRackauckas:
Close but there's still a bit to do for a first merge. But main thing, the homotopy parameter should definitely be separated from the full parameter vector.

BatchZero:
λ is now fully separated from the parameter vector across both PRs; SciML/SciMLBase.jl#1375 is updated with the `f(u, p, λ)` residual convention. The sweep no longer touches the user's `p`: each step wraps the residual in a `FixLambda` callable fixing λ as the trailing argument, so the inner solver sees a plain `(u, p)` / `(du, u, p)` problem of the same concrete function type every step and compiles once.

A few additional hardening changes landed alongside, so the diff holds no surprises: problem-level `prob.kwargs` are now forwarded to inner solves (per the documented SciMLBase contract), zero-arg `solve(prob)` routes to `HomotopySweep()` (previously a MethodError), a stall guard returns `ReturnCode.Stalled` when dλ underflows `eps(λ)` (details in the step-size thread), and failure returns carry `original = last_sol` for diagnostics, with the failure semantics documented in the docstring.

lib/NonlinearSolveBase/src/homotopy_sweep.jl
BatchZero: Correct — the comment's claim was false under pointer assignment, and the sharper version of the problem is on the copy-in side: with `alias = NonlinearAliasSpecifier(alias_u0 = true)` forwarded to the inner solves, a failed attempt iterates directly in its u0 buffer and leaves diverged garbage there, corrupting the warm start for the bisected retry. Fixed in both directions: each inner problem now receives `copy(u)`, and on success we take `u = copy(last_sol.u)` (defensive, in case an inner solver returns a view into persistent workspace). A regression test runs an aliased sweep against a non-aliased reference and requires exactly identical `u` (the step sequences are deterministic).

lib/NonlinearSolveBase/src/homotopy_sweep.jl
BatchZero: Addressed together with the pointer-assignment thread above — with copy-in (`copy(u)` per attempt) and copy-out (`copy(last_sol.u)` on success) in place, the "last converged iterate" claim is now actually guaranteed, and a regression test covers it.

lib/NonlinearSolveBase/src/homotopy_sweep.jl
BatchZero: Adopted as prescribed: `nsteps = nothing, adaptive = true, initial_step_factor = 0.1, min_dλ = nothing`; `adaptive = false` without an explicit `nsteps` throws an ArgumentError, and `min_dλ = nothing` resolves to `sqrt(eps(typeof(λ)))` at solve time so the floor scales with precision (`min_dλ <= 0` also throws — it would bisect forever). One judgment call to confirm: when `nsteps` is given together with `adaptive = true`, it sets the initial step to `span/nsteps` and bisection stays active; `initial_step_factor` only applies when `nsteps === nothing`. A second one: the eps floor is absolute, not scaled by the span width — on large-magnitude spans (e.g. `λspan = (1e9, 2e9)`) bisection can push dλ below `eps(λ)` before the floor bites, making `λ + dλ == λ`, so there is now a stall guard returning `ReturnCode.Stalled` instead of looping forever, with a test locking it. A span-scaled floor (`sqrt(eps(λT)) * abs(λ1 - λ0)`) would largely keep the sweep out of that regime — preference?

lib/NonlinearSolveBase/src/homotopy_sweep.jl
BatchZero: Agreed; keeping bisection-only here, with step regrowth and/or a predictor-corrector as the natural follow-up. One scoping question: with no regrowth, a single hard region that collapses dλ to the floor commits the sweep to up to `span/min_dλ` further inner solves. An outer `maxsteps` cap (or regrowth itself) would bound that — left out to keep this PR minimal.

ChrisRackauckas:
Bump the SciMLBase minimum to 3.19

test/homotopy_sweep_tests.jl
```diff
@@ -0,0 +1,188 @@
+@testitem "HomotopySweep construction + defaults" tags = [:core] begin
```
ChrisRackauckas: Update to the new test format.

ChrisRackauckas:
https://github.com/JuliaRegistries/General/pull/157666

BatchZero:
Done — `SciMLBase = "3.19"` in the root Project.toml and in lib/NonlinearSolveBase, master merged, and the tests are converted to the new layout (14 files under `test/Core/homotopy_sweep_tests__item*.jl`, registered in the Core group). All 14 pass locally against SciMLBase v3.19.0.

Three release-chain items came out of checking the CI semantics:
- NonlinearSolveBase's Aqua piracy check now treats `SciMLBase.HomotopyProblem` as own — the `solve(::HomotopyProblem, ...)` default-algorithm methods are all-foreign signatures, same pattern as the existing `NonlinearProblem` entry.
- Versions bumped: NonlinearSolveBase 2.31.0, NonlinearSolve 4.20.0, with the root NonlinearSolveBase floor raised to 2.31 since the root Core group now references `HomotopySweep`.
- Until a NonlinearSolveBase with `HomotopySweep` is registered, the lts grids (Pkg 1.10 ignores `[sources]`, so NonlinearSolveBase resolves from the registry) and the root Downgrade job will stay red; with the 2.31 floor they heal automatically once it's registered.

test/homotopy_sweep_tests.jl
BatchZero: Converted following the #961 layout — each `@testitem` is now a standalone `test/Core/homotopy_sweep_tests__itemN.jl` included via `@safetestset` from the Core group in `test/runtests.jl`.

lib/NonlinearSolveBase/test/qa/qa.jl
```diff
@@ -1,9 +1,12 @@
...
+using NonlinearSolveBase: AbstractNonlinearProblem, NonlinearProblem, SciMLBase
 
 Aqua.test_all(
     NonlinearSolveBase; piracies = false, ambiguities = false, stale_deps = false
 )
 Aqua.test_stale_deps(NonlinearSolveBase; ignore = [:TimerOutputs])
-Aqua.test_piracies(NonlinearSolveBase, treat_as_own = [AbstractNonlinearProblem, NonlinearProblem])
+Aqua.test_piracies(
+    NonlinearSolveBase,
+    treat_as_own = [AbstractNonlinearProblem, NonlinearProblem, SciMLBase.HomotopyProblem]
```
ChrisRackauckas, line 10: Should not pirate.

ChrisRackauckas: Oh for the default solver, ehhhhhh

ChrisRackauckas (approved):

ChrisRackauckas:
There are still a lot of things I want to do to improve the solver here, but this is a baseline that I think is good to start with.

lib/NonlinearSolveBase/src/homotopy_sweep.jl
```diff
@@ -0,0 +1,145 @@
...
+    λT = typeof(λ)
+    λend = λT(λ1)
+    span = λend - λ
+    dλ = alg.nsteps === nothing ? λT(alg.initial_step_factor) * span : span / alg.nsteps
+    min_dλ = alg.min_dλ === nothing ? sqrt(eps(λT)) : λT(alg.min_dλ)
+    u = copy(prob.u0)
+    local last_sol
+
+    while true
+        next_λ = abs(λend - λ) <= abs(dλ) ? λend : λ + dλ
```
ChrisRackauckas, line 95: This isn't evaluated at 0, so the initial is never found 😅

PR #970, opened 2026-06-13
Add missing InteractiveUtils test dep to NonlinearSolveHomotopyContinuation

ChrisRackauckas-Claude:
> [!NOTE]

`lib/NonlinearSolveHomotopyContinuation/test/runtests.jl` does `using InteractiveUtils` (like every sublibrary's test runner since #961), but this sublibrary's `Project.toml` never declared `InteractiveUtils` in `[extras]`/`[targets]`, so its `Pkg.test` fails immediately:

```
ERROR: LoadError: ArgumentError: Package InteractiveUtils not found in current path.
```

This is a pre-existing master failure: the `sublibraries / lib/NonlinearSolveHomotopyContinuation (julia 1)` Core and QA jobs fail with this exact error on the latest master Sublibrary CI run (633d93a00) — easy to miss because that run's overall conclusion shows "cancelled". The same two jobs also fail on every PR, e.g. #967.

Fix mirrors the declaration in the other nine sublibraries (`[compat] InteractiveUtils = "<0.0.1, 1"`, `[extras]` UUID, `[targets]` entry). All other sublibraries already declare it; this was the only one missed.

**Verification:** ran the full sublibrary test suite locally via the root dispatcher (`GROUP=NonlinearSolveHomotopyContinuation`, Julia 1.11) — previously it failed at the `using`; with this change all testsets pass ("Testing NonlinearSolveHomotopyContinuation tests passed", exit 0).

### DIFF
--- a/lib/NonlinearSolveHomotopyContinuation/Project.toml
+++ b/lib/NonlinearSolveHomotopyContinuation/Project.toml
@@ -30,6 +30,7 @@ DifferentiationInterface = "0.7.3"
 DocStringExtensions = "0.9.3"
 Enzyme = "0.13.90"
 HomotopyContinuation = "2.12.0"
+InteractiveUtils = "<0.0.1, 1"
 LinearAlgebra = "1.10"
 NaNMath = "1.1"
 NonlinearSolve = "4.10"
@@ -44,6 +45,7 @@ SafeTestsets = "0.1"
 SciMLTesting = "1"
 [extras]
 Enzyme = "7da242da-08ed-463a-9acd-ee780be4f1d9"
+InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
 NaNMath = "77ba4419-2d1f-58cd-9bb1-8ffee604a2e3"
 NonlinearSolve = "8913a72c-1f9b-4ce2-8d82-65094dcecaec"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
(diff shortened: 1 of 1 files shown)


ChrisRackauckas-Claude:
CI result: the previously-failing jobs are fixed — `sublibraries / lib/NonlinearSolveHomotopyContinuation` **Core passes on julia 1, lts, and pre**, and **QA passes on julia 1** (all failed before with the `InteractiveUtils not found` error).

Two remaining reds on this PR, neither caused by it:
- `[QA] (julia lts)`: a pre-existing dispatcher artifact, newly *exposed* because this sublibrary's full matrix only runs when it is touched — on Julia < 1.11 the dispatcher `Pkg.develop`s the `[sources]` closure into the sublibrary's own Project.toml before `Pkg.test`, and Aqua then flags those harness-injected deps as stale/uncompat'd. Filed with full analysis as #971.
- `downgrade-sublibraries`: the long-standing downgrade-resolution failure that hits all sublibraries on master too.

ChrisRackauckas-Claude:
This fix is still needed — the `InteractiveUtils not found` error is red on master across three lanes: Sublibrary CI HomotopyContinuation Core (Julia 1) + QA (Julia 1), and Downgrade Sublibraries test(lib/NonlinearSolveHomotopyContinuation). This PR's branch is behind master, though: its base predates the recent SciMLTesting migration / Enzyme-floor bumps (it still shows `Enzyme = "0.13"`, `Pkg` in the test target, and no `SciMLTesting`). It needs a rebase onto current master so the test target matches (current master uses `SciMLTesting` and `Enzyme = "0.13.90"`). Verified the InteractiveUtils fix on current master locally (Julia 1.10): `Testing NonlinearSolveHomotopyContinuation tests passed`. cc @ChrisRackauckas

Issue #971, opened 2026-06-13
Sublibrary QA fails Aqua stale_deps/deps_compat on Julia lts when [sources] reaches the umbrella root


ChrisRackauckas-Claude:
On Julia < 1.11 the sublibrary dispatcher in `test/runtests.jl` works around Pkg ignoring the `[sources]` table by `Pkg.activate`-ing the sublibrary and `Pkg.develop`-ing the transitive `[sources]` closure. `Pkg.develop` into the active project **adds those packages to the sublibrary's own `Project.toml` `[deps]`** in the working tree. The Aqua QA tests then read that mutated Project.toml and fail:

```
Stale dependencies: Test Failed
  Evaluated: isempty(Base.PkgId[NonlinearSolve, NonlinearSolveQuasiNewton, BracketingNonlinearSolve,
  SimpleNonlinearSolve, NonlinearSolveSpectralMethods, NonlinearSolveFirstOrder])
NonlinearSolveHomotopyContinuation does not declare a compat entry for the following deps:
  BracketingNonlinearSolve, NonlinearSolveFirstOrder, NonlinearSolveQuasiNewton,
  NonlinearSolveSpectralMethods, SciMLJacobianOperators, SimpleNonlinearSolve
```

The flagged set is exactly the `[sources]`-transitive develop set (umbrella root → all sublibraries, NonlinearSolveBase → SciMLJacobianOperators), confirming the mechanism. The same job on Julia 1 / pre passes because `[sources]` is honored and no mutation happens.

Affected: any sublibrary whose `[sources]` includes the umbrella root (`path = "../.."`), i.e. **NonlinearSolveHomotopyContinuation** and **SCCNonlinearSolve**, and only their `[QA] (julia lts)` jobs. Easy to miss because the sublibrary matrix is path-filtered: these jobs only run when the sublibrary itself is touched — first observed on #970 (see https://github.com/SciML/NonlinearSolve.jl/actions/runs/27450537066/job/81144716461).

Possible directions (each with trade-offs, so leaving to a maintainer decision rather than bundling into #970):
- Move the lts develop workaround from the dispatcher (which mutates the real Project.toml pre-test) into the sublibrary test runners, developing into the ephemeral test env instead — mirrors what the base-env groups do after #967; Aqua would then read the unmutated package Project.toml via `pkgdir`.
- Restore the sublibrary Project.toml after `Pkg.test` returns (doesn't help: Aqua runs *during* the test).
[...]

Issue #969, opened 2026-06-12
format-check failing on master since #961: sublibrary test/runtests.jl not Runic-formatted


ChrisRackauckas-Claude:
## Summary

The `format-check` (Runic) workflow has failed on every `master` run since commit 75178936d ("test: uniformize monorepo onto OrdinaryDiffEq canonical structure", #961, merged 2026-06-08). Six files are flagged:

- `lib/BracketingNonlinearSolve/test/runtests.jl`
- `lib/NonlinearSolveFirstOrder/test/runtests.jl`
- `lib/NonlinearSolveHomotopyContinuation/test/runtests.jl`
- `lib/NonlinearSolveQuasiNewton/test/runtests.jl`
- `lib/NonlinearSolveSpectralMethods/test/runtests.jl`
- `lib/SimpleNonlinearSolve/test/runtests.jl`

All six fail on the same construct, the `Pkg.develop([...])` call in the `activate_qa_env` helper that #961 added to each sublibrary's `test/runtests.jl`. Runic wants:

```diff
-        Pkg.develop([
-            Pkg.PackageSpec(path = joinpath(@__DIR__, "..")),
-            Pkg.PackageSpec(path = joinpath(@__DIR__, "..", "..", "NonlinearSolveBase")),
-        ])
+        Pkg.develop(
+            [
+                Pkg.PackageSpec(path = joinpath(@__DIR__, "..")),
+                Pkg.PackageSpec(path = joinpath(@__DIR__, "..", "..", "NonlinearSolveBase")),
+            ]
+        )
```

## Evidence

- Failing on 633d93a00: https://github.com/SciML/NonlinearSolve.jl/actions/runs/27316871345 (same 6 files)
- Failing on 45c88b901: https://github.com/SciML/NonlinearSolve.jl/actions/runs/27148997116 (same 6 files)
- First failing with this file set: the run on 75178936d itself, https://github.com/SciML/NonlinearSolve.jl/actions/runs/27134151806

Reproduced locally with Runic v1.7.0 on Julia 1.11: `Runic.main(["--check", file])` returns 1 for all six files (and 0 for e.g. `src/NonlinearSolve.jl`).

## Fix

Run `runic -i` on the six files (or on the repo) and commit. One mechanical formatting commit.

Issue #968, opened 2026-06-12
Docs build: new unresolvable @ref to SciMLBase.HomotopyProblem from HomotopySweep docstring (#962)


ChrisRackauckas-Claude:
## Summary

The `Documentation` workflow on `master` gained a **new** failure class starting with commit 633d93a00 (merge of #962, "Add `HomotopySweep`"): an unresolvable Documenter cross-reference.

```
┌ Error: Cannot resolve @ref for md"[`SciMLBase.HomotopyProblem`](@ref)" in docs/src/native/solvers.md.
...
ERROR: LoadError: `makedocs` encountered errors [:cross_references, :example_block] -- terminating build before rendering.
```

(From run https://github.com/SciML/NonlinearSolve.jl/actions/runs/27316871295, job "Build and Deploy Documentation".)

## Cause

The `HomotopySweep` docstring added in #962 (`lib/NonlinearSolveBase/src/homotopy_sweep.jl`, line 5) contains:

```
Natural-parameter continuation solver for a [`SciMLBase.HomotopyProblem`](@ref). The
```

`SciMLBase.HomotopyProblem`'s docstring is not included anywhere in this repo's docs build, so Documenter cannot resolve the `@ref`, and the docstring is rendered via the `HomotopySweep` entry in `docs/src/native/solvers.md`, turning this into a `:cross_references` build error.

## Pre-existing vs. new

The docs build was already failing on `master` before #962 with `:example_block` errors (`docs/src/tutorials/large_systems.md:324-338` "No matching function wrapper was found!" and four `docs/src/tutorials/snes_ex2.md` blocks failing with `FieldError: type Nothing has no field jacobian` — same errors on the previous master commit 45c88b901, run https://github.com/SciML/NonlinearSolve.jl/actions/runs/27148998298). Those are pre-existing and separate. This issue is specifically about the new `:cross_references` error introduced by #962.

## Suggested fix

Either:
- replace the `@ref` link in the `HomotopySweep` docstring with a plain code span: `` `SciMLBase.HomotopyProblem` ``, or
- make it an external reference (DocumenterInterLinks extref to the SciMLBase docs), or
- include `SciMLBase.HomotopyProblem` in an `@docs` block in this repo's docs.

The first option is the minimal one-line fix.

ChrisRackauckas-Claude:
[...]

PR #973, opened 2026-06-13
Add `ArcLengthContinuation`: pseudo-arclength continuation solver for `HomotopyProblem`

ChrisRackauckas-Claude:
> [!NOTE]

Adds a second continuation solver for `HomotopyProblem`, complementing `HomotopySweep` (#962). Where `HomotopySweep` marches the homotopy parameter `λ` monotonically — and therefore cannot pass a **fold** (a turning point where the solution branch reverses in `λ`) — `ArcLengthContinuation` tracks the solution curve `H(u, λ) = 0` by **arclength** in the augmented `(u, λ)` space. This is the method behind OpenModelica's global homotopy and the classic predictor-corrector path trackers (Keller; Allgower & Georg), and it was the first follow-up listed in the `HomotopySweep` PR thread.

## How it works

Each continuation step:

- **Predictor** — a secant through the last two accepted points (derivative-free, bootstrapped from a pure-`λ` step). No `∂H/∂λ` required.
- **Corrector** — solve the augmented `(n+1)` system with the `inner` solver:
  ```
  H(u, λ)                = 0      # n homotopy equations
  τ · ([u; λ] - x₀) - Δs = 0      # Keller pseudo-arclength constraint
  ```
  Because `λ` is a *free corrector variable* (not held fixed as in a sweep), the augmented Jacobian stays nonsingular at folds where `∂H/∂u` is singular — that is precisely what lets the path round a turning point. The inner solver supplies the needed derivatives by differentiating the augmented residual, exactly as for any `NonlinearProblem`.
- **Curvature control (`max_angle`)** — the per-step turn of the path direction is bounded; a sharper turn rejects the step and halves `Δs`, while `Δs` only grows on near-straight stretches. The curve is *smooth in arclength even at a fold*, so this forces small steps through a turning point while permitting large steps elsewhere — and it stops the secant predictor from overshooting onto a different branch (path-jumping). This is the analogue of OpenModelica's homotopy bend parameter, and it was necessary: without it the predictor overshoots and the corrector jumps branches.

The target is the curve point at `λ = λspan[2]`: the tracker follows the path until a step brackets that `λ`, then does one final `λ`-fixed correction to land on it exactly. A `maxsteps` guard guarantees termination, which is *required* here because the path is not monotone in `λ` (a closed loop or a branch escaping to infinity would otherwise never stop).

## Verification (local, Julia 1.11)

Canonical S-fold `u³ - 3u = -3 + 6λ`, folds at `u = ±1` (`λ = 5/6`, `1/6`). The connected branch from the λ=0 lower-sheet root (`u ≈ -2.104`) reaches the λ=1 upper-sheet root (`u ≈ +2.104`) only by traversing **non-monotone λ**: it rises past the first fold at `λ = 5/6`, reverses down toward the second at `λ = 1/6`, then climbs to `1`. The recorded corrector `λ`-trace confirms this (rises > 0.8, dips back below 0.3, then reaches 1). `ArcLengthContinuation` returns the correct root with target-system residual `~1e-13`.

Also verified: fold-free problems match `HomotopySweep`; Float32 stays Float32; in-place `f(du,u,p,λ)` and multi-dimensional (`n=2`) homotopies work; unreachable targets return a failure retcode instead of hanging.

Note on path-jumping vs. the sweep: on a *smooth scalar* fold problem the sweep's global polyalgorithm corrector often still reaches the same final root by jumping across the fold (for `λ` past the fold there is a single distant real root it converges to), so a "wrong final answer" head-to-head is not a clean demonstration on these examples. The robust, testable distinction is that `ArcLengthContinuation` *follows the connected branch* (non-monotone `λ`), which the test asserts directly; branch-faithful tracking is the property that matters on multi-branch / vector systems where jumping lands on a genuinely different solution.

## Tests

`test/Core/arclength_tests__item1-5.jl` (registered in `test/runtests.jl`), 48 assertions, all passing locally:
1. construction defaults + kwarg validation
2. fold-free happy path, matches `HomotopySweep`, decreasing `λspan`
3. **fold rounding** — S-curve, correct root + asserted non-monotone `λ` traversal
4. Float32 preservation, in-place residual, multi-dimensional system
5. unreachable target / tiny `maxsteps` → failure retcode, no hang

`NonlinearSolveBase` QA (Aqua + ExplicitImports) passes 14/14; Runic clean. `NonlinearSolveBase` bumped 2.31.0 → 2.32.0.

> Heads-up: #967 also bumps `NonlinearSolveBase` to 2.32.0 (it branched from the same master). Whichever merges second will need a re-bump to 2.33.0 — flagging to avoid a silent version collision.

## Possible follow-ups (out of scope)

- A true tangent predictor (nullvector of the augmented Jacobian) for higher-order accuracy, when `∂H/∂λ` is cheaply available.
- Weighted arclength (scaling `u` vs `λ`) for badly-scaled systems.
- Exposing a `solve(prob, ArcLengthContinuation())` default route is intentionally *not* added — `HomotopySweep` remains the `HomotopyProblem` default; arclength is opt-in.

### DIFF
--- a/docs/src/native/solvers.md
+++ b/docs/src/native/solvers.md
@@ -54,6 +54,7 @@ LimitedMemoryBroyden
 
 ```@docs
 HomotopySweep
+ArcLengthContinuation
 ```
 
 ## Nonlinear Least Squares Solvers
--- a/lib/NonlinearSolveBase/Project.toml
+++ b/lib/NonlinearSolveBase/Project.toml
@@ -1,6 +1,6 @@
 name = "NonlinearSolveBase"
 uuid = "be0214bd-f91f-a760-ac4e-3421ce2b2da0"
-version = "2.32.0"
+version = "2.33.0"
 authors = ["Avik Pal <avikpal@mit.edu> and contributors"]
 
 [deps]
--- a/lib/NonlinearSolveBase/src/NonlinearSolveBase.jl
+++ b/lib/NonlinearSolveBase/src/NonlinearSolveBase.jl
@@ -65,6 +65,7 @@ include("tracing.jl")
 include("wrappers.jl")
 include("polyalg.jl")
 include("homotopy_sweep.jl")
+include("arclength.jl")
 
 
 include("descent/common.jl")
@@ -115,7 +116,7 @@ export DescentResult, SteepestDescent, NewtonDescent, DampedNewtonDescent, Dogle
 
 export NonlinearSolvePolyAlgorithm
 
-export HomotopySweep
+export HomotopySweep, ArcLengthContinuation
 
 export NonlinearVerbosity
 
--- a/lib/NonlinearSolveBase/src/arclength.jl
+++ b/lib/NonlinearSolveBase/src/arclength.jl
@@ -216,8 +216,8 @@ function CommonSolve.solve(
         # points once history exists; a pure-λ step toward λend bootstraps the first step.
         if have_prev
             d = xcur .- xprev
-            nd = norm(d)
-            τ = nd > 0 ? d ./ nd : pack(zeros(Tx, n), sλ)
+            dnorm = norm(d)
+            τ = dnorm > 0 ? d ./ dnorm : pack(zeros(Tx, n), sλ)
         else
             τ = pack(zeros(Tx, n), sλ)
         end
--- a/lib/NonlinearSolveBase/src/arclength.jl
+++ b/lib/NonlinearSolveBase/src/arclength.jl
@@ -1,0 +1,304 @@
+"""
+    ArcLengthContinuation(; inner = nothing, initial_step_factor = 0.1,
+        adaptive = true, min_ds = nothing, max_step_factor = 1.0,
+        expand_factor = 2.0, expand_threshold = 2, max_angle = π / 6,
+        maxsteps = 10000)
+
+Pseudo-arclength continuation solver for a `SciMLBase.HomotopyProblem`. Unlike
+[`HomotopySweep`](@ref), which marches the scalar parameter ``λ`` monotonically, this
+solver tracks the solution curve ``H(u, λ) = 0`` parameterized by arclength ``s`` in the
+augmented ``(u, λ)`` space. Each step takes a predictor step along the path and corrects
+it by solving the *augmented* ``(n+1)``-dimensional system
+
+```
+H(u, λ)                 = 0          # n equations
+τ ⋅ ([u; λ] - x₀) - Δs  = 0          # Keller pseudo-arclength constraint
+```
+
+with the `inner` solver. Because ``λ`` is a free variable of the corrector (not held
+fixed), the augmented Jacobian stays nonsingular at *turning points* (folds) where
+``∂H/∂u`` is singular and ``λ`` is non-monotone along the path. This lets the solver
+round folds that defeat natural-parameter continuation — the canonical reason a sweep
+"fails to reach `λ = 1`" when a real solution at `λ = 1` does exist but only on a branch
+reachable by going around a fold.
+
+The target is the point on the curve where ``λ = λspan[2]``: the solver follows the path
+until a step brackets that ``λ``, then performs one final ``λ``-fixed correction to land
+on it exactly.
+
+Keyword arguments:
+
+  - `inner`: the inner nonlinear algorithm used for both the initial on-curve correction
+    and the augmented corrector; `nothing` selects NonlinearSolve's default polyalgorithm.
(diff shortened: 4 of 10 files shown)


ChrisRackauckas-Claude:
## CI status

**Green on CI** (previous run, before the lts fix): `tests / Core` on **julia 1 and julia pre** (ubuntu + macos), `Spell Check` (after the `nd`→`dnorm` rename), `Runic Suggestions`, `Downgrade`. Runic Format Check passes on all six files this PR adds (verified in its log; the job's failure is the pre-existing #969 on unrelated sublibrary `runtests.jl` files).

**One real PR failure, fixed:** `tests / Core (julia lts)` failed with `UndefVarError: ArcLengthContinuation not defined`. This is the Julia < 1.11 `[sources]` issue (same class as #967): the root test env ignores `[sources]`, so the base-env Core group resolved the *registered* `NonlinearSolveBase` 2.31.0 — which has no `ArcLengthContinuation` — instead of this branch's 2.32.0. Fixed in 83907b6eb by calling `_develop_inrepo_sources()` in the base-env path (the sublibrary and dep-adding paths already did this; same one-liner as #967, so expect a trivial merge with whichever lands first).

**Verification of the fix (local):** ran the full root `GROUP=Core` suite via `Pkg.test()` on Julia 1.10.11 — `Testing NonlinearSolve tests passed`, exit 0, with all five new testsets green:

```
ArcLengthContinuation construction + defaults                 | 25  25
ArcLengthContinuation happy path (fold-free, matches sweep)   |  6   6
ArcLengthContinuation rounds a fold (non-monotone λ)          |  5   5
ArcLengthContinuation Float32 / in-place / multi-dim          |  8   8
ArcLengthContinuation fails (no hang) on an unreachable target|  4   4
```

The CI rerun on 83907b6eb is still **queued** (the shared self-hosted runners have been heavily backlogged for hours); I have not yet seen the lts Core job run on this exact commit in CI. Flagging that explicitly rather than claiming a CI-green I haven't observed — the fix is verified locally in the matching environment (Julia 1.10, `Pkg.test`), and julia-1/pre already passed on CI.

**Remaining red jobs — all pre-existing or infra, none caused by this PR:**
- `Documentation` — `:cross_references` error is `[`SciMLBase.HomotopyProblem`](@ref)` in the **`HomotopySweep`** docstring (the #968 bug; its fix is in unmerged #967), plus the pre-existing `:example_block` errors in `large_systems.md` / `snes_ex2.md`. This PR's `ArcLengthContinuation` docstring uses a plain code span for `SciMLBase.HomotopyProblem` and `@ref` only to `HomotopySweep`, so it adds no new cross-ref error.
- `Runic Format Check` — pre-existing #969.
- `Trim` (DiffEqBase 6/7 compat), `downgrade-sublibraries` (FastClosures lower bound) — pre-existing compat staleness, red on master.
- `NonlinearSolveHomotopyContinuation` sublibrary — pre-existing, fix in #970 (+ #971 for the lts QA artifact).
- `CUDA` — tests pass; job fails in the Codecov step on a `~/.gitconfig` lock race (self-hosted infra).
- `BoundaryValueDiffEq` / `ModelingToolkit` / `OrdinaryDiffEq` downstream — downstream-red (BVDE master) / MTK BifurcationKit `UndefKeywordError: keyword argument P not assigned` / lost-runner infra.

ChrisRackauckas-Claude:
Rebased onto latest master (now at 7a1fb11ee). The branch is 2 clean commits and `mergeable` again. Notable points from the rebase:

- **Dropped the obsolete lts test-fix commit.** Master's `runtests.jl` was rewritten to dispatch through `SciMLTesting.run_tests`, which handles the Julia < 1.11 `[sources]` develop centrally — so my earlier `_develop_inrepo_sources()` base-env fix is no longer needed. The arclength test registration was re-grafted into the new `core` closure.
- **`NonlinearSolveBase` 2.31.2 → 2.32.0** (feature bump on top of master's latest release bump).
- **`#970` (the InteractiveUtils fix) merged** in the meantime, so it's now part of the base.

Verified after rebase: full root `GROUP=Core` suite via `Pkg.test()` on Julia 1.11 passes — `Testing NonlinearSolve tests passed`, all five ArcLengthContinuation testsets green (25+6+5+8+4 = 48 assertions); load+solve sanity check on the new base passes; Runic clean on all changed files.

Heads-up on a **pre-existing master issue** (not introduced here): running the root `Core` group via `Pkg.test()` on **Julia 1.10** currently fails during manifest validation with `BracketingNonlinearSolve depends on ChainRulesCore, but no such entry exists in the manifest` — this reproduces on a clean `origin/master` checkout (unrelated to this PR; ChainRulesCore is a weakdep of BracketingNonlinearSolve). Flagging in case the lts CI lane is affected.

PR #1001, opened 2026-07-04
Add mass-matrix damping to PseudoTransient

pepijndevos:
Generalize the pseudo-transient continuation damping term from the identity `(1/α) I` to a user-supplied mass matrix `(1/α) M`, so the descent solves `(J + (1/α) M) δu = -F`. This extends ψtc to differential-algebraic (DAE) steady-state problems (Coffey, Kelley & Keyes, 2003), where `M` is the mass matrix of the underlying `M u' = -F(u)` dynamics — e.g. MNA circuit systems.

Changes:
- `PseudoTransient(; mass_matrix = nothing, ...)`: new keyword. When `nothing`, behavior is bit-for-bit identical to before; if the problem's `NonlinearFunction` carries a non-identity `mass_matrix`, it is picked up automatically.
- `SwitchedEvolutionRelaxation(mass_matrix)`: the SER damping function now returns `(1/α) M` when a mass matrix is supplied, keeping the SER step-growth rule unchanged.
- `dampen_jacobian!!`: split the diagonal (`Number`/`Diagonal`) fast paths from a new general `AbstractMatrix` path that adds the full damping matrix. A diagonal mass matrix keeps the efficient diagonal update; a general sparse/dense matrix is fully supported.
- Tests covering regression (nothing == identity), diagonal and dense mass matrices, in-place problems, and auto-resolution from the `NonlinearFunction` mass matrix.

Claude-Session: https://claude.ai/code/session_017zCWMioFzF3DVwXrVCkCk4

## Checklist

### DIFF
--- a/lib/NonlinearSolveBase/src/descent/damped_newton.jl
+++ b/lib/NonlinearSolveBase/src/descent/damped_newton.jl
@@ -239,26 +239,42 @@ function InternalAPI.solve!(
 end
 
 dampen_jacobian!!(::Any, J::Union{AbstractSciMLOperator, Number}, D) = J + D
-function dampen_jacobian!!(J_cache, J::AbstractMatrix, D::Union{AbstractMatrix, Number})
+
+# Scalar damping (identity-style `(1/α) I` damping) only touches the diagonal of `J`.
+function dampen_jacobian!!(J_cache, J::AbstractMatrix, D::Number)
     ArrayInterface.can_setindex(J_cache) || return J .+ D
     J_cache !== J && copyto!(J_cache, J)
     if ArrayInterface.fast_scalar_indexing(J_cache)
-        if D isa Number
-            @simd ivdep for i in axes(J_cache, 1)
-                @inbounds J_cache[i, i] += D
-            end
-        else
-            @simd ivdep for i in axes(J_cache, 1)
-                @inbounds J_cache[i, i] += D[i, i]
-            end
+        @simd ivdep for i in axes(J_cache, 1)
+            @inbounds J_cache[i, i] += D
         end
     else
         idxs = diagind(J_cache)
-        if D isa Number
-            J_cache[idxs] .+= D
-        else
-            J_cache[idxs] .+= @view(D[idxs])
-        end
+        J_cache[idxs] .+= D
     end
+    return J_cache
+end
+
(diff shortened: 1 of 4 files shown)


ChrisRackauckas (approved):

pepijndevos:
I marked it as draft to go an do some testing on real daes in Cadnip to see how it behaves in practice. Also seems like ci is a bit unhappy.

ChrisRackauckas:
The algorithm is correct though. And it's a published thing that supposedly helps some cases. So we might as well merge it as a correct implementation of an applicable algorithm regardless of whether it helps your specific case.

pepijndevos:
Alright

pepijndevos:
Tested this against real MNA circuit mass matrices from Cadnip.jl (dev'd `NonlinearSolveBase`+`NonlinearSolveFirstOrder` from this branch directly, not just the stock `mass_matrix` kwarg on `NonlinearFunction` which predates this PR and can silently no-op if you don't also set `concrete_jac=true`).

Results on two real circuits (BJT bias networks, DC operating point):

- **Simple single-BJT bias circuit** (n=14, mostly-dense mass matrix from junction caps): identity-damped `PseudoTransient` never leaves `u=0` in 200 iterations — completely stuck. With `mass_matrix=C` (the real circuit mass matrix) it converges in 6 steps, matching plain `NewtonRaphson` exactly. Clear win.
- **Harder circuit** (n=24, monostable multivibrator with excess-phase BJT model, only 4/24 rows differential): identity damping again gets stuck at `u=0`. With the real mass matrix, it takes a much larger first step (since the ~20 algebraic rows get *zero* damping) and blows up into the same exponential overflow (`~1e75` residual) that plain undamped Newton hits on this circuit, bailing out with `Unstable` after 5 iterations instead of grinding uninformatively for 2000.

So this is a real, reproducible improvement for circuits where the mass matrix meaningfully separates dynamics from algebra, but for circuits where most of the danger lives in the algebraic rows (nonlinear device equations with no local capacitance), it trades a slow non-convergence for a fast one — no worse in practice (still fails), but worth knowing that "mass-matrix damping" isn't a free win when M is mostly zero on the rows that matter.

Thanks for merging this — we'll likely pick it up as one additional member in our `NonlinearSolvePolyAlgorithm` fallback chain (not a replacement) once a release is cut, since poly-algorithms just move on to the next member on failure either way.

PR #1014, opened 2026-07-08
Limit StaticArray CUDA limited-memory Broyden smoke test

ChrisRackauckas-Claude:

## Summary
- Keep scalar CUDA kernel-smoke coverage for `SimpleLimitedMemoryBroyden` at the default `threshold = Val(27)`.
- Use `threshold = Val(2)` only for StaticArray `SimpleLimitedMemoryBroyden` kernel-smoke variants, including the line-search variant.

## Context
The latest Sublibrary CI failure was `lib/SimpleNonlinearSolve [CUDA]`, where the self-hosted GPU runner OOMed while launching the `SVector{2,Float32}` kernel for `SimpleLimitedMemoryBroyden{..., Val{27}, ...}`. The log reported CUDA memory pool usage of 0 bytes, pointing to CUDA compile/module pressure rather than runtime array allocation.

## Validation
- `timeout 3600 env NONLINEARSOLVE_TEST_GROUP=CUDA ~/.juliaup/bin/julia +1 --project=. -e 'using Pkg; Pkg.test(; julia_args=["--check-bounds=auto"])'` from `lib/SimpleNonlinearSolve`\n  - Passed locally, but both CUDA testsets reported `Total 0` because this machine has `CUDA.functional() = false` and `CUDA.has_cuda() = false`.\n- `~/.juliaup/bin/julia +1 --startup-file=no -e 'Meta.parseall(read("lib/SimpleNonlinearSolve/test/gpu/cuda_tests__item2.jl", String)); println("parse ok")'`\n- `git diff --check`\n- `Runic.format_file(...; inplace=true)` on the changed file from a temporary Julia environment with Runic v1.7.0\n

### DIFF
--- a/lib/SimpleNonlinearSolve/test/gpu/cuda_tests__item2.jl
+++ b/lib/SimpleNonlinearSolve/test/gpu/cuda_tests__item2.jl
@@ -13,7 +13,18 @@ if CUDA.functional()
         return nothing
     end
 
+    limited_memory_broyden_for_kernel(::Type{<:Number}; linesearch = nothing) =
+        SimpleLimitedMemoryBroyden(; linesearch)
+
+    function limited_memory_broyden_for_kernel(::Type{<:StaticArray}; linesearch = nothing)
+        # The StaticArray path unrolls one generated block per threshold. Keep the
+        # kernel smoke test small enough for 16 GiB CI GPUs; scalar coverage keeps
+        # the default threshold.
+        return SimpleLimitedMemoryBroyden(; threshold = Val(2), linesearch)
+    end
+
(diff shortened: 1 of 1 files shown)


ChrisRackauckas-Claude:
Scratchpad update: the SimpleNonlinearSolve CUDA sublibrary lane passed in CI (`lib/SimpleNonlinearSolve [CUDA]`, 22m56s), which is the lane targeted by this threshold change. The remaining red checks are unrelated to the CUDA patch: MTK downstream is covered by SciML/.github#110, and the Adjoint resolver failure is covered by the temporary NonlinearSolve guard in #1017 plus the upstream SciMLBase FunctionProperties guard in SciMLBase#1424.

PR #1019, opened 2026-07-09
Add HomotopyPolyAlgorithm: staged polyalgorithm for HomotopyProblem

ChrisRackauckas-Claude:
> [!NOTE]

## Summary

Adds `HomotopyPolyAlgorithm` to NonlinearSolveBase: a polyalgorithm for `SciMLBase.HomotopyProblem` that tries a tuple of continuation algorithms in order and returns the first solution with a successful retcode. If every stage fails, the **last** stage's failed solution is returned (its `retcode`/`original` describe the most robust attempt).

The zero-argument default encodes the natural escalation for homotopy solves:

```julia
HomotopyPolyAlgorithm() == HomotopyPolyAlgorithm((HomotopySweep(), ArcLengthContinuation()))
```

- `HomotopySweep` is the cheap first attempt — natural-parameter continuation marches λ monotonically and reuses one inner-solver cache across the whole sweep, but it cannot reverse λ and therefore cannot follow a branch around a fold (turning point).
- `ArcLengthContinuation` is the robust fallback — it tracks the curve by pseudo-arclength in the augmented `(u, λ)` space, so λ is free to decrease and folds that defeat the sweep are rounded, at the higher cost of an `(n+1)`-dimensional corrector per step.

## Design

- New file `lib/NonlinearSolveBase/src/homotopy_polyalg.jl`, modeled loosely on `NonlinearSolvePolyAlgorithm` (`polyalg.jl`): a `@concrete struct HomotopyPolyAlgorithm <: AbstractNonlinearSolveAlgorithm` holding `algs <: Tuple`, plus the zero-argument default constructor. Deliberately kept simple: no cache/`init!` machinery — `CommonSolve.solve(prob::SciMLBase.HomotopyProblem, alg::HomotopyPolyAlgorithm, args...; kwargs...)` loops the stages and forwards `args...; kwargs...` to each.
- Exported from NonlinearSolveBase (reexported by NonlinearSolve) and added to the `@docs` block in `docs/src/native/solvers.md` under "Homotopy / Continuation Solvers".
- **The existing default routing is unchanged**: `solve(prob::HomotopyProblem, ::Nothing)` still selects `HomotopySweep()`. The polyalgorithm is opt-in. Whether the default should become `HomotopyPolyAlgorithm()` is left as a separate decision for the maintainer — it would make default homotopy solves strictly more robust, but adds the arclength attempt's cost to every currently-failing sweep and changes which `retcode`/`u` failing solves report.
- No Project.toml version bumps (per maintainer preference).

## Tests

New `test/Core/homotopy_polyalg_tests__item1.jl` (registered in `test/runtests.jl` in the core group), covering:

1. Construction/defaults (`algs[1] isa HomotopySweep`, `algs[2] isa ArcLengthContinuation`) and a custom stage tuple.
2. Happy path where stage 1 succeeds — with a `Ref` residual-evaluation counter proving the arclength stage never ran (the polyalg's evaluation count exactly equals a plain `HomotopySweep()` solve's count: 150 == 150 observed).
3. Fallback path on the S-fold `u³ - 3u = -3 + 6λ` (turning points at λ = 5/6 and 1/6): stage 1 fails, stage 2 rounds both folds and lands on the connected upper-sheet root `u ≈ +2.1038034` with target residual < 1e-6.
4. Both-stages-fail path on `(1-λ)u + λ(u²+1)` (no real root at λ = 1): `!successful_retcode`, finite `u`, and `sol.alg isa ArcLengthContinuation` (the last stage's solution).
5. Zero-width λspan and a Float32 problem (eltype preserved).

### Empirical calibration note (path jumping)

The fallback test needed calibration: with `inner = SimpleNewtonRaphson()` and unbounded iterations, `HomotopySweep` **path-jumps** on the S-fold — Newton on the cubic, warm-started from near the fold at `u = -1`, eventually reaches the far root on the upper sheet, so the sweep reports `Success` at `u ≈ +2.1038` rather than failing. The test therefore passes `maxiters = 10` through the solve kwargs: a budget that comfortably suffices for on-path warm-started corrector steps but not for the wild excursion of a jump. Observed on a `maxiters` scan of 2–10: the sweep fails with `MaxIters` at every value (stalling at `u ≈ -1.13`, next to the fold at `u = -1`) while `ArcLengthContinuation` succeeds at every value with target residual ~1e-15.

## What was run locally

- Full homotopy battery via `@safetestset` includes on Julia 1.11: all 21 `homotopy_sweep_tests__item*.jl`, all 6 `arclength_tests__item*.jl`, and the new `homotopy_polyalg_tests__item1.jl` — **185 Pass, 185 Total, 0 failures** (Test Summary: `homotopy battery | Pass 185 Total 185`, ~4m50s; run twice, the second time on the exact Runic-formatted code being committed).
- `Runic.main(["--check", ...])` exits 0 on every touched file.

### DIFF
--- a/docs/src/native/solvers.md
+++ b/docs/src/native/solvers.md
@@ -55,6 +55,7 @@ LimitedMemoryBroyden
 ```@docs
 HomotopySweep
 ArcLengthContinuation
+HomotopyPolyAlgorithm
 ```
 
 ## Nonlinear Least Squares Solvers
--- a/lib/NonlinearSolveBase/src/NonlinearSolveBase.jl
+++ b/lib/NonlinearSolveBase/src/NonlinearSolveBase.jl
@@ -66,6 +66,7 @@ include("wrappers.jl")
 include("polyalg.jl")
 include("homotopy_sweep.jl")
 include("arclength.jl")
+include("homotopy_polyalg.jl")
 
 
 include("descent/common.jl")
@@ -116,7 +117,7 @@ export DescentResult, SteepestDescent, NewtonDescent, DampedNewtonDescent, Dogle
 
 export NonlinearSolvePolyAlgorithm
 
-export HomotopySweep, ArcLengthContinuation
+export HomotopySweep, ArcLengthContinuation, HomotopyPolyAlgorithm
 
 export NonlinearVerbosity
 
--- a/lib/NonlinearSolveBase/src/homotopy_polyalg.jl
+++ b/lib/NonlinearSolveBase/src/homotopy_polyalg.jl
@@ -1,0 +1,62 @@
+"""
+    HomotopyPolyAlgorithm(algs::Tuple)
+    HomotopyPolyAlgorithm()
+
+A polyalgorithm for [`SciMLBase.HomotopyProblem`](@ref): a container for a tuple of
+continuation algorithms that are tried in order until one returns a solution with a
+successful retcode. The first success is returned immediately — later stages never run.
+If every stage fails, the *last* stage's failed solution is returned, so its `retcode`
+(and `original`, when the stage attaches one) describe the most robust attempt.
+
+The zero-argument form defaults to
+
+```julia
+HomotopyPolyAlgorithm((HomotopySweep(), ArcLengthContinuation()))
+```
+
+which encodes the natural escalation for homotopy solves: [`HomotopySweep`](@ref) is the
+cheap first attempt — natural-parameter continuation marches the scalar ``λ``
+monotonically across `λspan`, reusing one inner-solver cache across all steps, but it can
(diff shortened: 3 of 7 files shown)


ChrisRackauckas-Claude:
Per maintainer decision, the default routing is now included in this PR: `solve(prob::HomotopyProblem)` / `solve(prob, nothing)` route to `HomotopyPolyAlgorithm()` (sweep → arclength fallback) instead of bare `HomotopySweep()`. Happy path unchanged (sweep runs first, success short-circuits); fold-blocked problems now succeed under the default when arclength can round the fold. Full homotopy battery (21 sweep + 6 arclength + polyalg items) passes 185/185 locally on the committed code.

Issue #1020, opened 2026-07-10
Homotopy continuation stack: performance/robustness improvement roadmap


ChrisRackauckas-Claude:
# Improving the homotopy-continuation stack: research roadmap

Survey of concrete performance/robustness improvements for `HomotopySweep`, `ArcLengthContinuation`, and `SimpleHomotopySweep`, prioritized by impact × feasibility. Algorithm details verified against current sources of BifurcationKit.jl, HomotopyContinuation.jl (+ Timme 2021, arXiv:1902.02968), HOMPACK90, AUTO-07p, MatCont, and OpenModelica's `nonlinearSolverHomotopy.c`; the two headline numbers were measured locally against this repo.

## Priority 1 — Corrector-effort-driven step control + cheap rejection (perf + robustness, S/M)

Two coupled changes to all three drivers:

1. **Cap the inner corrector's iteration budget during tracking steps** (a `tracking_maxiters`, default ~10–20; never applied to the anchor or final λ-end solve, never overriding user-passed `maxiters`).
2. **Adapt dλ/ds on corrector iteration count on success** — today a 9-iteration success and a 2-iteration success feed the growth logic identically (beyond the ≤2 bypass), and the step never shrinks proactively; the only shrink signal is an outright failure at full inner-solver cost.

Measured on a fold problem where the sweep must fail (`u² + λ − 0.5 = 0`), total residual calls for the failing sweep:

```
default polyalg inner              retcode=MaxIters  residual calls = 89349
NewtonRaphson inner                retcode=MaxIters  residual calls = 46173
NewtonRaphson inner, maxiters=20   retcode=MaxIters  residual calls =  1093
polyalg inner, maxiters=20         retcode=MaxIters  residual calls =  3294
```

~80× reduction in wasted work per rejection ladder: every bisection retry currently re-runs the full `FastShortcutNonlinearPolyalg` ladder against `maxiters = 1000`. Reference practice: OpenModelica `homMaxNewtonSteps = 20`; HomotopyContinuation.jl caps Newton at 11 (typically runs 2); MatCont `MaxCorrIters = 10`; AUTO `ITNW = 5`.

[...]

PR #1021, opened 2026-07-10
ArcLengthContinuation: bordered tangent solve + θ-weighted arclength metric (#1020 item 3, Priorities 3+4)

ChrisRackauckas-Claude:
> [!IMPORTANT]
> **This PR should be ignored until it has been reviewed by @ChrisRackauckas.** Draft, not ready for merge.

Implements roadmap item 3 of the top-3 recommendation in #1020: **Priority 3 (bordered linear solve for the arclength tangent) and Priority 4 (θ-weighted arclength metric)**, shipped together as one "arclength at scale" change since they share the same semantic break.

## Part A — bordered tangent solve replacing `nullspace` (Priority 3)

`_arclength_tangent` computed the path tangent of the `:tangent` predictor as `LinearAlgebra.nullspace(J)`: a dense SVD — O(n³) with a large constant, allocating, and erroring for sparse `J`. It is replaced by the standard bordered solve (BifurcationKit `Bordered` tangent; Govaerts 1991 block elimination):

```
[J; wᵀ] t = e_{n+1},    w = θ-metric weighting of τ_prev
```

i.e. `J t = 0` with the normalization `⟨τ_prev, t⟩_θ = 1`, via one LU factorization, then θ-normalized and oriented by `sign(⟨t, τ_prev⟩_θ)`. The previous tangent `τ_prev` was already threaded through the driver as the orientation reference (the pure-λ seed provides the first bordering row), so no new state is needed.

**SVD fallback kept, never on the happy path:** the bordered matrix is singular exactly when `τ_prev` is θ-orthogonal to the true tangent (e.g. the pure-λ first-step seed meeting a fold head-on) or at a branch point where `null(J)` is 2-dimensional — the case for which BifurcationKit refuses the bordered tangent. A failed (`issuccess(lu(B; check = false)) == false`) or non-finite bordered solve falls back to `nullspace(Matrix(J))`, which stays robust there.

Reproduced timing (this machine, random n×(n+1) augmented Jacobians, 5-run averages, matching the issue's benchmark protocol; issue measured 11–17×):

```
n= 200  nullspace(SVD):    17.56 ms   1908 KiB | bordered LU:    1.565 ms    638 KiB  (11.2x)
n= 800  nullspace(SVD):   309.66 ms  30132 KiB | bordered LU:   35.649 ms  10051 KiB  ( 8.7x)
direction agreement (θ-dot of the two unit tangents): 1 - |⟨t_svd, t_lu⟩_θ| = 0.0
```

The bordered path also works verbatim with sparse Jacobians / any LU-capable backend, unlocking sparse `:tangent` mode (`nullspace` throws for sparse input).

## Part B — θ-weighted arclength metric (Priority 4) — **semantic change**

Every place the path metric appeared — the Keller constraint row in `AugmentedHomotopy`, tangent/secant normalization, orientation sign, realized chord length, and the bend-angle `cosang` — used the unweighted Euclidean dot on `[u; λ]`, so for large `n` the state block swamps the single λ component. All of them now consistently use the BifurcationKit `DotTheta` convention:

```
⟨(u₁,λ₁), (u₂,λ₂)⟩_θ = (θ/n)·⟨u₁,u₂⟩ + (1−θ)·λ₁λ₂,    n = length(u)
```

with a new `theta` kwarg on `ArcLengthContinuation` (default `0.5`, validated `0 < θ < 1`). The `1/n` normalization is the crucial automatic default: it makes the u-block/λ balance independent of system size.

> **⚠️ What this changes numerically:** `ds`-valued options (`initial_step_factor`, `min_ds`, `max_step_factor`) and `max_angle` are now measured in the weighted metric, not the Euclidean one. For a pure-λ motion a weighted `ds` corresponds to a λ-distance of `ds/sqrt(1−θ)`; for pure-u motion, a Euclidean u-distance of `ds/sqrt(θ/n)`. Values tuned against the previous unweighted geometry may need rescaling. This is documented in a dedicated "metric" admonition in the `ArcLengthContinuation` docstring. Final roots are unchanged on all existing tests (see below) — only the step geometry along the way changes.

The weighted dot is written on views (no allocation, no eltype forcing), so ForwardDiff duals pass through the constraint row and Float32 problems stay Float32 (tested).

## Tests

New `test/Core/arclength_tests__item7.jl` (registered in the core closure after item 6):

- unit tests of `_bordered_tangent` on known 1×2 Jacobians: `Jτ = 0`, θ-unit normalization, orientation, agreement with the analytic null direction at a regular point, and the ~vertical-in-λ tangent at the fold of `u³ − 3u = −3 + 6λ`;
- SVD fallback engaging without error on the exactly-singular bordered system (`τ_prev` θ-orthogonal to the tangent), unit-level and end-to-end (fold exactly at the start point on `u² = λ` with `:tangent`);
- `theta` validation throws on `0`, `1`, `-0.5`, `1.5`; default and `θ = 0.8` land on the previous roots of the S-fold and monotone standard problems (values, not step counts);
- an n = 50 fold-free coupled system solved with `:tangent` (exercises the 1/n-scaled metric + bordered solve at moderate n);
- Float32 `:tangent` preserves eltype.

All existing arclength (items 1–6) and homotopy-sweep (items 1–21) test files pass locally unmodified — **no recalibration was needed**: at n = 1 (most existing tests) θ = 0.5 scales the metric uniformly by √0.5, which changes effective step sizes but no test asserted absolute step geometry; all assertions are on final roots, retcodes, and qualitative path behavior (fold rounding), which are unchanged.

Fixes part of #1020.

### DIFF
--- a/lib/NonlinearSolveBase/ext/NonlinearSolveBaseLinearSolveExt.jl
+++ b/lib/NonlinearSolveBase/ext/NonlinearSolveBaseLinearSolveExt.jl
@@ -11,6 +11,8 @@ using LinearAlgebra: ColumnNorm, Symmetric
 
 using NonlinearSolveBase: NonlinearSolveBase, LinearSolveJLCache, LinearSolveResult, Utils, NonlinearVerbosity, InternalAPI, LinearSolveParameters
 
+Utils.is_extension_loaded(::Val{:LinearSolve}) = true
+
 function (cache::LinearSolveJLCache)(;
         A = nothing, b = nothing, linu = nothing,
         reuse_A_if_factorization = false, kwargs...
--- a/lib/NonlinearSolveBase/src/arclength.jl
+++ b/lib/NonlinearSolveBase/src/arclength.jl
@@ -2,8 +2,8 @@
     ArcLengthContinuation(; inner = nothing, initial_step_factor = 0.1,
         adaptive = true, min_ds = nothing, max_step_factor = 1.0,
         expand_factor = 2.0, expand_threshold = 2, max_angle = π / 6,
-        predictor = :secant, autodiff = nothing, tracking_maxiters = 10,
-        maxsteps = 10000, theta = 0.5)
+        predictor = :secant, autodiff = nothing, linsolve = nothing,
+        tracking_maxiters = 10, maxsteps = 10000, theta = 0.5)
 
 Pseudo-arclength continuation solver for a `SciMLBase.HomotopyProblem`. Unlike
 [`HomotopySweep`](@ref), which marches the scalar parameter ``λ`` monotonically, this
@@ -92,6 +92,9 @@ Keyword arguments:
   - `autodiff`: the automatic-differentiation backend (an `ADTypes.AbstractADType`) used
     to form the augmented Jacobian for the `:tangent` predictor; `nothing` (default)
     selects `AutoForwardDiff()`. Unused by the `:secant` predictor.
+  - `linsolve`: the LinearSolve.jl algorithm for the `:tangent` predictor's bordered
+    solve (the same knob the Newton descent methods expose); `nothing` (default) selects
+    LinearSolve's default. Unused by the `:secant` predictor.
   - `tracking_maxiters`: iteration cap for the augmented corrector solves (default 10,
     in the range used by MatCont, HomotopyContinuation.jl, and OpenModelica;
     `nothing` disables). A rejected step retries at half the arclength increment from a warm
@@ -129,6 +132,7 @@ Jacobian.
     max_angle
     predictor::Symbol
     autodiff
+    linsolve
     tracking_maxiters
     maxsteps::Int
     theta
@@ -138,7 +142,8 @@ function ArcLengthContinuation(;
         inner = nothing, initial_step_factor = 0.1, adaptive = true,
         min_ds = nothing, max_step_factor = 1.0, expand_factor = 2.0,
         expand_threshold = 2, max_angle = π / 6, predictor = :secant,
-        autodiff = nothing, tracking_maxiters = 10, maxsteps = 10000, theta = 0.5
+        autodiff = nothing, linsolve = nothing, tracking_maxiters = 10,
+        maxsteps = 10000, theta = 0.5
     )
     if !(0 < initial_step_factor <= 1)
         throw(
@@ -205,7 +210,7 @@ function ArcLengthContinuation(;
     end
     return ArcLengthContinuation(
         inner, initial_step_factor, adaptive, min_ds, max_step_factor,
-        expand_factor, expand_threshold, max_angle, predictor, autodiff,
+        expand_factor, expand_threshold, max_angle, predictor, autodiff, linsolve,
         tracking_maxiters, maxsteps, theta
     )
 end
@@ -290,7 +295,7 @@ end
 # factorize.
 function _arclength_jac_cache(
         prob::SciMLBase.HomotopyProblem{uType, iip}, alg, x,
-        n
+        n, stats
     ) where {uType, iip}
     gf = SciMLBase.NonlinearFunction{iip}(HomotopyResidual(prob.f, n))
     gprob = NonlinearProblem{iip}(gf, x, prob.p)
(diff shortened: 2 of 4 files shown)


ChrisRackauckas-Claude:
Both review items addressed, and the branch is rebased for merge (`0e031500`, MERGEABLE on current master):

## Allocations in the solve — confirmed and fixed

The bordered-tangent path did allocate per predictor step rather than reusing memory: the weighted border row (`similar`), the bordered matrix (`vcat`), the rhs (`zeros`), the LU factorization (which copies `B` again), the `\` solution, and the normalized copy — ~2(n+1)² + several (n+1) vectors per step. Now (`0e031500`):

- `B` and `t` are scratch buffers **allocated once in `solve`** and reused every step: the border row is written directly into `B`'s last row, `lu!` factorizes in place, `ldiv!` solves into `t`, normalization/orientation are in place, and the result is `copyto!`d into the driver's stable `τ` buffer (τprev is only read before the copy, so aliasing is safe). The only remaining happy-path allocation is `lu!`'s internal pivot vector (O(n) Ints).
- The secant branch's per-step chord temporaries reuse the same scratch; `τ` is now a stable buffer written in place on every predictor path.
- `J` (the Jacobian cache's buffer) is never destroyed — `B` holds the copy — so the SVD fallback still reads it.

Measured (n = 200 coupled system, whole solve): tangent 75196 → 72316 KiB, the delta matching exactly the eliminated per-step matrix/LU copies; S-fold roots identical for both predictors. The remaining bulk of arclength's solve allocation is the **corrector-side per-step rebuild** (fresh augmented problem + inner solve each step) — that predates this PR and is roadmap #1020 Priority 6, kept separate.

## Rebase

Rebased onto master with #1022 and #1024 now merged — four conflicts resolved as kwarg/registration unions (the constructor now carries both `tracking_maxiters` and `theta`; test registration keeps arclength item7 alongside the jac/effort/polyalg files). The `_bordered_tangent` unit tests in item7 were updated to the new in-place API.

Verified on the exact pushed code (Julia 1.11): homotopy_sweep items 1–21 + arclength items 1–7 + jac + effort 1–2 + polyalg — 268 assertions in the main battery + 34 in the updated arc7 (which unit-tests the in-place bordered solve, the SVD fallback, and the θ metric) — all pass; Runic clean.

ChrisRackauckas-Claude:
Per review, pushed `9fc46e8a`: the bordered tangent solve now runs through the **shared Newton linear-solve stack** instead of raw `lu!`/`ldiv!`:

- The workspace holds a `construct_linear_solver` cache — the same LinearSolve.jl-backed machinery the descent methods use — built once per solve and re-driven each step via `lincache(; A, b, linu, reuse_A_if_factorization = false)`. Backend selection, factorization-workspace reuse, and `NLStats` accounting (one shared stats object across the tangent's Jacobian cache and linear cache) all come from the stack.
- `ArcLengthContinuation` gains the standard **`linsolve` kwarg** (like `NewtonRaphson` et al.), forwarded to both the bordered solve and `construct_jacobian_cache`. Covered by a new item7 test with `linsolve = QRFactorization()`.
- NonlinearSolveBase only weak-depends on LinearSolve, so when the extension isn't loaded the workspace falls back to the `\`-backed `NativeJLLinearSolveCache` (same abstraction); the ext now defines `Utils.is_extension_loaded(::Val{:LinearSolve})` to make that detectable.
- **Subtlety this surfaced:** LinearSolve's default solver rescues a singular LU with pivoted QR and returns a finite *least-squares pseudo-solution* with a success retcode — which silently defeated the designed-singular → SVD-fallback routing (item7's fold-at-start test caught the tracker receiving a bogus pure-λ tangent). Acceptance is now gated on a backend-agnostic Govaerts-style residual check (`‖[J·t; ⟨w,t⟩−1]‖` at rounding level vs O(1) for the pseudo-solution), computed from `J`/`τprev` rather than `B` (which the factorization may destroy).

Verified on the pushed code (Julia 1.11): full battery — sweep items 1–21, arclength 1–7 (incl. the restored SVD-fallback unit tests + the `linsolve` test), jac, effort 1–2, polyalg — **298/298**; Runic clean. Still MERGEABLE on current master.

PR #1024, opened 2026-07-10
Homotopy drivers: `tracking_maxiters` cap, effort-band step control, `maxsteps` guard (#1020 Priority 1)

ChrisRackauckas-Claude:
> [!NOTE]

Implements roadmap item **Priority 1** of #1020 (plus its folded-in small fix) across all three homotopy drivers: `HomotopySweep`, `ArcLengthContinuation`, and `SimpleHomotopySweep`.

## 1. `tracking_maxiters` (default 100; `nothing` disables)

An iteration cap merged into the inner solver's options for **interior tracking steps only** — never the sweep's `λspan[1]` anchor solve, never the final step landing on `λspan[2]` (in the cached `HomotopySweep` driver these are exempted via full-budget standalone rescues so the happy path keeps a single inner workspace; arclength's anchor/landing go through `_arclength_fixed_solve`, which stays uncapped), and never overriding an explicit user-passed `maxiters` (solve-level or problem-level kwargs win).

Motivation and measurement (failing fold sweep `u² + λ − 0.5 = 0` from `u0 = 0.7`, total residual calls; issue #1020 measured the same problem):

```
polyalg inner, uncapped        93049   (issue: 89349)
polyalg inner, default cap 100 11534
polyalg inner, cap 20           3257   (issue:  3294)
NewtonRaphson, uncapped        48132   (issue: 46173)
NewtonRaphson, cap 20           1050   (issue:  1093)
SimpleHomotopySweep uncapped   48126
SimpleHomotopySweep cap 20      1094
```

The generous default of 100 keeps currently-converging tracking steps converging (~8× cheaper rejections); 10–20 (OpenModelica `homMaxNewtonSteps`, MatCont `MaxCorrIters`, HC.jl's Newton cap) makes rejections ~30–45× cheaper.

## 2. Iteration-count band step controller (AUTO-07p `ADPTDS` style)

On success, the growth factor is scaled by the accepted step's corrector iteration count (`sol.stats.nsteps`, per-solve even under the cached driver; `stats === nothing` falls back to the classic behavior), layered **under** the existing gates (streak + `expand_quality` still veto growth): very cheap steps earn the full `expand_factor`, moderate ones earn half the growth, expensive ones hold (streak preserved so growth resumes when evidence recovers), and near-budget steps **proactively halve** the increment even on success — the earliest warning that the next full-size step would be rejected. `ArcLengthContinuation` gains the same bands alongside its bend-angle gate (it previously had no corrector-effort signal at all).

## 3. `maxsteps` guard for both sweeps (default 10000, matching arclength)

Caps total predictor-corrector attempts (accepted + bisection retries), returning `ReturnCode.MaxIters` with the last converged iterate. Previously the sweep's loop was bounded only by `span/min_dλ ≈ 7×10⁷` worst-case. Constructor-validated (≥ 1), as is `tracking_maxiters` (≥ 1 or `nothing`).

## Drive-by bug fix (pre-existing, exposed by this feature)

`init(prob, polyalg; maxiters)` **silently dropped `maxiters`**: the polyalg `__init` captured it without forwarding to the subcaches, and the specialized `solve!(::NonlinearSolvePolyAlgorithmCache)` runs each subcache to its own termination without consulting the polyalg-level field. The one-shot `__solve` path forwards `maxiters` to every subsolver; `__init` now matches. (This is why the cached sweep driver could not previously enforce any per-step budget with the default inner.)

## Verification (run locally, Julia 1.11.9; re-verified independently after the implementing agent's run)

- New `test/Core/homotopy_effort_tests__item{1,2}.jl`: eval-count reduction with generous margins over the measurements above; anchor/landing exemptions (problems whose anchor needs ~14 Newton iterations and whose landing needs > 1 succeed under caps of 5 and 1); user-kwarg precedence (solve-level and problem-level `maxiters` restore full-budget behavior); `maxsteps` guard returns `MaxIters` with finite last-converged iterate; constructor validation; effort-band non-regression on an easy path (growth still beats `expand_factor = 1`); arclength fold-rounding under caps of 100 and 20.
- Full battery re-run on the exact committed code: all 21 `homotopy_sweep` + 6 `arclength` + polyalg + both new effort files + polyalg cache/NoInit regression files + PolyAlgorithms basic — **277 assertions pass** (273 in the main runner + the polyalg-aliasing file run in its proper test env, 4/4).
- `SimpleHomotopySweep` AllocCheck test: **2/2** — the StaticArray sweep stays allocation-free with the new kwargs.
- Runic `--check` exit 0 on all seven touched files. No Project.toml version bumps.

## Coordination

Sibling PRs #1021 (arclength tangent/metric) and #1022 (sweep jac passthrough) touch the same files in disjoint regions; sequential rebases may be needed depending on merge order.

### DIFF
--- a/lib/NonlinearSolveBase/src/arclength.jl
+++ b/lib/NonlinearSolveBase/src/arclength.jl
@@ -2,7 +2,7 @@
     ArcLengthContinuation(; inner = nothing, initial_step_factor = 0.1,
         adaptive = true, min_ds = nothing, max_step_factor = 1.0,
         expand_factor = 2.0, expand_threshold = 2, max_angle = π / 6,
-        predictor = :secant, autodiff = nothing, tracking_maxiters = 100,
+        predictor = :secant, autodiff = nothing, tracking_maxiters = 20,
         maxsteps = 10000)
 
 Pseudo-arclength continuation solver for a `SciMLBase.HomotopyProblem`. Unlike
@@ -67,21 +67,14 @@ Keyword arguments:
   - `autodiff`: the automatic-differentiation backend (an `ADTypes.AbstractADType`) used
     to form the augmented Jacobian for the `:tangent` predictor; `nothing` (default)
     selects `AutoForwardDiff()`. Unused by the `:secant` predictor.
-  - `tracking_maxiters`: an iteration cap merged into the inner solver's options for the
-    augmented corrector solves only — never the `λspan[1]` anchor solve or the final
-    λ-fixed landing on `λspan[2]`, and never overriding an explicit user-passed
-    `maxiters` (a `maxiters` in the solve or problem kwargs always wins). A rejected
-    step is retried at half the arclength increment from a warm start, which is usually
-    far cheaper than letting the inner solver grind out its full default budget of 1000
-    iterations first (measured on a failing sweep of the analogous natural-parameter
-    driver: ~89000 residual calls at the default versus ~1000–3000 at a cap of 20). The
-    default of 100 is deliberately generous; lowering it to 10–20 (the range used by
-    OpenModelica, MatCont, and HomotopyContinuation.jl) makes rejections cheap on hard
-    paths. `nothing` disables the cap. On success the step growth is additionally scaled
-    by the corrector's iteration count (the AUTO-07p `ADPTDS` bands) alongside the
-    bend-angle gate: a near-free corrector earns the full `expand_factor`, moderate
-    effort milder growth, effort past a quarter of the budget holds `Δs`, and a success
-    that nearly exhausted the budget proactively halves it.
+  - `tracking_maxiters`: iteration cap for the augmented corrector solves (default 20,
+    the range used by OpenModelica, MatCont, and HomotopyContinuation.jl; `nothing`
+    disables). A rejected step retries at half the arclength increment from a warm
+    start, so failing fast is far cheaper than exhausting the inner solver's full
+    budget. Never applied to the anchor or final λ-fixed landing solves; an explicit
+    user-passed `maxiters` always wins. On success, step growth is additionally
+    scaled by the corrector's iteration count (AUTO-style bands) alongside the
+    bend-angle gate.
   - `maxsteps`: a hard cap on the total number of predictor-corrector attempts (including
     bisection retries). Required because the path is *not* monotone in `λ`, so a sweep
     that never reaches the target — a closed loop, or a branch escaping to infinity —
@@ -115,7 +108,7 @@ function ArcLengthContinuation(;
         inner = nothing, initial_step_factor = 0.1, adaptive = true,
         min_ds = nothing, max_step_factor = 1.0, expand_factor = 2.0,
         expand_threshold = 2, max_angle = π / 6, predictor = :secant,
-        autodiff = nothing, tracking_maxiters = 100, maxsteps = 10000
+        autodiff = nothing, tracking_maxiters = 20, maxsteps = 10000
     )
     if !(0 < initial_step_factor <= 1)
         throw(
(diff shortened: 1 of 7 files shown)


ChrisRackauckas-Claude:
Per review, pushed `3d4403de`:

- **`tracking_maxiters` default is now 20** (was 100), matching OpenModelica's `homMaxNewtonSteps`, MatCont's `MaxCorrIters`, and HomotopyContinuation.jl's Newton cap. The 100 default was hedging: a rejected step's warm-started retry at half the increment *is* the designed recovery, so there's no reason to let a struggling interior corrector grind toward a large budget first. Failing-fold cost now drops ~30× out of the box (uncapped ~93k residual calls → ~3.3k) instead of ~8×. Anchor and final landing solves remain uncapped and an explicit user `maxiters` still wins, so branch selection and target accuracy are unaffected.
- **Docstring bullets trimmed to the essentials** (purpose, default + provenance, exemptions, precedence) — the measurement tables live in the commit message and this PR instead.

Re-verified on the committed code: full battery (21 sweep + 6 arclength + polyalg + both effort files) **232/232**, `SimpleHomotopySweep` AllocCheck **2/2**, Runic clean. The eval-count tests are ratio-based, so they hold unchanged under the tighter default.

ChrisRackauckas-Claude:
Investigated `tracking_maxiters = 10` vs `20` per review (`9af8d835`). The investigation found and fixed a real controller bug before the comparison was even meaningful.

## The bug the benchmark exposed

The proactive-shrink band compared the accepted step's `stats.nsteps` against the per-solver budget — but a **polyalgorithm inner reports `nsteps` aggregated across its whole ladder** (Broyden 20 + Klement 20 + Newton 4 ≈ 44 > ¾·20). On systems large enough that the quasi-Newton members burn their budget before a Newton member succeeds cheaply, *every accepted step* read as "corrector struggling" and `dλ` halved every step. Measured on an n = 50 coupled system:

```
uncapped:  297 residual calls, ok
cap = 20:  119445 calls (400× blowup), ok only by crawling
cap = 10:  hits the maxsteps guard, FAILS
```

Neither the original tests nor the review battery caught this — no sweep test used n ≥ 50 (suite blind spot, now covered by the fix's rationale in the code comment).

**Fix:** the shrink signal is trusted only when `nit ≤ budget` — an over-budget count *proves* ladder aggregation, not corrector struggle — and an untrustworthy count may withhold growth but never shrink. Post-fix the same problem runs at 484–534 calls across caps of 10/15/20.

## 10 vs 20 on the fixed controller

| case | cap=10 | cap=20 | uncapped |
|---|---|---|---|
| fail-fold, polyalg inner | 2711 | 3257 | 93049 |
| fail-fold, NewtonRaphson | 779 | 1050 | 48132 |
| fail-fold, SimpleHomotopySweep | 790 | 1094 | 48126 |
| sharp-turn trap, polyalg | 524 | 728 | 5165 |
| trap, SimpleHomotopySweep | 231 | 351 | 2873 |
| n=50 coupled, polyalg | 484 | 534 | 297 |
| easy / steep / arclength S-fold (sec+tan) ×3 | identical | identical | identical |

**10 wins or ties everywhere** — 17–34% cheaper on failing/hard paths, identical on easy ones. Risk is asymmetric in its favor: a too-tight cap costs one warm-started bisected retry, a too-loose one taxes every rejection. It also matches field practice (MatCont `MaxCorrIters = 10`, HomotopyContinuation.jl's Newton cap 11, AUTO `ITNW = 5`; OpenModelica's 20 is the high outlier). **Default is now 10** in all three drivers.

Re-verified on the committed code: full battery **232/232**, `SimpleHomotopySweep` AllocCheck **2/2**, Runic clean.

PR #1036, opened 2026-07-11
Inner-algorithm retention for the homotopy continuation drivers (sticky polyalg rung + lazy reinit!)

ChrisRackauckas-Claude:
Part of #1020; addresses the default-inner overhead found in the post-#1029 benchmarks.

## Problem

The homotopy continuation drivers (`HomotopySweep`, the `ArcLengthContinuation` corrector) solve a sequence of warm-started problems through one persistent inner cache (`init`/`reinit!`/`solve!`). With the default inner (the `FastShortcutNonlinearPolyalg` ladder), every `reinit!` restarted the ladder at `start_index` and eagerly reinitialized **all** subcaches. That cost the drivers twice:

1. **Ladder re-failing**: for `n > 25` (`start_index = 1`) every warm step re-failed Broyden/Klement before Newton succeeded again.
2. **Eager `reinit!` cost**: every subcache `reinit!` evaluates the residual at the new `u0` (`Utils.reinit_common!`), so each continuation step paid one residual call *per ladder rung* (6 for the default ladder) versus 1 for a bare `NewtonRaphson()` inner. For small systems the default ladder already starts at Newton (`start_index = 3` for `u0_len ≤ 25`), so this eager-reinit cost was the *entire* measured 2x overhead there — not ladder re-failing.

Result on the post-#1029 benchmark set: the default inner cost ~1.6–2x a plain `NewtonRaphson()` inner everywhere.

### `reinit!(cache; retain_best = true)` on the polyalgorithm cache (opt-in)

- **Sticky start**: the polyalg cache records which subalgorithm produced the last success (`cache.best`, now also set on the `solve!` path). A `retain_best = true` `reinit!` starts the next `solve!` from that index instead of `start_index`. Plain `reinit!` (the default) is the exact status-quo full restart, so no polyalg user outside the homotopy drivers sees any behavior change.
- **Upward escalation preserved**: if the sticky rung fails, the ladder continues up exactly as before.
- **Wrap-around, floored at `start_index`**: when the retained start was past `start_index` and the top of the ladder fails too, the ladder wraps around to the skipped cheaper rungs before falling through to the lowest-residual selection — retention never tries *fewer* rungs than a full ladder run, and never tries rungs the algorithm's own `start_index` excludes.
- **Lazy subcache reinitialization**: a retaining `reinit!` reinitializes only the starting subcache; escalation freshens each further subcache at the moment it is attempted. This reduces the per-step `reinit!` cost from one residual call per rung to one per rung *attempted* (usually just the retained one) — this is what closes the gap on small systems.
- **Periodic re-probe**: when the retained rung sits above `start_index`, every 8th retained `reinit!` starts one solve from `start_index` again, so a cheap rung that failed once transiently is rediscovered instead of locked out forever. Bounded cost: at most one status-quo-style ladder step per interval.

### Drivers arm it on tracking steps only

`HomotopySweep` and the `ArcLengthContinuation` corrector call `reinit_retaining!` (retaining for a polyalg cache, plain `reinit!` otherwise) on their warm-started tracking `reinit!`s. The cold anchor solve at `λspan[1]` (and arclength's λ-fixed anchor/landing solves) still runs the full ladder — that is where the winner is discovered. `SimpleHomotopySweep` is untouched: it deliberately reconstructs a fresh stack-allocated `solve` per step (no persistent inner cache), and its default inner is `SimpleNewtonRaphson`, not a polyalgorithm, so there is nothing to retain there.

### Wrap-around decision (measured)

Three variants were measured on the benchmark set:

- **Wrap to rung 1**: made the *failure* path more expensive than the status quo (P2 sweep cost-to-fail 2944 vs 2506 on master), because small systems (`start_index = 3`) suddenly attempted Broyden/Klement — rungs the status-quo ladder never runs — on every bisection retry at the fold.
- **No wrap (report failure)**: cheapest failures, but a wrapped rung genuinely can rescue a stagnating retained rung, and a driver-side corrector failure feeds bisection (more retries), so cheap-but-avoidable failures are not free either.
- **Wrap floored at `start_index`** (chosen): when the retained rung *is* the ladder's `start_index` (the common steady state), the failure path is exactly the status-quo ladder — the wrap only engages in the rare retained-past-start case, keeping the robustness of trying every rung of a full ladder run. Measured: P2 sweep cost-to-fail drops to 2344 (below the 2506 status quo, thanks to the lazy reinit), P2 arclength 291 → 267.

## Benchmarks (residual calls via `Ref` counter, measured on this branch at its final state)

P1 scalar monotone `(1-λ)(u-4) + λ(u²-4)`; P2 cubic S-curve fold `u³-3u-(-3+6λ)`; P3 n=50 coupled cubic iip; P4 hard ramp `u³-1-7λ` from `u0=100`:

| problem | default (master) | default + retention | `NewtonRaphson()` | Newton-first static `(NR, TR)` |
|---|---|---|---|---|
| P1 sweep | 150 | **94** | 86 | 88 |
| P2 sweep (cost-to-fail) | 2506 | **2344** | 746 | 1294 |
| P2 arclength (full solve) | 381 | **267** | 188 | 210 |
| P3 sweep n=50 | 2295 | **1577** | 1400 | 1451 |
| P4 sweep | 246 | **204** | 123 | 146 |

- P1 lands within 9% and P3 within 13% of the `NewtonRaphson()` inner (target was 10–20%), while staying fully adaptive; the static Newton-first ladder is matched or beaten on the success-path problems.
- P4's remaining gap is **entirely the anchor**: anchor-only cost (zero-width span) is 140 (default) vs 59 (Newton), and 204−123 = 140−59 = 81 — interior tracking overhead with retention is zero. The cold anchor deliberately runs the full ladder (under the tracking cap, then exempt-rerun), which is pre-existing behavior this PR intentionally does not change.
- P2's sweep row is a *cost-to-fail* (the sweep cannot pass the fold; failure hands off to arclength). A polyalgorithm failure intrinsically runs its whole ladder, so Newton-parity is unreachable there; retention + lazy reinit still brings it below master. Same effect explains most of P2-arclength's remaining gap (fold-region corrector failures).

### Where a quasi-Newton rung genuinely wins (honest findings)

Two constructed n=100 problems (`start_index = 1`):

| problem | default (master) | default + retention | `NewtonRaphson()` | Klement-only |
|---|---|---|---|---|
| P5 mild `u + 0.1λ sin(u) + 0.25u₊ − 1` | 2882 | **1556** | 1229 | — |
| P6 near-linear `u − 1 + 0.05λ tanh(u) + 0.05λ u₊` | **918** | 1806 | 1229 | 1088 |

- On P6 a `Klement()` inner (1088) genuinely beats `NewtonRaphson()` (1229) — quasi-Newton *can* win steady-state.
- On that same P6, **master's full-restart default (918) beats everything**, including retention (1806) and the static Newton-first ladder (1330). Mechanism: the status quo re-runs Broyden first every step, which succeeds at moderate effort; that effort keeps the step controller's dλ small-but-cheap. Once retention escalates to Newton (after one transient quasi-Newton failure), Newton's near-free correctors grow dλ, and at the larger step size the cheap rungs *fail* their periodic re-probes — so sticky-best stays locked on Newton's more-expensive-per-call equilibrium. The re-probe (kept: bounded cost, and it does rescue lock-in when the cheap rung succeeds at the prevailing step size) cannot recover this case; recovering it would need cost-aware retention (retain the cheapest recent winner, not the last winner), which interacts with the step controller and is deliberately left as future work.
- On P5 (Broyden mostly failing warm) retention wins big (2882 → 1556). The two problems bracket the trade: retention wins when the cheap rungs mostly fail, ties on small systems, and can lose to full restart on problems where a cheap rung succeeds-with-effort every step. Given P1–P5 and that the drivers' benchmark set motivating this PR all improve, the trade looks right for the homotopy drivers' default.

## Stage-level retention (follow-up)

Carrying the winning inner rung across the sweep → arclength handoff (stage-level retention in `HomotopyPolyAlgorithm`) is deliberately **not** part of this PR: #1030 is churning `homotopy_polyalg.jl` (warm handoff), so that follow-up should land after #1030 merges.

## Tests

- `test/Core/polyalg_retention_tests__item1.jl`: sticky start picks the winning rung on the next solve (asserted via subsolver `nsteps` and exact residual-call counts of the lazy reinit), escalation past a failing sticky rung, wrap-around rescuing via a skipped cheaper rung, wrap-around refusing to go below `start_index`, the periodic re-probe rediscovering a cheaper rung, and `retain_best = false` ⇒ exact status-quo behavior (eager reinit of every subcache, ladder restart at `start_index`).
- `test/Core/homotopy_retention_tests__item1.jl`: residual-call reduction on the benchmark problems with generous margins (each margin trips on master's numbers), anchor still runs the full ladder (an anchor that *requires* escalation), and a difficulty-jump path where the retained Jacobian-based rung hard-fails mid-continuation and escalation must rescue the sweep.

All existing homotopy sweep/arclength/effort/tolerance/jac tests, the homotopy polyalg tests, the core polyalg/reinit!/cache/issue tests, and NonlinearSolveBase's polyalg solution-type tests pass locally (output in the PR conversation).

### DIFF
--- a/lib/NonlinearSolveBase/src/arclength.jl
+++ b/lib/NonlinearSolveBase/src/arclength.jl
@@ -76,6 +76,11 @@ Keyword arguments:
 
   - `inner`: the inner nonlinear algorithm used for both the initial on-curve correction
     and the augmented corrector; `nothing` selects NonlinearSolve's default polyalgorithm.
+    A polyalgorithm inner runs the augmented corrector with best-subalgorithm retention
+    (see [`NonlinearSolvePolyAlgorithm`](@ref)): after the first corrector solve
+    discovers the winning subalgorithm, each warm-started corrector resumes from it
+    instead of re-running the ladder from its start, escalating only when it fails. The
+    λ-fixed anchor and landing solves always run the full ladder.
   - `initial_step_factor`: the initial arclength step `Δs` as a fraction of the `λspan`
     width. Must be in `(0, 1]`.
   - `adaptive`: when `true` (default), a corrector failure halves `Δs` and retries from
@@ -810,7 +815,11 @@ function CommonSolve.solve(
 
         @. xpred = xcur + ds * τ
         aug.ds = Tx(ds)
-        SciMLBase.reinit!(corr_cache, xpred)
+        # Retaining reinit!: a polyalgorithm inner resumes from the subalgorithm that
+        # won the previous corrector solve (the first corrector runs the full ladder
+        # and discovers the winner) instead of re-failing the cheaper ladder members
+        # on every warm-started step.
+        reinit_retaining!(corr_cache, xpred)
         last_sol = CommonSolve.solve!(corr_cache)
 
         if SciMLBase.successful_retcode(last_sol)
--- a/lib/NonlinearSolveBase/src/homotopy_sweep.jl
+++ b/lib/NonlinearSolveBase/src/homotopy_sweep.jl
@@ -17,7 +17,13 @@ The inner solver is initialized once and re-driven each step through the
 `init`/`reinit!`/`solve!` cache interface, so the continuation loop reuses the inner
 solver's workspace (Jacobian buffers, linear-solver storage) instead of reconstructing
 the solver every step; the sweep's own per-step state lives in a fixed set of
-preallocated buffers.
+preallocated buffers. When the inner solver is a polyalgorithm (as the default is),
+the warm-started tracking steps additionally arm best-subalgorithm retention on that
+cache (see [`NonlinearSolvePolyAlgorithm`](@ref)): each step resumes from the
+subalgorithm that produced the previous step's success instead of re-running the
+polyalgorithm ladder from its start, escalating (and eventually wrapping around) only
+when the retained subalgorithm fails. The cold anchor solve at `λspan[1]` always runs
+the full ladder — that is where the winning subalgorithm is discovered.
 
 Optional derivative fields of the problem's `NonlinearFunction` (which
 [`SciMLBase.HomotopyProblem`](@ref) requires to follow the same λ-extended argument
@@ -526,7 +532,10 @@ function _homotopy_sweep_solve(
             _sweep_warmstart!(guess, u)
         end
         fixλ.λ = next_λ
-        SciMLBase.reinit!(cache, guess)
+        # Retaining reinit!: a polyalgorithm inner resumes from the subalgorithm that
+        # won the previous solve (the anchor's full-ladder run discovers the winner)
+        # instead of re-failing the cheaper ladder members on every warm-started step.
+        reinit_retaining!(cache, guess)
         last_sol = CommonSolve.solve!(cache)
 
         if next_λ == λend
--- a/lib/NonlinearSolveBase/src/polyalg.jl
+++ b/lib/NonlinearSolveBase/src/polyalg.jl
@@ -26,6 +26,41 @@ using NonlinearSolve
 
 alg = NonlinearSolvePolyAlgorithm((NewtonRaphson(), Broyden()))
 ```
+
+### Best-subalgorithm retention (`reinit!(cache; retain_best = true)`)
+
+When the polyalgorithm's cache is reused across a sequence of warm-started solves
+(`init`/`reinit!`/`solve!`), each `reinit!` by default restarts the ladder at
+`start_index`, so every solve re-fails the same cheap subalgorithms before reaching
+the one that works. Passing `retain_best = true` to `reinit!` makes the next `solve!`
+start from the subalgorithm that produced the most recent success instead. Escalation
+is preserved: if that subalgorithm fails, the ladder continues upward as usual, and
+once the last subalgorithm fails it wraps around to the subalgorithms that were
+skipped at the start (they occasionally succeed where the retained one stagnates), so
+retention never tries fewer subalgorithms than a full ladder run. When no success has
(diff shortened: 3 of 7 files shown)


ChrisRackauckas-Claude:
Local verification output (scratch env with all local packages `Pkg.develop`ed by path):

**New tests** (`test/Core/polyalg_retention_tests__item1.jl`, `test/Core/homotopy_retention_tests__item1.jl`):
```
Test Summary:     | Pass  Total  Time
polyalg retention |   32     32  7.4s
Test Summary:      | Pass  Total   Time
homotopy retention |   16     16  41.7s
```
(The polyalg retention file has since grown the re-probe testset; it is included in the full run below.)

**Full branch check** — every existing `test/Core/homotopy*`, `test/Core/arclength*` file, the homotopy polyalg tests, the core polyalg/cache/reinit! tests (`core_tests__item{3,5,13}`, `issue_tests__item{1,2,3}`, `reinit_tests__item1`, `PolyAlgorithms/core_tests__item2`), NonlinearSolveBase's `polyalg_solution_type.jl`, plus both new retention files, all run against the final branch state:
```
Test Summary: | Pass  Total      Time
branch check  |  565    565  11m18.4s
```

Benchmark reproduction scripts and raw numbers are in the PR body; the "master" column was measured on the same machine by disabling the drivers' retention hook (method override of `reinit_retaining!`), and reproduces the motivating post-#1029 numbers exactly (150 / 2506 / 2295 / 246).

PR #1041, opened 2026-07-11
Stop continuation drivers boxing solution field reads (Any-typed last_sol)

ChrisRackauckas-Claude:

Stacks on #1039 (`maybe-pinv-workspace`); its commits appear here until #1039 merges, then this rebases to just the top commit.

## Root cause

Profiling attributed ~384 B/step of a HomotopySweep continuation to SciMLBase's solution `getproperty`. It turned out **getproperty is blameless** — it's already `sym === :ps ? … : getfield(A, sym)`, non-allocating (0 B) for a concretely-typed `NonlinearSolution`. The real cause: the sweep and arclength drivers assign `last_sol` from two sources — the inner `solve!(cache)` (concretely inferred) and a fresh `solve(inner_prob, inner)` in the anchor/landing/exempt path (inference on the inner polyalgorithm gives up → `Any`). The union collapses `last_sol` to `Any` (`@code_warntype`: `last_sol::ANY`), so every inline `last_sol.u`/`.resid`/`.retcode`/`.stats` read dynamically dispatches and **heap-boxes the field** (~560 B/read). A plain `getfield` swap does NOT help — `getfield(::Any, sym)` boxes identically.

## Fix

Route the solution field reads in both drivers through typed-argument accessors (`_sol_u`/`_sol_resid`/`_sol_retcode`/`_sol_nsteps`) so Julia specializes each on the concrete runtime type (provably identical between the two solve sites) and the read compiles to a non-boxing `getfield`. No SciMLBase change — getproperty is not at fault.

## Measured (5000-step HomotopySweep, local)
- Profile.Allocs bytes on solution `getproperty`: **58528 → 0**
- warmed `@allocated solve(prob, alg)`: **104816 → 46288** (more than halved)

## Tests
- New `test/Core/homotopy_sweep_tests__item22.jl`: 8/8, deterministic across 3 runs (accessor loop-delta = 0 alloc, `@inferred` stability, `sol.u`/`successful_retcode` still work, sweep converges u ≈ 2.0)
- Existing homotopy sweep items 2/3/7/11/17/19/20/21 + arclength item2: all pass


ChrisRackauckas-Claude:
Superseded by #1044, which fixes the root cause instead of the symptom.

Per @ChrisRackauckas's diagnosis, the instability is the `original` field. The `Any`-typed `last_sol` this PR worked around with typed-argument `_sol_*` accessors originates from two things: the drivers build their inner λ-fixed `NonlinearProblem` with the default `AutoSpecialize` `NonlinearFunction`, whose `FunctionWrappersWrapper` prototype inference gives up on the `FixLambda` wrapper-of-a-wrapper (so `solve(inner_prob, inner)` in `_sweep_exempt_solve` / `_arclength_fixed_solve` infers `Any`), and the drivers then store that `Any`-inferred solution into their own returned solution via `build_solution(...; original = last_sol)`, pinning the returned solution's `original` type-slot to `Any`.

#1044 fixes both at the source — `FullSpecialize` on the inner `NonlinearFunction` so the inner solve infers, and a `store_original::Val` gate (default `Val(false)`, mirroring `NonlinearSolvePolyAlgorithm`/#878) that drops `original` and pins its slot to `Nothing`. With the instability fixed at the source, the `_sol_*` accessor barriers here are no longer necessary.

Leaving this open for @ChrisRackauckas to decide.

PR #1066, opened 2026-07-13
Restore GPU Broyden inverse initialization

ChrisRackauckas-Claude:
Ignore until reviewed by @ChrisRackauckas.

## Summary

- avoid constructing the cached `A * X = I` inverse workspace for array backends that do not support fast scalar indexing
- retain sparse-matrix dispatch before falling back to the backend's native `pinv`
- cover the fallback with a CPU backend wrapper and add explicit CUDA Broyden residual checks for both in-place and out-of-place problems

## Root cause

PR #1039 changed QuasiNewton inverse initialization from the backend-native `pinv(A)` path to a cached LinearSolve factorization with a matrix identity right-hand side. On `CuArray`, the cuSOLVER QR matrix-RHS path throws `DimensionMismatch: array could not be broadcast to match destination`, which breaks both Broyden variants in the CUDA job.

`linsolve_workspace` now reports no generic cached workspace for backends where `ArrayInterface.fast_scalar_indexing` is false. `linsolve_identity!!` first calls the dispatched workspace constructor, so the existing sparse specialization still builds its dense workspace; it uses `pinv` only when dispatch returns no workspace.

## Adjacent CI boundary

- parent `d616cd994` passed the root CUDA suite 20/20: https://github.com/SciML/NonlinearSolve.jl/actions/runs/29179879552/job/86615528686
- merge `6a7687f542` failed the root CUDA suite 18/20 with the same two Broyden `DimensionMismatch` errors: https://github.com/SciML/NonlinearSolve.jl/actions/runs/29180426325/job/86616954929
- PR #1039 contained the sole feature commit `9f4a985acea68aef65915dcf261cb2105c814709`
- the current failure is visible in https://github.com/SciML/NonlinearSolve.jl/actions/runs/29281131540/job/86923811735

The adjacent jobs also resolved LinearSolve 4.3 and 5.0 respectively, so this is an exact failing-state boundary rather than a claim that the source diff was tested locally against an otherwise frozen GPU environment. The current failure uses LinearSolve 5.0.

## Local validation

- `GROUP=Core NONLINEARSOLVE_TEST_GROUP=Core julia +1.12 --project=lib/NonlinearSolveBase -e 'using Pkg; Pkg.test()'` (passed; workspace 52/52)
- `GROUP=Core NONLINEARSOLVE_TEST_GROUP=Core julia +1.12 --project=lib/NonlinearSolveQuasiNewton -e 'using Pkg; Pkg.test()'` (passed; Broyden 594/594, Klement 216/216, LimitedMemoryBroyden 99/99, and associated interface/termination tests)
- direct LinearSolve 5.0 Broyden regression in `test/gpu` environment (passed 2/2 for in-place/out-of-place residuals)
- `julia +1.12 --project=../runic-format-env -m Runic --check .` (passed)

This machine has no functional CUDA driver (`CUDA.functional() == false`), so the CuArray regression is intentionally asserted in the CUDA CI test rather than claimed as locally executed.

### DIFF
--- a/lib/NonlinearSolveBase/src/utils.jl
+++ b/lib/NonlinearSolveBase/src/utils.jl
@@ -235,6 +235,9 @@ linsolve_workspace(A::Diagonal) = nothing, A
 linsolve_workspace(A::SMatrix) = nothing, A
 function linsolve_workspace(A::AbstractMatrix)
     LinearAlgebra.checksquare(A)
+    # Backends without fast scalar indexing may not implement factorization solves with
+    # a matrix right-hand side. Leave those arrays on their native `pinv` path.
+    ArrayInterface.fast_scalar_indexing(A) || return nothing, A
     # the linear solve promotes e.g. integer eltypes; the buffers must hold the promoted
     # values. The `size` form of `similar` gives a dense buffer for structured input
     # (e.g. Tridiagonal), which the generically-dense inverse and identity RHS need.
@@ -283,15 +286,20 @@ end
 function linsolve_identity!!(workspace, A::AbstractMatrix)
     LinearAlgebra.checksquare(A)
     # a `nothing` workspace (the preinverted-init path in NonlinearSolveQuasiNewton)
-    # builds a transient linear-solve cache; it allocates, but runs at most once per solve
-    ws = workspace === nothing ? first(linsolve_workspace(A)) : workspace
+    # builds a transient linear-solve cache; it allocates, but runs at most once per solve.
+    # Sparse dispatch gets a chance to construct its dense workspace here, while backend
+    # arrays that cannot use the cached matrix-RHS solve stay on their native `pinv`.
+    if workspace === nothing
+        workspace, workspace_A = linsolve_workspace(A)
+        workspace === nothing && return pinv(workspace_A)
+    end
     # the previous solve may have consumed the RHS buffer, so refill it with I. The
     # lincache call copies A into its internal buffer before overwriting the solution
     # buffer, so A is never mutated and passing the previously returned inverse as A is
     # safe. On singular A the default algorithm's pivoted-QR rescue returns a finite
     # least-squares generalized inverse (not the SVD `pinv`).
-    make_identity!!(ws.rhs, true)
-    return ws.lincache(; A, b = ws.rhs).u
+    make_identity!!(workspace.rhs, true)
+    return workspace.lincache(; A, b = workspace.rhs).u
 end
 
 function initial_jacobian_scaling_alpha(α, u, fu, ::Any)
--- a/lib/NonlinearSolveBase/test/linsolve_workspace.jl
+++ b/lib/NonlinearSolveBase/test/linsolve_workspace.jl
@@ -1,7 +1,17 @@
 using NonlinearSolveBase: NonlinearSolveBase, Utils
+using ArrayInterface
 using LinearSolve  # linsolve_identity!! routes matrices through the LinearSolve default
 using LinearAlgebra
 using Test
+
+struct NoFastScalarMatrix{T} <: AbstractMatrix{T}
+    data::Matrix{T}
+end
+
+Base.size(A::NoFastScalarMatrix) = size(A.data)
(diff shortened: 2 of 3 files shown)


ChrisRackauckas-Claude:
Investigation scratchpad:

1. Reproduced the current failure from the complete root CUDA job log: the in-place and out-of-place Broyden cases both fail in cuSOLVER's QR matrix-RHS `ldiv!` with a `DimensionMismatch`.
2. Located the adjacent CI boundary: `d616cd994` passes CUDA 20/20, while `6a7687f542` fails 18/20 with the same stack; the sole #1039 feature commit is `9f4a985acea`.
3. Kept sparse behavior intact by routing through dispatched `linsolve_workspace(A)` before deciding whether no workspace means native `pinv`.
4. Added a backend-trait regression in NonlinearSolveBase and explicit in-place/out-of-place CUDA Broyden residual assertions.
5. Ran the NonlinearSolveBase Core suite, the full NonlinearSolveQuasiNewton Core suite, the direct two-case Broyden regression under the CUDA environment's LinearSolve 5.0 graph, and repository-wide Runic. All locally runnable checks passed.

Implementation commit: `3caef159be775d0c05508cfc9b725b166fc12e15`.

Local CUDA execution is unavailable because this host's CUDA driver is not functional. The PR CUDA job is the required validation of the CuArray path.

ChrisRackauckas-Claude:
CI baseline classification: the two immediate downgrade failures reproduce on the exact upstream base commit `41681c5e4` and are unrelated to this patch.

- upstream root Downgrade failure: https://github.com/SciML/NonlinearSolve.jl/actions/runs/29279140544/job/86915711319
- upstream SimpleNonlinearSolve downgrade failure: https://github.com/SciML/NonlinearSolve.jl/actions/runs/29279140506/job/86915732833

In both the base and PR runs, the downgrade harness explicitly requires SciMLOperators `1.22.0`, while the checked-out NonlinearSolveBase/NonlinearSolveQuasiNewton graph requires SciMLOperators `1.24`, leaving no resolvable version. The root-job resolver trace is exact; the SimpleNonlinearSolve trace reaches the same `1.22.0` versus `1.24` conflict through a slightly different fixed in-tree dependency chain.

The root CUDA job remains the gating check for this PR's behavior change.

ChrisRackauckas-Claude:
Scheduled one-hour CI follow-up at 2026-07-13 19:39 EDT: 62 checks pass, 3 matrix placeholders are skipped, 6 completed failures are the already-classified clean-master resolver/ExplicitImports and generalized-Rosenbrock failures, and 22 checks remain queued. I inspected the newly completed current/prerelease Core logs; both fail only Generalized Rosenbrock alg #4. The decisive root CUDA job is still queued without a runner, so this run does not yet provide GPU verification of the patch. Local CUDA execution remains unavailable on this machine; the GPU result will be reported only after that job actually runs.

ChrisRackauckas-Claude:
GPU verification completed on the self-hosted V100 runner. The root CUDA group passed 22/22 in 57.3 seconds, versus the prior 20-test baseline; the two added tests are the in-place and out-of-place Broyden inverse-initialization regressions. The subsequent termination-allocation checks passed 20/20, and the harness ended with NonlinearSolve tests passed. This is the first actual CUDA execution result for the patch; local GPU execution was unavailable.

ChrisRackauckas-Claude:
Rebased the single focused commit onto current upstream `60736b5b6` after fetching and auditing the exact prior tip `3caef159b`, its authorship, same-account comments, and absence of reviews. `git range-diff` reports the patch unchanged (`3caef159b = 0060a0227`).

Fresh Julia 1.12.6 validation on the rebased commit: NonlinearSolveBase `GROUP=Core` passed, including `linsolve_identity!! workspace` 52/52, routing 12/12, operator dispatch 1/1, and dense-LU allocations 14/14. Repository Runic and `git diff --check` passed. The previously recorded V100 root CUDA run remains 22/22, including both added Broyden regressions.

The branch update used `--force-with-lease` pinned to the audited prior SHA; no contributor commits were overwritten.

PR #1067, opened 2026-07-14
Overload `init_similar_array!!` for sparse arrays

hexaeder:
This PR has two changes:
1. It drops the `can_setindex(x)` guard in `init_similar_array!!` (used by `safe_similar`) which would -- other than the name implies -- NOT return a zeroed out similar array. This guard was introduced way back for sanitizing BigFloat arrays, the function then slowly became a general `safe_similar`. I think in this context it is better to error loud rather than returning non-initialized similars. This would, for example, return uninitialized jacobian buffers after the (truthfull) https://github.com/JuliaArrays/ArrayInterface.jl/pull/495#event-27947734569
2. `fill!` does work on CPU sparse arrays but not on GPU sparse arrays. So we overload `init_simliar_array!!` for AbstractSparseArrays by filling on `nonzeros`. 

## Checklist

### DIFF
--- a/lib/NonlinearSolveBase/ext/NonlinearSolveBaseSparseArraysExt.jl
+++ b/lib/NonlinearSolveBase/ext/NonlinearSolveBaseSparseArraysExt.jl
@@ -1,7 +1,8 @@
 module NonlinearSolveBaseSparseArraysExt
 
 using ArrayInterface: ArrayInterface
-using SparseArrays: AbstractSparseMatrix, AbstractSparseMatrixCSC, nonzeros, sparse
+using SparseArrays: AbstractSparseArray, AbstractSparseMatrix, AbstractSparseMatrixCSC,
+    nonzeros, sparse
 
 using NonlinearSolveBase: NonlinearSolveBase, Utils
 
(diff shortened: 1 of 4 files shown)


lib/NonlinearSolveBase/src/utils.jl
```diff
@@ -123,7 +123,7 @@ safe_similar(x::AbstractSciMLOperator, ::Type) = x
 init_similar_array!!(x) = x
 
 function init_similar_array!!(x::AbstractArray{<:T}) where {T <: Number}
-    ArrayInterface.can_setindex(x) && fill!(x, T(0))
```
ChrisRackauckas, line 126: Not all arrays can setindex though, this will fail on StaticArrays?

hexaeder: Isn't it surprising if a mutating init call becomes a no-op on non-mutable inputs? After all, the argument is not "initialized" after the call. But yes, can drop, i am more interested in the AbstractSparseArray overload for sparse GPU arrays. Does the same argument apply there to? Would we expect non-mutable sparse arrays?

ChrisRackauckas: It's a `!!` function, so it's either mutating or returns, and the return matches the structure. 

Why is this necessary for the GPUs to work? They can setindex?

hexaeder: dropped the wrong removal of `can_setindex`. The sparse path is gueaded by ArrayInterface.ismutable(nonzeros(sprs)) because CUSPARSE implement no setindex at all.

ChrisRackauckas (approved):

PR #1060, opened 2026-07-13
Raise SciMLOperators public API floors and resolve cache ambiguity

ChrisRackauckas-Claude:
Ignore this PR until it has been reviewed by @ChrisRackauckas.

## Summary

- require SciMLOperators 1.24, the first registered release that declares `AbstractSciMLOperator` public, in every subpackage that directly imports it
- remove the now-obsolete ExplicitImports allowlist entries from NonlinearSolveBase, NonlinearSolveQuasiNewton, and SciMLJacobianOperators
- add the missing `Nothing` specialization for `JacobianCache{<:AbstractSciMLOperator}`, resolving its intersection with the generic deprecated call method
- cover the operator-cache dispatch with an in-place `MatrixOperator` regression test
- patch-bump NonlinearSolveBase to 2.34.3, NonlinearSolveFirstOrder to 2.2.1, NonlinearSolveQuasiNewton to 1.14.1, and SciMLJacobianOperators to 0.1.16

SciMLOperators commit `a28a25ac5f81e3922351fecc2535e2b264ec8f2c` made `AbstractSciMLOperator` public. Registry and tree-history verification show that release 1.23.0 predates that commit and release 1.24.0 contains it. Establishing that public API floor permits the operator-specific disambiguation without relying on dependency internals.

The minimum-compatible sublibrary environments also depend on [julia-actions/julia-downgrade-compat#55](https://github.com/julia-actions/julia-downgrade-compat/pull/55), stacked on [#54](https://github.com/julia-actions/julia-downgrade-compat/pull/54), so direct local runtime dependencies are retained and resolved while producing the downgraded manifest.

## Process

1. Reproduced NonlinearSolveBase QA on exact upstream `579761dd26ccd522fae20b29cef8e9ed945569c5`.
2. Traced the owner-package public declaration and mapped it to the first registered release.
3. Located every direct `AbstractSciMLOperator` consumer and raised its SciMLOperators floor.
4. Removed only the corresponding QA exceptions and patch-bumped the affected subpackages.
5. Added the operator-cache method required to make the dispatch lattice unambiguous and exercised it directly.
6. Exercised minimum-compatible and latest-compatible environments locally.

## Local verification

- Final combined head, Julia 1.10.11 minimum-compatible NonlinearSolveBase `Core`: 112/112 passed, including the operator-cache regression 1/1.
- Julia 1.10 minimum-compatible downgrade (`alldeps`), with SciMLOperators 1.24.0:
  - SciMLJacobianOperators `Core` passed.
[...]

PR #1057, opened 2026-07-13
Preserve triangular solves for Broyden Jacobian inversion

ChrisRackauckas-Claude:
> [!IMPORTANT]
> Ignore this PR until it has been reviewed by @ChrisRackauckas.

## Summary

- Preserve triangular solve semantics when initializing an inverse Jacobian from a nonsingular CPU strided triangular matrix.
- Preserve an aliased prior inverse before refilling the reusable identity RHS, so direct resets and nested reinversion remain correct.
- Keep general matrices and singular triangular matrices on LinearSolve's existing default path, including its pivoted-QR rescue.
- Add direct workspace/reset coverage and the exact generalized-Rosenbrock bad-Broyden regression.
- Preserve all existing expected-broken coverage; this does not adopt #1069's broken-test change.

## Root cause

The current master Core suite at `247a8bfcb` fails generalized Rosenbrock with:

    Broyden(init_jacobian = Val(:true_jacobian), update_rule = Val(:bad_broyden))

A clean adjacent-commit reproduction and `git bisect` identified `6a7687f542cde52b4bfeef5b473c6156c8d453a1` (#1039) as the first bad commit:

- `d616cd994` (its parent): `Success`, 17 iterations, residual 0.0
- `6a7687f542` and affected master: `Unstable`, 24 iterations, residual 4.4

The persistent workspace itself was not the cause. The true Jacobian is dense lower triangular, and the generic pivoted-LU solve introduced different roundoff than the previous structure-preserving triangular inverse. Bad Broyden is sensitive to that perturbation. Solving the identity RHS through `LowerTriangular` restores the prior trajectory without reinstating matrix-sized allocations.

The fast path is restricted to `StridedMatrix` inputs with a nonzero diagonal. GPU/non-strided inputs, general matrices, and singular triangular matrices retain the existing LinearSolve route. The workspace-owned coefficient buffer handles the case where a prior triangular result aliases the identity RHS during reset.

This implementation uses exported `LinearAlgebra` API only: `diagind`, `istriu`, `istril`, `UpperTriangular`, `LowerTriangular`, and `ldiv!`. It does not add, remove, or change any public NonlinearSolve API.

## Verification

The branch was safely rebased onto current master at `247a8bfcb`, after re-fetching the remote branch and verifying that its two commits still contained no intervening contributor work. The following commands were then run locally on Julia 1.10.11 with current compatible dependencies, including LinearSolve 5.0.0:

[...]

Issue #1056, opened 2026-07-13
Brown bad-Broyden broken test is environment-sensitive


ChrisRackauckas-Claude:
## Summary

The root Core group can fail on unmodified master because the Brown almost-linear problem unexpectedly passes its existing broken assertion for Broyden with a true initial Jacobian and the bad-Broyden update.

The same NonlinearSolve source commit produces both outcomes across environments, so this does not have a source-commit bisect boundary. The broken expectation also predates the test-suite migration in 75178936d7.

## Local reproduction

Source: b491a4e8d787 (master)

Environment:
- Julia 1.10.11
- Linux x86_64, AMD EPYC 7502
- SciMLOperators 1.24.3

Command:

    NONLINEARSOLVE_TEST_GROUP=Core julia +1.10 --project=. -e 'using Pkg; Pkg.test()'

Output:

    8: Brown almost linear function | alg #4: Error During Test
     Unexpected Pass
     Expression: norm(res, Inf) <= epsilon

Algorithm 4 is:

    Broyden(init_jacobian = Val(:true_jacobian), update_rule = Val(:bad_broyden))

## Contrasting CI result

SciMLBase Integration run 29220004511 tested the same NonlinearSolve commit with SciMLOperators 1.24.1. The Brown case remained broken there, so it did not raise an unexpected-pass error, while the generalized Rosenbrock case failed independently.

Run: https://github.com/SciML/SciMLBase.jl/actions/runs/29220004511

The result appears sensitive to the numerical environment. Removing problem 8 from broken_tests would make the test fail in environments where the solver still does not converge, so this issue intentionally does not change or silence the test.

ChrisRackauckas-Claude:
I reproduced and traced this on the exact source commit `247a8bfcbcadc0d76749242f07bbc4bd2761dae9`.

The affected entry is problem 8, “Brown almost linear function”, with:

```julia
Broyden(;
    init_jacobian = Val(:true_jacobian),
    update_rule = Val(:bad_broyden),
)
```

### Supported CI matrix

The exact-base Core run shows a stable architecture split across all tested Julia channels:

| Julia | Ubuntu 24.04 x86_64 | macOS 26 arm64 |
|---|---|---|
[...]

PR #1070, opened 2026-07-15
Align source-tree downgrade dependency floors

ChrisRackauckas-Claude:
Please ignore this PR until it has been reviewed by @ChrisRackauckas.

## Summary

Make the root and local source-package compat bounds describe combinations that the current NonlinearSolve source tree can actually load:

- require StaticArrays 1.9.8 rather than the broad 1.9 floor;
- align `NonlinearSolveFirstOrder`'s SciMLJacobianOperators floor with the source-tree stack at 0.1.1;
- require LinearSolve 5 in the root, Base, FirstOrder, and QuasiNewton projects.

## Why

The deployed downgrade action resolves the monorepo's `[sources]` projects together. The previous bounds admitted three incompatible endpoints:

- StaticArrays 1.9.7 predates the API used by the current extensions; 1.9.8 is the first compatible 1.9 release.
- `NonlinearSolveFirstOrder` allowed SciMLJacobianOperators 0.1.0 while the current local Base stack requires behavior available from 0.1.1.
- LinearSolve 4.3 is the only 4.x endpoint admitted by the existing compat, but current NonlinearSolve caches use the LinearSolve 5 interface. At 4.3 the local FirstOrder and QuasiNewton packages fail to precompile because `LinearSolveJLCache` is not callable. There is no later 4.x release to raise to, so this removes an advertised combination that does not work rather than dropping a working endpoint.

## Process

1. Reproduced the exact lower-bound failures with deployed `julia-downgrade-compat` `v2.6.1` (`fab1def`).
2. Identified StaticArrays 1.9.8 as the first release containing the required API.
3. Aligned the path-source SciMLJacobianOperators floors.
4. Verified that LinearSolve has no compatible post-4.3 release in major 4, then selected the supported 5.x line.
5. Re-resolved the exact floor environment and compared its real Core test result with a clean, latest-compatible checkout at the same master SHA.

## Local verification

- Exact deployed downgrade action completed successfully. Its root floor manifest selects LinearSolve 5.0.0, StaticArrays 1.9.8, and a mutually compatible SciMLJacobianOperators stack.
- Julia 1.10 strict floor run, `GROUP=Core`, `allow_reresolve=false`: the downgrade-only StaticArrays and LinearSolve precompile errors are gone.
- The remaining Core result is `94 passed, 1 failed, 1 errored, 19 broken` in the Broyden robustness table.
- A clean latest-compatible checkout at the exact same master SHA produces the identical `94 passed, 1 failed, 1 errored, 19 broken` result, including the Generalized Rosenbrock algorithm-4 residual `4.4 > 0.001`. This is an existing master failure, not introduced or hidden by this PR; an independent clean reproduction/bisection is in progress under the repository's testing policy.
- The targeted SimpleNonlinearSolve strict Core floor test passed.
- TOML parsing passed for all four edited projects.
- Runic 1.7 check passed for all 373 Julia files.
- `git diff --check` passed.

The strict comparison command was equivalent to:

```sh
GROUP=Core julia +1.10 --project=. -e \
  'using Pkg; Pkg.test(; coverage=true, force_latest_compatible_version=false, allow_reresolve=false)'
```


ChrisRackauckas-Claude:
Added `34d336fc` after reproducing both downgrade failures locally with Julia 1.10.11 and the deployed `julia-downgrade-compat` action at `fab1defb76df9fd672f63c94df73ce131d32e134`.

The action resolves registry dependencies while temporarily removing `[sources]` packages. `SciMLOperators` was therefore selected transitively at 1.22.0, and re-adding the local `NonlinearSolveQuasiNewton` / `NonlinearSolveBase` sources made the locked manifest invalid because those packages now require 1.24. Adding `SciMLOperators = "1.24"` as an explicit test extra/target in the root and `SimpleNonlinearSolve` projects makes the floor visible during the downgrade resolution itself.

The adjacent master bisect identifies `41681c5e40f92a517262cdb7c90d1cd95ed7f1cd` (#1060) as the first bad commit; its parent `579761dd26f29ff8d22202e8ff955a8ea35f0ab8` resolves successfully under the same downgrade procedure.

Local verification:
- Root deployed downgrade selected SciMLOperators 1.24.0 and strict `GROUP=Core` with `allow_reresolve=false` passed resolution/precompilation and all earlier Core files, then reached the separately known Broyden baseline failures (`94 passed, 1 failed, 1 errored, 19 broken`).
- `SimpleNonlinearSolve` deployed `alldeps` downgrade selected SciMLOperators 1.24.0; strict `NONLINEARSOLVE_TEST_GROUP=Core` with `allow_reresolve=false` passed.
- Root `GROUP=QA`: 15/15 passed.
- `SimpleNonlinearSolve` QA: 16/16 passed.
- Runic 1.7.0 `--check`, TOML parsing, and `git diff --check` passed.

ChrisRackauckas-Claude:
CI confirms this PR fixes the downgrade resolver failure: `julia-downgrade-compat` and build steps now pass, and the test job runs through the suite to the separately reproduced Broyden baseline (`94 pass / 1 fail / 1 unexpected-pass / 19 broken`). The remaining Generalized Rosenbrock failure is the #1039 regression addressed by #1057; the Brown platform-dependent Unexpected Pass is tracked by #1056. This PR does not suppress or relabel either test.

PR #1071, opened 2026-07-15
Add Kantorovich homotopy continuation

ChrisRackauckas-Claude:
> **Please ignore this draft until it has been reviewed by @ChrisRackauckas.**

## Summary

- Add `KantorovichHomotopy`, a documented `HomotopyProblem` solver whose parameter-step controller is adapted from `ImplicitDiscreteSolve.KantorovichTypeController`.
- Reuse the `HomotopySweep` path driver, including cache reuse, endpoint polishing, predictor trust, tracking limits, Jacobian forwarding, and warm handoff to `ArcLengthContinuation`.
- Measure inner-corrector residual contraction when the selected solver exposes an iterative cache; use the controller's `Θmin` convention for polyalgorithms and no-init solvers.
- Add strict contraction rejection, the Deuflhard step-size formula, constant/secant predictor orders, validation, tests, docs, and the required minor version bumps.

## Why

`ImplicitDiscreteSolve` uses a Newton--Kantorovich continuation-style controller across accepted discrete states, but it is not itself one fixed `HomotopyProblem`: its residual mutates the previous accepted state. The reusable part is the contraction-based path-step controller. This PR implements that controller for a genuine fixed homotopy while retaining NonlinearSolve's existing continuation machinery.

OrdinaryDiffEq already has a generic `HomotopyNonlinearSolveAlg` adapter that builds the normalized per-stage homotopy from the known zero-step anchor. No adapter-specific source change is needed: callers can pass `HomotopyNonlinearSolveAlg(KantorovichHomotopy(...))` once this solver is available.

This feature branch is rebased directly onto current `master`; the former #1075 and #1079 prerequisites are now included upstream.

## Comparison

After correcting the controller to use the actual first residual contraction (predictor residual to first corrected residual), explicit finite-difference Newton comparisons are mixed. Values below are homotopy residual-function evaluations.

| Problem | Sweep (secant) | Kantorovich (constant) | Kantorovich (secant) | ArcLength |
|---|---:|---:|---:|---:|
| Smooth cubic | 66 | 51 | 57 | 101 |
| Curved exponential | 96 | 150 | 99 | 230 |
| Narrow transition | 78 | 216 | 285 | 1497 |
| Coupled cubic (n=50) | 538 | 792 | 483 | 707 |
| Two-fold S curve | 478 success via path jump | 3163 strict failure | 2843 strict failure | 320 success |

[...]

PR #1075, opened 2026-07-15
Fix HomotopyContinuation downgrade resolution

ChrisRackauckas-Claude:
## Summary

- require registered NonlinearSolve 4.21 for the HomotopyContinuation test target
- align the SciMLBase floor with NonlinearSolve 4.21
- restore compatibility with the local NonlinearSolveBase LinearSolve 5 floor merged in #1070

## Cause

The downgrade action pins SciMLBase 3.33 and later Pkg.test resolves the NonlinearSolve test extra to 4.20.3. That registered release only accepts LinearSolve through 4, while the local NonlinearSolveBase 2.35 source now requires LinearSolve 5.

## Verification

- reproduced job 87327953938 locally with the exact Julia 1.10.11 downgrade action and strict allow_reresolve=false test command
- reran the same downgrade action after the fix: SciMLBase 3.34.0, NonlinearSolve 4.21.0, and LinearSolve 5.0.0
- exact HomotopyContinuation Core downgrade tests: AllRoots 169/169 and Single Root 50/50
- Runic 1.7.0 check: 373/373 files
- Project.toml parsed with TOML stdlib


ChrisRackauckas-Claude:
CI note: the root Downgrade / Downgrade Tests - Core job reaches the same current-master Broyden baseline documented on #1071: 94 tests pass, the generalized Rosenbrock case fails for algorithm 4, Brown has a platform-dependent unexpected pass, and 19 are already broken. The clean-master reproduction/bisect points to the existing #1057 and #1056 work; this dependency-floor PR does not change or suppress those tests.

ChrisRackauckas-Claude:
Verified this floor fix against the current Kantorovich feature branch (#1071), not only against clean master.

Using the deployed `julia-downgrade-compat` action at `fab1def` in `alldeps` mode, the #1071 branch stacked only on this commit resolves:

- NonlinearSolve 4.21.0
- in-tree NonlinearSolveBase 2.36.0
- SciMLBase 3.34.0
- SciMLLogging 2.0.3
- LinearSolve 5.0.0

The subsequent strict Julia 1.10 test with `force_latest_compatible_version=false` and `allow_reresolve=false` passes AllRoots 169/169 and Single Root 50/50.

The exact clean-master comparator at `6264446f4` instead fails in the Pkg test sandbox before running tests: SciMLBase 3.33.0 constrains registered NonlinearSolve to at most 4.20.3 (LinearSolve at most 4.3), conflicting with in-tree NonlinearSolveBase 2.35.0 requiring LinearSolve 5. This confirms the two compat-floor changes here are sufficient for #1071; the independent root/Base floor PR #1079 is not part of that feature stack.

PR #1077, opened 2026-07-15
Test Brown bad-Broyden convergence on Linux x86_64

ChrisRackauckas-Claude:
> [!IMPORTANT]
> Please ignore this draft until it has been reviewed by @ChrisRackauckas.

## Summary

- Require Brown almost-linear problem 8 to converge for true-Jacobian bad Broyden only on the supported Linux x86_64 lanes.
- Retain the existing broken expectation on macOS arm64 and every unverified platform.
- Leave the unrelated generalized Rosenbrock expectation unchanged.

## Root cause

The exact-base failure is an unexpected pass, not a new loss of convergence. A frozen-dependency bisect identified 6a7687f542cde52b4bfeef5b473c6156c8d453a1 (#1039) as the first commit where this unglobalized solve enters the convergent basin. The inverse-workspace change is at roundoff scale, but the Brown start is ULP-sensitive: adjacent floating-point starts switch between a residual of about `7.31e-13` and an `Unstable` return with residual `5.5`.

The supported CI matrix consistently splits by platform across Julia 1.10.11, 1.12.6, and 1.13.0-rc1: Ubuntu 24.04 x86_64 converges, while macOS 26 arm64 remains above the `1e-3` tolerance. Current and downgrade-floor dependency graphs both converge locally on Linux x86_64. Detailed reproduction, dependency, CI, bisect, and perturbation evidence is recorded in #1056.

Li–Fukushima globalization makes representative adjacent-float starts converge, but changing Broyden's default globalization would be a broad algorithm behavior change. This PR therefore scopes the stronger test expectation only to the platform proven by the supported matrix. Unlike #1069, it does not alter the unrelated Rosenbrock expectation or require Brown to converge on macOS arm64.

## Validation

- Julia 1.12.6, current graph: Brown problem 8 passed 20/20 times; residual `7.314149286230531e-13` every run.
- Julia 1.12.6, downgrade-floor graph: Brown problem 8 passed 20/20 times; same residual every run.
- Full Broyden robustness file, current graph: `95 passed, 1 failed, 19 broken`; Brown alg #4 passed normally. The sole failure is the pre-existing generalized Rosenbrock alg #4 result (`4.4 > 1e-3`), which is being investigated separately.
- Full Broyden robustness file, downgrade-floor graph: identical `95 passed, 1 failed, 19 broken` result and the same sole pre-existing Rosenbrock failure.
- Runic 1.7.0: `--check .` passed.
- `git diff --check` passed.

The official downgrade workflow at the investigated base stopped during package resolution before Core tests ran; the floor graph above was therefore reproduced in a pinned local environment.

### DIFF
--- a/test/Core/23_test_problems_tests__item7.jl
+++ b/test/Core/23_test_problems_tests__item7.jl
@@ -12,11 +12,6 @@ alg_ops = (
 broken_tests = Dict(alg => Int[] for alg in alg_ops)
 broken_tests[alg_ops[2]] = [1, 5, 8, 11, 18]
 broken_tests[alg_ops[4]] = [5, 6, 8, 11]
-if Sys.islinux() && Sys.ARCH === :x86_64
-    # Roundoff selects different basins: supported Linux x86_64 lanes converge,
-    # while macOS arm64 does not.
-    broken_tests[alg_ops[4]] = [5, 6, 11]
-end
 if Sys.isapple()
     broken_tests[alg_ops[1]] = [1, 5, 11]
     broken_tests[alg_ops[3]] = [1, 5, 6, 9, 11]
--- a/test/Core/23_test_problems_tests__item7.jl
+++ b/test/Core/23_test_problems_tests__item7.jl
@@ -12,6 +12,11 @@ alg_ops = (
 broken_tests = Dict(alg => Int[] for alg in alg_ops)
 broken_tests[alg_ops[2]] = [1, 5, 8, 11, 18]
 broken_tests[alg_ops[4]] = [5, 6, 8, 11]
+if Sys.islinux() && Sys.ARCH === :x86_64
+    # Roundoff selects different basins: supported Linux x86_64 lanes converge,
+    # while macOS arm64 does not.
+    broken_tests[alg_ops[4]] = [5, 6, 11]
+end
 if Sys.isapple()
     broken_tests[alg_ops[1]] = [1, 5, 11]
     broken_tests[alg_ops[3]] = [1, 5, 6, 9, 11]


ChrisRackauckas-Claude:
Withdrawing this draft: downgrade CI disproved the proposed Linux x86_64 expectation.

Official downgrade [job 87345863358](https://github.com/SciML/NonlinearSolve.jl/actions/runs/29413493910/job/87345863358) used Julia 1.10.11 and the action-generated graph (LinearSolve 5.0.0, NonlinearProblemLibrary 0.1.7, SciMLBase 3.34.0, SciMLOperators 1.24.0, SciMLJacobianOperators 0.1.15) on Linux x86_64/EPYC 9354. Brown alg #4 produced residual `5.5`; the Broyden summary was `94 passed, 2 failed, 19 broken`.

An exact clean-depot reproduction of the action SHA/mode, merge commit, Pkg.test flags, and package versions on Linux x86_64/EPYC 7502 instead made Brown pass; the summary was `95 passed, 1 failed, 19 broken`, with only the independently tracked Rosenbrock residual `4.4`.

Changing only SciMLJacobianOperators to the in-tree 0.1.16 source did not change the local result. Its runtime source is byte-identical to registered 0.1.15; only version/compat metadata differs, and the tested SciMLOperators 1.24.0 already satisfies that metadata.

Therefore there is no stable OS/architecture/package condition for this expectation. The solve is ULP/hardware-sensitive. Robustness requires a broader globalization behavior change, not test bookkeeping.

Commit `66bc20484` reverts the proposed condition, so this branch now has no source diff from master. Detailed corrected evidence is in https://github.com/SciML/NonlinearSolve.jl/issues/1056#issuecomment-4980987394.

PR #1082, opened 2026-07-15
Add optional post-update `corrector` hook to first-order solvers

pepijndevos:
### What

Adds an optional `corrector` slot to `GeneralizedFirstOrderAlgorithm`, threaded
through `NewtonRaphson(; corrector = ...)`. When set, it runs **once per
iteration, at the seam between the state update (`u ← u + αδu`) and the residual
re-evaluation** — i.e. immediately after `axpy!` and before `evaluate_f!` — in
the `:None` and `:LineSearch` globalizations. It is called as

```julia
corrector(u, u_prev, p)   # mutates u in place
```

so the corrected `u` is what the subsequent function evaluation and termination
check observe.

- **Default `nothing` is a byte-identical no-op** (guarded `=== nothing` return;
  no allocation, no call). Behavior of every existing solver is unchanged.
- **`:TrustRegion` is unsupported** and documented as such: it produces its update
  and residual together inside the trust-region cache, so there is no post-update
  seam. A corrector passed to a trust-region algorithm is simply never invoked.

### Why

Projected / limited Newton schemes need to pull the proposed iterate onto a
feasible manifold *between* the linear step and the next residual evaluation.
The motivating case is **SPICE-style junction limiting** (the "correct" phase of
the Predictor/Corrector Newton-Raphson scheme of Aadithya, Keiter & Mei,
*"Achieving Robust and High-Fidelity Circuit Simulation…"*, Sandia
SAND2018-5689C): after each Newton update, the limiting components of the state
are replaced by the limited branch voltages the devices recorded during
stamping, which then act as the linearization point (`vold`) for the next
iteration. There is currently no hook at this point — `callback_into_cache!`
fires *after* the residual and termination check, and Eisenstat–Walker forcing
fires *pre*-update and only tunes the linear-solve tolerance — so this is a
genuinely new insertion point.

### Consumer evidence

Downstream in [Cadnip.jl](https://github.com/NyanCAD/Cadnip.jl) (an MNA circuit
simulator) this hook replaces a hand-rolled predictor/corrector Newton loop. The
Cadnip corrector copies the devices' recorded limited voltages into the trailing
"limit" slots of the state; `p` carries the simulator workspace holding them.
Driving the diode/BJT DC-operating-point benchmarks through
`NewtonRaphson(corrector = ...)` reproduces the hand-rolled loop **iteration-for-
iteration**, with solutions agreeing to the solve's conditioning-limited floor:

| circuit (Vsrc) | hand-rolled iters | corrector-hook iters | rel. soln. agreement |
|----------------|------------------:|---------------------:|---------------------:|
| rectifier (5 V)   | 7  | 7  | exact |
| rectifier (50 V)  | 6  | 6  | exact |
| chain3 (50 V)     | 6  | 6  | exact |
| graetz (20 V)     | 14 | 14 | 1.0e-11 |
| graetz (325 V)    | 14 | 14 | 5.5e-10 |
| mul4 (50 V)       | 14 | 14 | 5.7e-16 |
| darlington (3 V)  | 12 | 12 | 1.7e-15 |
| darlington (5 V)  | 12 | 12 | 2.8e-15 |

(All converge with `‖F‖ < 1e-10`. The 325 V case's `~5e-10` relative difference
is the floor of a `abstol=1e-10` solve on a matrix with 1 MΩ ground resistors,
and stems only from the linear-solver rounding, not the hook.)

### Tests

`lib/NonlinearSolveFirstOrder/test/misc_tests.jl` gains two items:
- `:None` globalization — verifies the corrector is invoked once per step with
  `(u_proposed, u_prev, p)` in the right order, that its in-place mutation is seen
  by the subsequent residual, that `nothing`/an inert corrector are no-ops
  matching a corrector-less run, and that the projecting corrector converges the
  slaved system in fewer steps.
- `:LineSearch` honors the seam; `:TrustRegion` leaves the corrector inert.

### Notes / follow-up

This PR ships only the generic hook. A follow-up will propose a flagship
`PCNR`-style algorithm instantiation and an increment-based (`‖u_{n+1}−u_n‖ ≤
tol`) `Success` termination mode, which together let a recorded-`w` corrector
terminate without the one-iterate residual lag inherent to reading the
*previous* stamp.

### DIFF
--- a/lib/NonlinearSolveFirstOrder/src/raphson.jl
+++ b/lib/NonlinearSolveFirstOrder/src/raphson.jl
@@ -2,7 +2,7 @@
     NewtonRaphson(;
         concrete_jac = nothing, linsolve = nothing, linesearch = missing,
         autodiff = nothing, vjp_autodiff = nothing, jvp_autodiff = nothing,
-        forcing = nothing,
+        forcing = nothing, corrector = nothing,
     )
 
 An advanced NewtonRaphson implementation with support for efficient handling of sparse
@@ -26,11 +26,15 @@ for large-scale and numerically-difficult nonlinear systems.
     linear system is solved at each iteration. Use `EisenstatWalkerForcing2()` for the
     classical Eisenstat-Walker adaptive forcing strategy. Defaults to `nothing` (fixed
     tolerance from the termination condition).
+  - `corrector`: An optional post-update hook run once per iteration, immediately after the
+    Newton step and before the residual re-evaluation. Called as `corrector(u, u_prev, p)`,
+    expected to mutate `u` in place. Defaults to `nothing` (no-op). See
+    [`GeneralizedFirstOrderAlgorithm`](@ref) for the full contract.
 """
 function NewtonRaphson(;
         concrete_jac = nothing, linsolve = nothing, linesearch = missing,
         autodiff = nothing, vjp_autodiff = nothing, jvp_autodiff = nothing,
-        forcing = nothing,
+        forcing = nothing, corrector = nothing,
     )
     return GeneralizedFirstOrderAlgorithm(;
         linesearch,
@@ -38,6 +42,7 @@ function NewtonRaphson(;
         autodiff, vjp_autodiff, jvp_autodiff,
         concrete_jac,
         forcing,
+        corrector,
         name = :NewtonRaphson
     )
 end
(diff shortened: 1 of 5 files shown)


ChrisRackauckas:
This should be covered by the pre/postconditioners

PR #1085, opened 2026-07-16
Solve a HomotopyProblem + standard nonlinear algorithm at its target λ

ChrisRackauckas-Claude:

## Summary

Adds a fallback so a `SciMLBase.HomotopyProblem` handed a **standard (non-continuation) nonlinear algorithm** is solved by the continuation `HomotopyPolyAlgorithm` instead of erroring.

```julia
function CommonSolve.solve(
        prob::SciMLBase.HomotopyProblem, ::AbstractNonlinearSolveAlgorithm, args...; kwargs...)
    return CommonSolve.solve(prob, HomotopyPolyAlgorithm(), args...; kwargs...)
end
```

## Why

A `HomotopyProblem`'s residual is the λ-extended `f(u, p, λ)`, so it cannot be driven by an ordinary Newton-type algorithm. But `HomotopyProblem <: AbstractNonlinearProblem`, so any init path that selects a *default nonlinear solver for the problem type* hands it one and hits a `MethodError` (`get_concrete_problem(::HomotopyProblem)`):

- SciMLBase's `OverrideInit` computes `nlsolve_alg = default_nlsolve(..., initializeprob, ...)`, which for a `HomotopyProblem` (an `AbstractNonlinearProblem`) returns `FastShortcutNonlinearPolyalg` and then calls `solve(initprob, that_alg)`.
- SteadyStateDiffEq and other callers do the same.

This is exactly the integration needed for ModelingToolkit's new **homotopy-operator initialization** (SciML/ModelingToolkit.jl companion PR): when the init system contains Modelica `homotopy(actual, simplified)` nodes, MTK builds a `HomotopyProblem` initializeprob, and this fallback lets `solve(odeprob)` initialize it by continuation with no per-integrator glue. The existing `solve(prob::HomotopyProblem, ::Nothing)` method already anticipated this ("SciMLBase OverrideInit with a default nlsolve"); this generalizes it to the concrete default the integrators actually pass.

## Dispatch safety

`HomotopySweep`, `ArcLengthContinuation`, and `HomotopyPolyAlgorithm` each have a strictly more-specific `solve(::HomotopyProblem, ::T)` method, so this fallback fires **only** for non-continuation algorithms, and the `solve(prob, HomotopyPolyAlgorithm())` call dispatches to the specific polyalgorithm method — no recursion. (Routing to the default `HomotopyPolyAlgorithm()` rather than threading the passed algorithm as an inner corrector avoids feeding a `Simple*` solver — e.g. the `SimpleTrustRegion` the integrators pick for `StaticArray` `u0` — into the sweep's cache interface.)

## Tests

Adds `test/Core/homotopy_polyalg_tests__item3.jl` (registered in `runtests.jl`): standard algorithms (`NewtonRaphson`, `TrustRegion`, `FastShortcutNonlinearPolyalg`) on a `HomotopyProblem` are rerouted to continuation and rescue an out-of-basin `atan` guess to the true root; the fallback produces the same answer as `HomotopyPolyAlgorithm()`; and the continuation algorithms keep their own dispatch (no recursion). Verified locally under Julia 1.12 against NonlinearSolveBase 2.35.0 + this method (also end-to-end through the MTK companion: `solve(odeprob)` initializes a homotopy init system to the correct root).

Minor version bump `2.35.0 → 2.36.0` (adds a `solve` dispatch).

## Process

Companion to SciML/ModelingToolkit.jl (init → homotopy routing). Design chosen after an adversarial review showed the alternative — a `default_nlsolve(::HomotopyProblem) = nothing` in OrdinaryDiffEqNonlinearSolve — has a `default_nlsolve` method ambiguity for `StaticArray` `u0` and only fixes the OrdinaryDiffEq caller; this single method fixes every caller. Runic-formatted.

### DIFF
--- a/lib/NonlinearSolveBase/Project.toml
+++ b/lib/NonlinearSolveBase/Project.toml
@@ -1,6 +1,6 @@
 name = "NonlinearSolveBase"
 uuid = "be0214bd-f91f-a760-ac4e-3421ce2b2da0"
-version = "2.35.0"
+version = "2.36.0"
 authors = ["Avik Pal <avikpal@mit.edu> and contributors"]
 
 [deps]
--- a/lib/NonlinearSolveBase/src/homotopy_polyalg.jl
+++ b/lib/NonlinearSolveBase/src/homotopy_polyalg.jl
@@ -214,3 +214,21 @@ function CommonSolve.solve(
     end
     return
 end
+
+# A `HomotopyProblem` handed a *standard* nonlinear algorithm cannot be driven by it: the
+# residual is the λ-extended `f(u, p, λ)`, not a plain `f(u, p)`, so the ordinary
+# `AbstractNonlinearProblem` solve pipeline (which would call `f(u, p)`) does not apply.
+# Route any such call to the continuation `HomotopyPolyAlgorithm` default. This lets init
+# paths that select a *default* nonlinear solver for the problem — SciMLBase's OverrideInit,
+# SteadyStateDiffEq, etc., all of which see a `HomotopyProblem` as the `AbstractNonlinearProblem`
+# it subtypes — solve it robustly by continuation without any per-caller glue.
+#
+# `HomotopySweep`, `ArcLengthContinuation`, and `HomotopyPolyAlgorithm` each have their own,
+# strictly more specific `solve(::HomotopyProblem, ::T)` methods, so this fallback fires only
+# for non-continuation algorithms; the `solve(prob, HomotopyPolyAlgorithm())` call below
+# dispatches to the specific method, so there is no recursion.
+function CommonSolve.solve(
+        prob::SciMLBase.HomotopyProblem, ::AbstractNonlinearSolveAlgorithm, args...; kwargs...
+    )
+    return CommonSolve.solve(prob, HomotopyPolyAlgorithm(), args...; kwargs...)
+end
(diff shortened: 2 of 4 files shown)


ChrisRackauckas-Claude:
Companion (MTK-side) PR: SciML/ModelingToolkit.jl#4766 — routes DAE/ODE initialization to a `HomotopyProblem` when the init system contains Modelica `homotopy` operators. Both are drafts; please ignore until reviewed by @ChrisRackauckas.

ChrisRackauckas-Claude:
Reworked per @ChrisRackauckas's note that the previous approach was wrong: instead of routing every algorithm to `HomotopyPolyAlgorithm`, the method now fixes λ at `λspan[2]` and solves the resulting `NonlinearProblem` with the requested algorithm (reusing `_sweep_nonlinear_function` to λ-fix the residual + jac/prototype/sparsity/colorvec). Continuation stays opt-in via `nothing`/`HomotopySweep`/`ArcLengthContinuation`/`HomotopyPolyAlgorithm`. The init integration moves to a `default_nlsolve → nothing` companion in OrdinaryDiffEq (linked below). 19/19 tests.

ChrisRackauckas-Claude:
Init companion: SciML/OrdinaryDiffEq.jl#3989 (`default_nlsolve → nothing` for homotopy init).

PR #1087, opened 2026-07-16
Restore root LinearSolve 5 floor

ChrisRackauckas-Claude:
> [!IMPORTANT]
> Ignore this PR until it has been reviewed by @ChrisRackauckas.

## Summary

- Restore the root `LinearSolve` compatibility floor to version 5, matching the current in-tree Base, FirstOrder, and QuasiNewton source packages.
- Bump the root package patch version from 4.21.1 to 4.21.2 so the compatibility correction is releasable.

## Root cause

PR #1070 established that the current source-tree caches require the LinearSolve 5 interface: LinearSolve 4.3 fails when the local FirstOrder and QuasiNewton packages use callable `LinearSolveJLCache` behavior. PR #1079 subsequently broadened only the root compatibility entry from `5` back to `4.3, 5`, while the in-tree `NonlinearSolveQuasiNewton` 1.14.1 project still requires LinearSolve 5.

The deployed downgrade action therefore selected and explicitly pinned LinearSolve 4.3.0 from the root floor. `Pkg.test(...; allow_reresolve=false)` then combined that manifest with the in-tree source packages and failed before loading test code:

```text
LinearSolve possible versions: 0.1.0-5.0.0
restricted to version 5 by NonlinearSolveQuasiNewton 1.14.1
restricted to version 4.3.0 by an explicit requirement — no versions left
```

This is the failure in PR #1061's unrelated downgrade job 87701889361 (run 29522250552). The exact same failure reproduces on clean upstream `master` at `60736b5`, so this PR contains no #1061 code.

A formal bisect over the relevant clean-master range reported:

```text
60736b5b6a30a42001c1938647362ef138b1c6a7 is the first bad commit
```

The good midpoint, `cbabebe8` (#1075), selected LinearSolve 5.0.0 and reached the Core test body. The first bad commit is `60736b5` (#1079), whose root-only compatibility change reopened the incompatible 4.3 endpoint.

## Validation

All Julia validation used Julia 1.10.11 and workspace-local depots.

- The exact deployed `julia-actions/julia-downgrade-compat@v2` implementation (`fab1defb76df9fd672f63c94df73ce131d32e134`) completed on the final diff with the workflow-derived stdlib and `[sources]` skip list, mode `deps`, Julia floor `1`, and `Mooncake` no-promote setting.
- The resulting manifest contains `NonlinearSolve 4.21.2` and `LinearSolve 5.0.0`.
- The exact CI test invocation reached and ran Core tests instead of failing in dependency resolution:

  ```julia
[...]

PR #1073, opened 2026-07-15
Use public cache construction for ForwardDiff wrappers

ChrisRackauckas-Claude:
> **Ignore this PR until it has been reviewed by @ChrisRackauckas.**

## Summary

- stop naming the non-public `FunctionWrappersWrappers.SingleCacheStorage` type
- derive the cache-storage type selected by the public `SingleCache()` constructor through documented type parameters
- use the documented explicit-parameter wrapper constructor without inference widening
- exercise first use and cached reuse through the non-isbits `BigFloat` fallback path
- require FunctionWrappersWrappers 1.9.3, the first release providing the public constructor, and bump NonlinearSolveBase to 2.35.3

## Root cause

`NonlinearSolveBaseForwardDiffExt` directly constructed `FunctionWrappersWrappers.SingleCacheStorage`, which is intentionally not public. ExplicitImports therefore failed the NonlinearSolveBase QA job with `NonPublicQualifiedAccessException`.

The first public-constructor prototype removed that violation but widened both continuation driver return types to `Any`. The final implementation asks the public cache-mode constructor which documented `CS` type it selected, then uses the public `FunctionWrappersWrapper{FW,P,CS}(f)` constructor with a concrete heterogeneous wrapper tuple.

The compat floor matters: the explicit public constructor is absent in FunctionWrappersWrappers 1.0 and first appears in 1.9.3. Version 2.35.3 avoids the independently reserved NonlinearSolveBase 2.35.1 and #1061 version 2.35.2.

## Investigation

- reproduced the original QA failure on clean upstream with FunctionWrappersWrappers 1.10.1 and ExplicitImports 1.15.0
- bisected the QA transition to `0343930c2e53b47da8fa35d06379b12f5875b584` (#1046)
- reproduced the continuation inference regression independently of the Jacobian-reuse feature
- inspected released FunctionWrappersWrappers tags to establish 1.9.3 as the first valid public-API floor
- kept all access on documented public API

## Validation

- Julia 1.10.11, FunctionWrappersWrappers pinned exactly to 1.9.3, `GROUP=Core Pkg.test()` for `lib/NonlinearSolveBase`: passed
  - wrapper fallback: 6/6
  - linsolve workspace: 53/53
  - linear routing: 12/12
  - dense LU allocation coverage: 14/14
  - remaining Base Core testsets passed
- Julia 1.12.6 Base Core: passed
- Julia 1.12.6 Base QA: 18/18
- exact repository continuation inference regression: 26/26
[...]

PR #1089, opened 2026-07-16
Preserve inference for operator-backed linear caches

ChrisRackauckas-Claude:

Depends on #1073 and #1087.

## Summary

- materialize and own convertible operator Jacobians for the default and concrete-matrix linear solver paths
- preserve operator-backed caches for explicitly matrix-free linear solvers
- refresh an owned concrete cache from the externally maintained operator without rebinding or aliasing its storage
- unwrap `AutoSpecialize` problems only for cache consumers that actually use Enzyme-backed differentiation
- add an end-to-end inference and solve regression for the default `TrustRegion` path, with Enzyme loaded before cache inference is checked
- bump NonlinearSolveBase to 2.35.1

## Root cause

OrdinaryDiffEq's reused `W` is represented as a `MatrixOperator`. Embedding the full operator-backed `LinearCache` in `LinearSolveJLCache` exceeded Julia's inference budget, so `construct_linear_solver` inferred an existential inner cache type even though the runtime value was concrete. This broke OrdinaryDiffEqBDF's existing `@inferred` cache-construction test after OrdinaryDiffEq commit `9a2f7e7b`.

Default and factorization paths already require a convertible concrete matrix. Materializing an owned copy at initialization keeps their cache type inferable and prevents in-place factorization from modifying the externally owned operator storage. Explicit Krylov paths still retain the operator and remain matrix-free. When the ODE layer refreshes its operator, the LinearSolve extension now copies the new concrete coefficients into the owned cache buffer.

The minimum-compatibility CI graph also loads Enzyme before the operator regression, which selects `AutoEnzyme` as the default VJP backend. `NonlinearSolveFirstOrder.__init` previously unwrapped the problem eagerly for every AD-related cache. Recovering an `AutoSpecialize` function through its type-erased raw-function field then made the Jacobian, descent, and trust-region cache parameters abstract, even though the default direct-operator `TrustRegion` path does not use a VJP.

Problem unwrapping is now scoped to the cache that consumes each AD backend. Generic trust-region schemes unwrap only for Yuan and Bastin, which construct AD-backed operators; custom schemes retain the prior conservative behavior. Line-search, forcing, Jacobian, and Levenberg-Marquardt paths continue to unwrap when their selected backend actually requires Enzyme.

## Validation

- exact CI-floor graph on Julia 1.10.11 with the downgraded manifest and `allow_reresolve=false`: full `NONLINEARSOLVE_TEST_GROUP=Core` passed, including `TrustRegion` 1596/1596, `LevenbergMarquardt` 45/45, and `SciMLOperator Jacobians` 16/16
- latest-compatible graph on Julia 1.12.6 after `Pkg.update()`: full `NONLINEARSOLVE_TEST_GROUP=Core` passed, including `TrustRegion` 1176/1176, `LevenbergMarquardt` 33/33, and `SciMLOperator Jacobians` 16/16
- `NONLINEARSOLVE_TEST_GROUP=Core julia +1.12 --project=lib/NonlinearSolveBase -e 'using Pkg; Pkg.test()'` — passed
- `lib/NonlinearSolveFirstOrder/test/operator_jacobian.jl` with the exact registered LinearSolve 4.3 dependency stack — passed (12/12, 2/2, 2/2)
- OrdinaryDiffEqBDF `test/inference_tests.jl` with the exact registered LinearSolve 4.3 dependency stack — passed; the unmodified base reproduces the inference failure
- repository-wide `Runic --check .` and `git diff --check` — passed

### DIFF
--- a/lib/NonlinearSolveBase/Project.toml
+++ b/lib/NonlinearSolveBase/Project.toml
@@ -1,6 +1,6 @@
 name = "NonlinearSolveBase"
 uuid = "be0214bd-f91f-a760-ac4e-3421ce2b2da0"
-version = "2.35.0"
+version = "2.35.1"
 authors = ["Avik Pal <avikpal@mit.edu> and contributors"]
 
 [deps]
--- a/lib/NonlinearSolveBase/ext/NonlinearSolveBaseLinearSolveExt.jl
+++ b/lib/NonlinearSolveBase/ext/NonlinearSolveBaseLinearSolveExt.jl
@@ -6,6 +6,7 @@ using CommonSolve: CommonSolve, init, solve!
 using LinearSolve: LinearSolve, QRFactorization, SciMLLinearSolveAlgorithm
 using SciMLBase: SciMLBase, ReturnCode, LinearProblem, LinearAliasSpecifier
 using SciMLLogging: @SciMLMessage
+using SciMLOperators: AbstractSciMLOperator
 
 using LinearAlgebra: ColumnNorm, Symmetric
 
@@ -107,6 +108,23 @@ function set_lincache_A!(lincache, new_A)
         return
     end
     lincache.A = new_A
+    return
+end
+function set_lincache_A!(lincache, new_A::AbstractSciMLOperator)
+    if lincache.A isa AbstractSciMLOperator
+        lincache.A = new_A
+        return
+    end
+
+    # A concrete cache owns its materialization, so refresh that buffer rather than
+    # rebinding it to the externally maintained operator.
+    A = convert(AbstractMatrix, new_A)
+    if ArrayInterface.can_setindex(lincache.A)
+        copyto!(lincache.A, A)
+        lincache.A = lincache.A
+    else
+        lincache.A = A
+    end
     return
 end
 
--- a/lib/NonlinearSolveBase/src/linear_solve.jl
+++ b/lib/NonlinearSolveBase/src/linear_solve.jl
@@ -91,9 +91,14 @@ function construct_linear_solver(
         # workspace), so no defensive copy of A is made
         linprob = LinearProblem(A, b, LinearSolveParameters(u_fixed, p); u0 = u_cache)
     elseif A isa AbstractSciMLOperator
-        # A SciMLOperator `A` is externally maintained (refreshed in place by
-        # `update_coefficients!`), so alias it: copying would sever those in-place updates
-        # and, for some operators, change the concrete type (breaking the later `A`-rebind).
+        # An owned materialization protects operator storage from factorizations and keeps
+        # large operator closures out of the default solver's cache type. Explicit
+        # matrix-free solvers retain the operator and its coefficient updates.
+        A = if linsolve === nothing || needs_concrete_A(linsolve)
+            copy(convert(AbstractMatrix, A))
+        else
+            A
+        end
         linprob = LinearProblem(A, b, LinearSolveParameters(u_fixed, p); u0 = u_cache)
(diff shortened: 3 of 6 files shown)


ChrisRackauckas-Claude:
Investigation record for commit `20239c9c9`:

1. Reproduced the downstream OrdinaryDiffEqBDF inference failure on clean OrdinaryDiffEq `aa895b563` with registered NonlinearSolveBase 2.35.0, NonlinearSolveFirstOrder 2.2.0, LinearSolve 4.3.0, and SciMLOperators 1.24.3.
2. With that manifest frozen, OrdinaryDiffEq commit `3b799e583a93adc25b104718242778d3cf7364e1` passed and its child `9a2f7e7b79141951df24a0f2191a8aa20e94f49a` failed. The child introduced reuse of the ODE `W` as an operator-backed inner Jacobian.
3. Reduced the inference boundary to `NonlinearSolveBase.construct_linear_solver`: the runtime `LinearSolveJLCache` was concrete, while its inferred return type existentially abstracted the inner `LinearCache` only when it held the large `MatrixOperator` closure.
4. Verified that materializing an owned matrix for default/concrete solvers restores exact inference. An initial implementation exposed the need to refresh existing concrete caches without assigning the operator into them; the final extension method preserves that ownership invariant.
5. Ran the exact downstream regression, focused operator tests, both affected Core suites, and repository-wide Runic checks before publishing.

ChrisRackauckas-Claude:
Pushed the Enzyme/minimum-compatibility follow-up in `27490c477`.

The CI-only inference regression came from eager problem unwrapping after Enzyme selected `AutoEnzyme` as the default VJP backend. The direct SciMLOperator/default-`TrustRegion` path did not consume that VJP, but recovering the `AutoSpecialize` function through the type-erased raw-function field made the nested cache parameters abstract. Unwrapping is now scoped to the cache consumer; Yuan and Bastin retain the Enzyme-backed operator paths they require.

Local validation after the change:

- Julia 1.10.11, exact downgraded CI-floor manifest, `allow_reresolve=false`, full `NONLINEARSOLVE_TEST_GROUP=Core`: passed; `TrustRegion` 1596/1596, `LevenbergMarquardt` 45/45, `SciMLOperator Jacobians` 16/16.
- Julia 1.12.6, latest-compatible resolver after `Pkg.update()`, full `NONLINEARSOLVE_TEST_GROUP=Core`: passed; `TrustRegion` 1176/1176, `LevenbergMarquardt` 33/33, `SciMLOperator Jacobians` 16/16.
- Repository-wide `Runic --check .` and `git diff --check`: passed.

The regression now loads Enzyme before the existing `@inferred init(prob, TrustRegion())` assertion. No clean-base test failure was observed during this follow-up. The PR body now records its dependencies on #1073 and #1087.

ChrisRackauckas-Claude:
Hosted downgrade classification for head `27490c4771a352791b8fdb42c204ea867c6a9e2a`:

- The downgrade action resolved the unstacked PR merge ref from root `Project.toml`, selecting the declared root floor `LinearSolve 4.3.0`.
- `Pkg.test(...; allow_reresolve=false)` then incorporated the in-tree `NonlinearSolveQuasiNewton 1.14.1`, which requires `LinearSolve 5`, and stopped before loading tests with the expected `4.3.0` explicit-pin versus `5` source-package conflict (job 87761167898).
- This PR explicitly depends on #1087, which restores the root LinearSolve 5 floor. The exact stacked Julia 1.10.11 downgrade graph has already passed the full Core group locally, as recorded above and in the PR body.

No test, assertion, tolerance, dependency skip, or resolver bypass is being changed for this unstacked dependency failure.

Issue #1096, opened 2026-07-20
Machine-dependent failure: 23 Test Problems Broyden alg #4 on Brown almost linear (unmarked broken in #1051, fails on CI runners)


ChrisRackauckas-Claude:
## Summary

`test/Core/setup_robustnesstesting.jl` "23 Test Problems: Broyden" — problem "8: Brown almost linear function", alg #4 (`Broyden(; update_rule = Val(:bad_broyden), init_jacobian = Val(:true_jacobian))`) — fails on master's own CI with

```
Expression: norm(res, Inf) ≤ ϵ
 Evaluated: 5.5 ≤ 0.001
```

but passes on other machines with byte-identical package versions, so the outcome is machine/BLAS-dependent, not version-dependent.

## Evidence

All observations at the same test expression, `test/Core/setup_robustnesstesting.jl:25`:

| Where | Julia | Deps | Result |
| --- | --- | --- | --- |
| master push CI @ 74405d30, `tests / Core (julia lts)` (self-hosted) — [job 88243697971](https://github.com/SciML/NonlinearSolve.jl/actions/runs/29706251772/job/88243697971) | 1.10.11 | latest | fail, 5.5 |
| PR #1095 `tests / Core (julia 1)` (GitHub-hosted) — [job 88367640144](https://github.com/SciML/NonlinearSolve.jl/actions/runs/29746999112/job/88367640144) | 1 (1.12) | latest | fail, 5.5 (only failure in the job) |
| PR #1095 `Downgrade / Downgrade Tests - Core` (self-hosted) — [job 88367372467](https://github.com/SciML/NonlinearSolve.jl/actions/runs/29746998888/job/88367372467) | 1.10.11 | compat floors | fail, 5.5 (only failure in the job) |
| Local x64 Linux reproduction of the Downgrade job (same skip list, same `downgrade.jl`, resolved package versions verified byte-identical to the CI job's `Pkg.status` output) | 1.10.11 | compat floors | **pass** (full Core group green) |
| Downgrade run 2026-07-15 @ cbabebe8 (GitHub-hosted, pre-#1051 tree) — [run 29446704704](https://github.com/SciML/NonlinearSolve.jl/actions/runs/29446704704) | 1.10.11 | compat floors | problem 8 alg #4 **passed** while still marked broken ("Unexpected Pass" error), problem 1 failed |

[...]

Issue #1076, opened 2026-07-15
Julia 1.12 fails homotopy allocation guard due to escaped reshape headers


ChrisRackauckas-Claude:
## Summary

On clean `master` at `6264446f4ce3ad82fea62c71590c64ac04f69bac`, Julia 1.12 fails the homotopy per-step allocation guard in `test/Core/homotopy_alloc_tests__item1.jl`:

- `HomotopyNonlinearSolve.Sweep()`: **154.4 B/step** (expected `< 48`)
- `HomotopyNonlinearSolve.PseudoArcLength()`: **80.0 B/step** (expected `< 48`)

The same code and compatible dependency stack on Julia 1.11.9 reports **0.16 B/step** and **0.0 B/step**, respectively.

## Reproduction

From a clean checkout of `SciML/NonlinearSolve.jl` at `6264446f4ce3ad82fea62c71590c64ac04f69bac`:

```sh
JULIA_PKG_PRECOMPILE_AUTO=0 julia +1.12 --startup-file=no --project=. -e 'using Test; include("test/Core/homotopy_alloc_tests__item1.jl")'
```

Observed on both Julia 1.12.0 and Julia 1.12.6:

```text
test/Core/homotopy_alloc_tests__item1.jl:83
Expression: sweep_per_step(prob_big) < 48
 Evaluated: 154.4 < 48
```

Measuring both warmed helpers from the same test gives:

```text
VERSION=1.12.6 sweep=154.4 arc=80.0
VERSION=1.11.9 sweep=0.16 arc=0.0
```

## Bisect

A formal bisect using the exact test item (and treating commits before the test existed as good) identifies the first failing commit as:

```text
0343930c2e53b47da8fa35d06379b12f5875b584
Make AutoSpecialize infer through functor residuals; zero-alloc homotopy stepping (#1046)
```

The exact allocation values also reproduce when checking out that commit. The commit introduced both the allocation optimization and its guard; its recorded measurements were from Julia 1.11.

## Allocation profile

`Profile.Allocs` shows one recurring pair of escaped array headers per Jacobian evaluation on Julia 1.12:

- 48 B `Matrix{Float64}` header from ForwardDiff's chunk-mode Jacobian result path, which reshapes an already correctly shaped matrix.
- 32 B `Vector{Float64}` header from `ArrayInterface.restructure` through `NonlinearSolveBase.Utils.restructure`, which reshapes an already correctly shaped vector.

[...]

PR #1101, opened 2026-07-22
Stabilize unglobalized bad Broyden steps

ChrisRackauckas-Claude:

## Summary

- Guard unglobalized bad-Broyden steps without adding a function evaluation or allocation to the ordinary path.
- Activate the safeguard only when the initial inverse produces a gross displacement (`norm(du) > 10 * max(norm(u), 1)`), then retain a state-scale bound for that solve.
- Leave ordinary unscaled trajectories bit-for-bit unchanged by returning a sentinel and skipping the displacement mutation.
- Clear the safeguard through public `reinit!` so a reused solver cache cannot inherit limiter state.
- Add an in-place Brown-almost-linear regression through the cached public `init` / `solve!` / `reinit!` interface.
- Bump NonlinearSolve to `4.23.3` and NonlinearSolveQuasiNewton to `1.14.4`.

No test expectation, threshold, tolerance, skip, or broken-test classification is changed.

## Cause

Problem 8 in the robustness matrix, Brown almost linear, fails for algorithm 4:

```julia
Broyden(
    init_jacobian = Val(:true_jacobian),
    update_rule = Val(:bad_broyden),
)
```

At the standard start, the state norm is about `1.58`, while the true-Jacobian displacement norm is about `5288` and the full-step residual is about `1.09e28`. Bad-Broyden inverse updates then follow a roundoff-sensitive trajectory: the same source and dependency graph can converge or return the initial best residual `5.5` depending on the BLAS kernel.

The hard CI boundary is `74405d303` (#1051), which removed Brown from algorithm 4's expected-broken list without changing this solver path. Existing issue #1096 records the machine dependence. A controlled clean-master reproduction on Julia 1.12.6 with `OPENBLAS_CORETYPE=Sandybridge` returned `Unstable`, residual `5.5`, after five steps.

## Current rebase

- Base: `2b0dda5f64402c5177507d9b1bb2e73ed0db4e5b`
- Head: `fb57e9fbde97eaccaddc510d265a7d5c50a34180`
- The current master skip/expected-broken matrix is preserved unchanged.
- The final diff contains only the three source/regression files plus the two package version bumps.

## Local validation

From `lib/NonlinearSolveQuasiNewton` on Julia 1.12.6:

```sh
OPENBLAS_CORETYPE=Sandybridge GROUP=Core \
  julia +1.12 --startup-file=no --project=. \
  -e 'using Pkg; Pkg.test(; julia_args=["--check-bounds=auto", "--compiled-modules=yes", "--depwarn=yes"], force_latest_compatible_version=false, allow_reresolve=true)'
```

Passed all 1001 assertions:

- Broyden: `599 / 599`
- Broyden iterator: `2 / 2`
- Broyden termination: `27 / 27`
- Klement: `216 / 216`
- Klement iterator: `2 / 2`
- Klement termination: `27 / 27`
- LimitedMemoryBroyden: `99 / 99`
- LimitedMemoryBroyden iterator: `2 / 2`
- LimitedMemoryBroyden termination: `27 / 27`
- Final output: `Testing NonlinearSolveQuasiNewton tests passed`

Repository-wide Runic 1.7.0 `--check .` and `git diff --check` both exited successfully.

The previous macOS CI validation of the final safeguard reported the complete root Broyden matrix at `95 pass / 20 expected broken`; the remaining red Core jobs in that run were the separate homotopy allocation baseline tracked in #1076.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "NonlinearSolve"
 uuid = "8913a72c-1f9b-4ce2-8d82-65094dcecaec"
-version = "4.23.2"
+version = "4.23.3"
 authors = ["SciML"]
 
 [deps]
--- a/lib/NonlinearSolveQuasiNewton/Project.toml
+++ b/lib/NonlinearSolveQuasiNewton/Project.toml
@@ -1,7 +1,7 @@
 name = "NonlinearSolveQuasiNewton"
 uuid = "9a2c21bd-3a47-402d-9113-8faf9a0ee114"
 authors = ["Avik Pal <avikpal@mit.edu> and contributors"]
-version = "1.14.3"
+version = "1.14.4"
 
 [deps]
 ArrayInterface = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
--- a/lib/NonlinearSolveQuasiNewton/src/broyden.jl
+++ b/lib/NonlinearSolveQuasiNewton/src/broyden.jl
@@ -122,6 +122,11 @@ end
     tight_step_limit::Bool
 end
 
+function InternalAPI.reinit!(cache::BroydenUpdateRuleCache; kwargs...)
+    cache.tight_step_limit = false
+    return
+end
+
 function unglobalized_step_size(cache::BroydenUpdateRuleCache, u, du, first_step::Bool)
     cache.rule isa BadBroydenUpdateRule || return true
     du_norm = cache.internalnorm(du)
@@ -130,9 +135,10 @@ function unglobalized_step_size(cache::BroydenUpdateRuleCache, u, du, first_step
     if first_step
         cache.tight_step_limit = du_norm > 10 * u_scale
     end
+    cache.tight_step_limit || return true
     # Bad Broyden updates propagate the first displacement through every later inverse
     # update, so retain a state-scale bound when the initial inverse produces a gross step.
-    step_limit = cache.tight_step_limit ? u_scale : 10 * u_scale
+    step_limit = u_scale
     du_norm ≤ step_limit && return true
     return step_limit / du_norm
 end
--- a/lib/NonlinearSolveQuasiNewton/src/broyden.jl
+++ b/lib/NonlinearSolveQuasiNewton/src/broyden.jl
@@ -121,6 +121,17 @@ end
     rule <: Union{BadBroydenUpdateRule, GoodBroydenUpdateRule}
 end
 
+function unglobalized_step_size(cache::BroydenUpdateRuleCache, u, du)
+    cache.rule isa BadBroydenUpdateRule || return true
+    du_norm = cache.internalnorm(du)
+    iszero(du_norm) && return one(du_norm)
+    u_scale = max(cache.internalnorm(u), one(du_norm))
+    # Bad Broyden updates can magnify an oversized displacement in every later inverse
+    # update, so bound unglobalized steps to one order of magnitude relative to the state.
+    step_limit = 10 * u_scale
+    return min(one(du_norm), step_limit / du_norm)
+end
+
 function InternalAPI.solve!(
         cache::BroydenUpdateRuleCache, J⁻¹, fu, u, du; kwargs...
     )
--- a/lib/NonlinearSolveQuasiNewton/src/broyden.jl
+++ b/lib/NonlinearSolveQuasiNewton/src/broyden.jl
@@ -108,7 +108,7 @@ function InternalAPI.init(
     else
         @bb du_cache = similar(du)
     end
-    return BroydenUpdateRuleCache(J⁻¹dfu, dfu, u_cache, du_cache, internalnorm, alg)
+    return BroydenUpdateRuleCache(J⁻¹dfu, dfu, u_cache, du_cache, internalnorm, alg, false)
 end
 
 @concrete mutable struct BroydenUpdateRuleCache <:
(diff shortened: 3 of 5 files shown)


ChrisRackauckas-Claude:
Follow-up `00f93ab9c` addresses the first revision's downgrade CI failures without changing any benchmark expectation or tolerance:

- avoids mutating `δu` when the step-size sentinel says the step is unscaled;
- selects a persistent state-scale bad-Broyden bound only when the initial inverse produces a grossly oversized step;
- keeps the ordinary `10×` path for problems whose initial step is well scaled.

Exact downgraded-dependency validation on Julia 1.10.11:

- Watson / true-Jacobian good Broyden: `Success`, residual `2.3647750424515834e-14`, 219 steps;
- Brown / true-Jacobian bad Broyden: `Success`, residual `1.3988810110276972e-14`, 151 steps;
- full Broyden matrix: 96 pass / 19 unchanged expected broken.

Fresh validation after the commit:

- NonlinearSolveQuasiNewton `GROUP=Core`: 998/998 pass on Julia 1.12.6;
- root Broyden matrix: 96 pass / 19 expected broken;
- repository-wide Runic: clean;
- root `GROUP=Core`: all preceding functional groups pass, then the known clean-master Julia 1.12 allocation pair reproduces unchanged (`154.4` and `80.0` B/step), as audited in #1076.

ChrisRackauckas-Claude:
Follow-up `bb3cd6731` addresses the first macOS CI result without changing the robustness expectations:

- Brown alg #4 passed on macOS.
- The earlier loose `10×` cap changed ordinary trajectories and caused Watson alg #4 plus Freudenstein-Roth alg #3 to fail.
- The safeguard now activates only when the *initial* bad-Broyden displacement is gross; otherwise the original path is bit-for-bit untouched.
- `reinit!` now clears the limiter state, with a cached-reuse regression.

Local verification after the revision:

- Julia 1.12.6 root Broyden matrix: 95 pass / 20 expected broken.
- Exact Julia 1.10.11 downgrade floor: 95 pass / 20 expected broken; cached Brown solve succeeds in 151 steps with residual `1.3988810110276972e-14`.
- NonlinearSolveQuasiNewton `GROUP=Core`: all 1001 assertions pass.
- Runic 1.5.1 and `git diff --check`: clean.

Fresh CI is running on `bb3cd6731`; the macOS Core lanes are the decisive remaining gate.

ChrisRackauckas-Claude:
Fresh CI validation for `bb3cd673`:

- macOS LTS Core is fully green and reports `23 Test Problems: Broyden | 95 pass / 20 expected broken / 115`.
- macOS Julia 1.12.6 Core reports the same clean Broyden matrix. Its only failures are the pre-existing #1076 homotopy allocation assertions (`1111.68 < 48` and `576 < 48` bytes/step).
- macOS Julia 1.13-rc1 Core also reports the same clean Broyden matrix. Its only failures are the same #1076 assertions (`926.4 < 48` and `480 < 48` bytes/step).
- The exact dependency-downgrade workflow is green and reports the same `95 / 20 / 115` matrix.

Thus Brown is fixed on the affected macOS ARM lanes without changing the master expected-broken set or regressing Freudenstein-Roth/Watson. The remaining red Core statuses are unrelated allocation baselines already tracked in #1076.

ChrisRackauckas-Claude:
The remaining QuasiNewton sublibrary QA failure was also confirmed unrelated: clean master reproduces `19 pass / 1 fail` solely from SciMLTesting 2.4 enabling the reexport audit by default. Draft #1102 now covers both root and QuasiNewton QA; its targeted QuasiNewton CI job 89018282756 is green at `20 / 20`, with the audit still enabled.

ChrisRackauckas-Claude:
Rebased onto current clean master `2b0dda5f64402c5177507d9b1bb2e73ed0db4e5b`.

New head: `fb57e9fbde97eaccaddc510d265a7d5c50a34180`.

Reconciliation during the rebase:

- preserved master's current Generalized Rosenbrock skip and expected-broken matrix unchanged;
- bumped NonlinearSolve `4.23.2 → 4.23.3`;
- bumped NonlinearSolveQuasiNewton `1.14.3 → 1.14.4`;
- retained only the focused three-file implementation/regression diff.

Exact local validation on the rebased head:

```sh
cd lib/NonlinearSolveQuasiNewton
OPENBLAS_CORETYPE=Sandybridge GROUP=Core \
  julia +1.12 --startup-file=no --project=. \
  -e 'using Pkg; Pkg.test(; julia_args=["--check-bounds=auto", "--compiled-modules=yes", "--depwarn=yes"], force_latest_compatible_version=false, allow_reresolve=true)'
```

Result: all 1001 assertions passed:

- Broyden `599 / 599`, iterator `2 / 2`, termination `27 / 27`
- Klement `216 / 216`, iterator `2 / 2`, termination `27 / 27`
- LimitedMemoryBroyden `99 / 99`, iterator `2 / 2`, termination `27 / 27`
- final output: `Testing NonlinearSolveQuasiNewton tests passed`

Whole-repository Runic 1.7.0 and `git diff --check` also exited successfully. The branch was force-updated only after fetching origin, verifying that the remote still ended at `bb3cd6731`, and confirming that its three commits were unchanged and authored by the existing PR author.

ChrisRackauckas-Claude:
Closing in favor of https://github.com/SciML/NonlinearSolve.jl/pull/1125.

Problem #8 (Brown almost linear) on `Broyden(; init_jacobian = Val(:true_jacobian), update_rule = Val(:bad_broyden))` is the same knife-edge case as problem #1, which https://github.com/SciML/NonlinearSolve.jl/pull/1108 already moved into `skip_tests` for this exact algorithm. #8 was simply never moved after https://github.com/SciML/NonlinearSolve.jl/commit/74405d3035f759a5631ea5f5776d27d4091063bc (#1051) removed it from the broken list. The consistent fix is the one-line `skip_tests[alg_ops[4]] = [1, 8]`, not a solver change.

The safeguard here changes unglobalized bad-Broyden trajectories for every user whose first step trips `norm(du) > 10 * max(norm(u), 1)`, in order to push one marginal robustness-matrix entry onto the convergent side on one class of hardware. It does not address the underlying machine dependence documented in https://github.com/SciML/NonlinearSolve.jl/issues/1096 — which has failed in both directions, including an "Unexpected Pass" error while #8 was still marked broken — so the next BLAS kernel change can move a different problem across the same cliff. Unglobalized bad Broyden is genuinely unstable on this problem; the test matrix should record that rather than the solver being bent to hide it.

PR #1107, opened 2026-07-23
Add AutoDePSpecialize specialization level (de-specialize p via OpaqueParams)

ChrisRackauckas-Claude:
> [!NOTE]
> Draft PR — please ignore until reviewed by @ChrisRackauckas.

## Summary

Implements the **`AutoDePSpecialize`** specialization level for NonlinearSolve, the analog of SciML/OrdinaryDiffEq.jl#3692 (DiffEqBase). It extends `AutoSpecialize` by additionally *de-specializing the parameter object*: for an in-place `NonlinearProblem` whose `p` is an `isbits` non-`NullParameters` payload, `p` is packed into a `RespecializeParams.OpaqueParams` container and the residual's `FunctionWrappersWrapper` signature carries `OpaqueParams` in the parameter slot (via `RespecializeParams.OpaqueVoid`). The concretized problem type — and the solver compilation — becomes **independent of the user's parameter struct type**, so one precompiled solve is shared across all such types, while the residual still runs fully specialized on `p` (recovered by a type-stable, allocation-free unpack).

```julia
struct MyParams; a::Float64; b::Float64 end
resid!(res, u, p::MyParams) = (res[1] = u[1]^2 - p.a; res[2] = u[2]^2 - p.b; nothing)
prob = NonlinearProblem(NonlinearFunction{true, AutoDePSpecialize}(resid!), [1.0, 1.0], MyParams(2.0, 3.0))
sol = solve(prob, NewtonRaphson())          # hits the shared wrapped path regardless of MyParams
p_recovered = RespecializeParams.unpack(sol.prob.p, MyParams)
```

The marker (`SciMLBase.AutoDePSpecialize`, released in SciMLBase 3.37) and the mechanism (`OpaqueVoid`, `wrap_void_opaque`, `pack`, in **RespecializeParams 1.0**) are the *same* shared components DiffEqBase uses — this PR is the second consumer they were built for.

## How it works (mirrors the existing AutoSpecialize path)

NonlinearSolve already wraps IIP residuals for `AutoSpecialize` in `lib/NonlinearSolveBase/src/autospecialize.jl` (`maybe_wrap_nonlinear_f` → `AutoSpecializeCallable` holding a `FunctionWrappersWrapper` + the original). This PR adds the parallel opaque path:

[...]

PR #1110, opened 2026-07-23
Add and precompile the non-isbits (OpaqueRef) AutoDePSpecialize path

ChrisRackauckas-Claude:

Completes the non-isbits story for `AutoDePSpecialize` on NonlinearSolve, mirroring the ODE side (SciML/OrdinaryDiffEq.jl#4005).

**Two parts, which only make sense together:**

1. **Non-isbits code path** (`NonlinearSolveBase`). #1107 merged `AutoDePSpecialize` with an *isbits-only* policy (`should_opaque_p(p) = isbits(p) && !NullParameters`), so a non-isbits `p` (e.g. a struct with a `Vector` field, or a bare array) was left concrete and got no de-specialization. This relaxes the policy to de-specialize any non-`NullParameters`, non-already-packed `p`, packing via `pack_auto`: isbits → `OpaqueParams` (byte copy), non-isbits → `OpaqueRef` (by reference).

2. **Precompilation** (`FirstOrder`, `QuasiNewton`, `SpectralMethods`). #1107 shipped no workload exercising `AutoDePSpecialize` at all, so even the isbits opaque path compiled on first use. This adds one `AutoDePSpecialize` problem per workload covering **both** containers (isbits `p.a`, non-isbits `p[1]`). Each container is a single wrapped-residual signature shared across all parameter types of its kind, so one precompiled solve per container serves first solves with any struct/array parameter.

### Measured (fresh process, first solve of a never-seen non-isbits parameter struct)

| alg | FullSpecialize | AutoDePSpecialize (this PR) |
|---|---|---|
| NewtonRaphson | 3.02s | **0.88s** |
| Broyden | — | 0.96s |
| DFSane | — | 1.26s |

Without part 1, the non-isbits workload problem in part 2 would hit `should_opaque_p == false` and never take the opaque path. Verified end-to-end: `NonlinearSolveBase` AutoDePSpecialize test suite (incl. new OpaqueRef + idempotency tests) passes 18/18; the three sublibraries precompile cleanly (workload solves run during precompilation and return `Success`); fresh-process first solves timed above.

Version bumps: `NonlinearSolveBase` minor (extends the opt-in feature); `FirstOrder`/`QuasiNewton`/`SpectralMethods` patch.

### Caveats (same posture as the isbits path)
Reverse-mode AD of `p` (SciMLSensitivity) and symbolic parameter indexing read the concrete parameter and are **not** supported under the opaque path — it is a forward-solve latency tool. Reverse-mode support is a planned follow-up. Recover the payload with `RespecializeParams.unpack(sol.prob.p, typeof(p))`.

https://claude.ai/code/session_01UuAbA8hSnrirMrJjFqxwn5

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "NonlinearSolve"
 uuid = "8913a72c-1f9b-4ce2-8d82-65094dcecaec"
-version = "4.23.0"
+version = "4.23.1"
 authors = ["SciML"]
 
 [deps]
--- a/lib/NonlinearSolveBase/Project.toml
+++ b/lib/NonlinearSolveBase/Project.toml
@@ -1,6 +1,6 @@
 name = "NonlinearSolveBase"
 uuid = "be0214bd-f91f-a760-ac4e-3421ce2b2da0"
-version = "2.38.0"
+version = "2.39.0"
 authors = ["Avik Pal <avikpal@mit.edu> and contributors"]
 
 [deps]
--- a/lib/NonlinearSolveBase/src/autospecialize.jl
+++ b/lib/NonlinearSolveBase/src/autospecialize.jl
@@ -219,12 +219,21 @@ end
 """
     should_opaque_p(p) -> Bool
 
-Policy for whether the AutoDePSpecialize path de-specializes `p`: `true` for an
-`isbits` non-`NullParameters` payload, `false` otherwise. Defined locally rather
-than taken from SciMLBase so this does not depend on an unreleased SciMLBase
-symbol.
+Policy for whether the AutoDePSpecialize path de-specializes `p`: `true` for any
+payload except a `NullParameters` sentinel or an already-packed container.
+`isbits` payloads pack into a `RespecializeParams.OpaqueParams` (byte copy),
+everything else into a `RespecializeParams.OpaqueRef` (by reference). Defined
+locally rather than taken from SciMLBase so this does not depend on an unreleased
+SciMLBase symbol.
+
+De-specializing makes the concretized `prob.p` an opaque container, so the level
+is a forward-solve latency tool: reverse-mode AD of `p` and symbolic parameter
+indexing read the concrete parameter and are not supported (reverse-mode support
+is a planned follow-up). Recover the payload with
+`RespecializeParams.unpack(sol.prob.p, typeof(p))`.
 """
-@inline should_opaque_p(p) = isbits(p) && !(p isa SciMLBase.NullParameters)
+@inline should_opaque_p(p) = !(p isa SciMLBase.NullParameters) &&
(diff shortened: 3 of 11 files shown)


ChrisRackauckas-Claude:
Added the same MTK-safety guard as the ODE side (SciML/OrdinaryDiffEq.jl#4005). `MTKParameters` is non-isbits, so without a guard the OpaqueRef path would sweep it in and break MTK initialization + symbolic parameter indexing. `maybe_opaque_wrap` now bails via `!SciMLBase.has_sys(prob.f)`, so a `NonlinearProblem` carrying a symbolic system falls back to normal specialization (a no-op for it anyway). Regression test uses a `SymbolCache` sys; NonlinearSolveBase AutoDePSpecialize suite now 23/23.

ChrisRackauckas-Claude:
Good catch — the polyalgorithms were not covered, and chasing it turned up **two** separate gaps on the paths users actually hit.

**1. Polyalgorithm / no-algorithm entry.** My earlier commit only added the opaque-p problems to the FirstOrder/QuasiNewton/SpectralMethods workloads, which sweep *individual* algorithms. The polyalg + no-alg sweep lives in the main `NonlinearSolve` workload, so `solve(prob)` (default selection via `FastShortcutNonlinearPolyalg`) never saw an `AutoDePSpecialize` problem.

**2. Keyword-argument shape.** The workload’s default sweep passes `abstol`/`verbose`, and the keyword `NamedTuple`’s type participates in the specialization. So precompiling `solve(prob; kwargs...)` does **not** cover the bare `solve(prob)` most user code writes. Measured directly:

```
solve(prob; abstol=1e-2, verbose=NonlinearVerbosity())   0.61s   (workload shape)
solve(prob)                                              1.68s   (bare — uncovered)
```

Fix adds the opaque-p problems to the main workload and gives them a bare no-kwarg `solve(prob)` sweep.

### Measured (fresh process, non-isbits struct `p`, `solve(prob)` with no algorithm)

| | first solve |
|---|---|
| FullSpecialize | 5.04s |
| AutoDePSpecialize — before | 3.11s |
| AutoDePSpecialize — + polyalg problems | 1.69s |
| **AutoDePSpecialize — + bare sweep** | **0.61s** |

0.61s matches the single-algorithm figure (0.72s), so the polyalgorithm path now gets the full first-solve win — an 8.2× improvement over `FullSpecialize` rather than the 1.6× it had.

Verified: NonlinearSolveBase AutoDePSpecialize suite 23/23; polyalg / NewtonRaphson / Broyden / DFSane all converge on a non-isbits opaque-p problem. Patch-bumps `NonlinearSolve` to 4.23.1.

Same class of gap as the ODE side (SciML/OrdinaryDiffEq.jl#4005), where `solve(prob)` with no algorithm was going 11.25s → 0.45s once added to the Default workload.

PR #1115, opened 2026-07-26
Reset force_stop/retcode in polyalg reinit!; add get_u/get_fu

singhharsh1708:
Two gaps in `NonlinearSolvePolyAlgorithmCache` surface when a polyalgorithm cache is driven one `step!` at a time — the pattern implicit ODE/DAE solvers use to run a nonlinear solve across an integrator step — rather than through `solve!`.

## `reinit!` doesn't reset `force_stop`/`retcode`

`InternalAPI.reinit!(::NonlinearSolvePolyAlgorithmCache)` resets `current`/`nsteps`/`stats`/`total_time` but leaves `force_stop` and `retcode` at their converged values. `solve!` (the `@generated` polyalgorithm method) drives the subcaches directly, so it is unaffected. But `CommonSolve.step!` guards on `not_terminated(cache) = !cache.force_stop && cache.nsteps < cache.maxiters`, reading `force_stop` off the *polyalgorithm* cache. So once any branch has converged, a warm-started `reinit!` + `step!` loop no-ops on the first `step!` and reuses the stale iterate.

Minimal reproduction with the public API, no downstream package:

```julia
using NonlinearSolve, SciMLBase
import NonlinearSolveBase, SymbolicIndexingInterface as SII

f(u, p) = u .^ 2 .- p
cache = init(NonlinearProblem(f, [1.0], 4.0), RobustMultiNewton())
drive!(c) = while NonlinearSolveBase.not_terminated(c); step!(c); end

drive!(cache)                          # root of u^2 = 4  -> [2.0]
SciMLBase.reinit!(cache, [1.0]; p = 9.0)
@show cache.force_stop                 # true   (should be false after reinit!)
drive!(cache)                          # root of u^2 = 9  -> should be [3.0]
@show SII.state_values(cache)          # [1.0]  (step! never ran)
```

Fix: reset `force_stop`/`retcode` in `reinit!`, next to the existing `nsteps`/`total_time` resets. `best` is deliberately left untouched so `retain_best` continuation is unaffected (`polyalg_retention_tests` still passes 43/43).

## `get_u`/`get_fu` throw on a polyalgorithm cache

The cache holds `u`/`fu` on `caches[current]`, not at top level, so the generic `get_u(cache) = cache.u` / `get_fu(cache) = cache.fu` fallbacks throw. `SII.state_values` already delegates correctly via `symbolic_container`; this adds the matching `get_u`/`get_fu` methods so the two accessor families agree.

## Tests

`test/Core/polyalg_step_reinit_tests__item1.jl` drives a polyalgorithm cache one `step!` at a time across a `reinit!` for both `RobustMultiNewton` and `FastShortcutNonlinearPolyalg`, and checks `get_u`/`get_fu`. It fails on current `master` (both the `not_terminated` guard and the `get_u`/`get_fu` accessors) and passes with this change. Existing `reinit_tests` (18/18) and `polyalg_retention_tests` (43/43) are unchanged.

Found while fixing SciML/OrdinaryDiffEq.jl#3859, where `NonlinearSolveAlg` wrapping a polyalgorithm hit both issues; the OrdinaryDiffEq side can drop its local workaround once this is released.

## AI Disclosure

Claude assisted with this work.


singhharsh1708:
@ChrisRackauckas

ChrisRackauckas (approved):

PR #1128, opened 2026-07-28
Fix symmetry factor calculation used modified function values after b…

Proektsoft-EOOD:
## Checklist

## Additional context
After bisection fallback, the symmetry factor was calculated using y1 and y2 that have been modified by Anderson-Bjorck correction factors. Not this is fixed by storing the original (unmodified) function values f1 and f2 to be used for that purpose instead of y1 and y2.


Proektsoft-EOOD:
This fix has been proposed by Assoc. Prof. L. Tomov (algorithm co-author).

PR #1146, opened 2026-08-07
Reset the quasi-Newton update rule state on `reinit!`

### LINKED ISSUE #1140: reinit! on a Broyden/Klement cache keeps the previous solve's residual in the update rule cache
`SciMLBase.reinit!(cache, u0; p)` on a `QuasiNewtonCache` resets the counters, retcode, trace and termination cache, but not the update rule's own state. For `Broyden()` the update rule cache holds `dfu`, the residual of the previous iterate that the secant update differences against. After a `reinit!` it still contains the residual at the previous problem's root, so the first secant update of the new solve differences the new residual against a stale one and pushes the iterate in the wrong direction. The first step after a reinit still matches a fresh cache, because `reinit_self!` on `QuasiNewtonCache` does zero `nsteps` and the initialization cache does reset the Jacobian approximation, so the two solves only diverge from the second step onward.

Versions: NonlinearSolve 4.24.0, NonlinearSolveQuasiNewton 1.14.2, NonlinearSolveBase 2.39.1.

## Reproducer (NonlinearSolve only)

```julia
using NonlinearSolve, SciMLBase

f(u, p) = u .^ 2 .- p

probA = NonlinearProblem(f, [1.5, 1.5], [2.0, 3.0])
probB = NonlinearProblem(f, [1.0, 1.0], [9.0, 16.0])

fresh = init(probB, Broyden())
sol_fresh = solve!(fresh)

cache = init(probA, Broyden())
solve!(cache)
@show cache.update_rule_cache.dfu          # residual at A's root
SciMLBase.reinit!(cache, [1.0, 1.0]; p = [9.0, 16.0])
@show cache.update_rule_cache.dfu          # unchanged
@show f([1.0, 1.0], [9.0, 16.0])           # what a fresh cache holds
sol_reused = solve!(cache)
```

Output:

```
cache.update_rule_cache.dfu = [-1.6606849229106047e-10, -3.0631275294012994e-10]   # after solve A
cache.update_rule_cache.dfu = [-1.6606849229106047e-10, -3.0631275294012994e-10]   # after reinit!, unchanged
f([1.0, 1.0], [9.0, 16.0])  = [-8.0, -15.0]                                        # what a fresh cache holds
```

On this easy problem the reused cache still reaches the root, just later: `nsteps = 15`, `nf = 17` against the fresh cache's `nsteps = 14`, `nf = 16`.

## Causation

Hand resetting that one field makes the reused cache match a fresh one exactly:

```julia
cache2 = init(probA, Broyden())
solve!(cache2)
SciMLBase.reinit!(cache2, [1.0, 1.0]; p = [9.0, 16.0])
copyto!(cache2.update_rule_cache.dfu, f([1.0, 1.0], [9.0, 16.0]))
sol_fixed = solve!(cache2)

sol_fixed.u == sol_fresh.u          # true, bitwise
cache2.nsteps == fresh.nsteps       # true, 14 == 14
cache2.stats.nf == fresh.stats.nf   # true, 16 == 16
```

## Where it hurts

The extra iteration above is the benign case. Under `step!` driving with zeroed tolerances, which is how OrdinaryDiffEq's `NonlinearSolveAlg` drives an inner cache one step per outer Newton iteration, the reused cache visibly wastes work before recovering, and on harder problems it walks away from the root entirely. In OrdinaryDiffEq this shows up as `Broyden`, `Klement` and `FastShortcutNonlinearPolyalg` (whose first quasi-Newton branch is `Broyden`) going `Unstable` on stiff test problems where every first order inner solver is clean, because the integrator calls `reinit!` on the inner cache at every stage.

```
reused (p = 9, want 3): 1.5, 2.25, 1.5, 3.3, 2.906, 2.995, 3.0
fresh  (p = 9, want 3): 1.5, 2.25, 3.3, 2.959, 2.998, 3.0, 3.0
```

## Root cause

`@internal_caches QuasiNewtonCache ... :update_rule_cache ...` generates a `reinit!` that recurses into each internal cache and then calls `reinit_self!` on the outer cache. `InternalAPI.reinit_self!` has a silent no op fallback for `::Any`, and neither `BroydenUpdateRuleCache` nor `KlementUpdateRuleCache` defines a method for it, so the recursion lands on the fallback and their state survives. `BroydenUpdateRuleCache.dfu` is the stale field for Broyden; `KlementUpdateRuleCache.fu_cache` plays the same role for Klement. The fix is a `reinit_self!` method for each update rule cache that restores the residual at the (re)initialized iterate, the same way the constructor seeds it.

There is no reset call on the caller's side that would avoid this: `SciMLBase.reinit!(cache, u0; p)` is the whole public API for reusing a cache.

ChrisRackauckas-Claude:
>
> Rebased onto `master` now that #1145 has merged; the diff is this fix alone.

Closes #1140.

## What changed and why

`SciMLBase.reinit!` on a `QuasiNewtonCache` left the update rule's own state alone, so `BroydenUpdateRuleCache.dfu` and `KlementUpdateRuleCache.fu_cache` still held the residual at the *previous* problem's root. The first secant update of the reused cache therefore differenced the new residual against a stale one.

The reason `InternalAPI.reinit_self!` cannot fix this in the update rule cache itself: `@internal_caches` generates a `reinit!` that recurses into the nested caches *first* and calls `reinit_self!` on the enclosing cache afterwards — but the residual at the new iterate is only computed inside that later call, by `Utils.reinit_common!`. The nested cache has nothing to reseed from at the time it runs, and the `::Any` fallback for `reinit_self!` silently leaves it untouched.

So `QuasiNewtonCache`'s `reinit_self!` reseeds the update rule cache right after `reinit_common!`, which is exactly the pattern it already uses for the termination cache two lines below. The hook is a new `NonlinearSolveBase.reset_update_rule_state!(cache, fu)` on `AbstractApproximateJacobianUpdateRuleCache`, with a no-op default — correct for update rules whose cache holds only scratch buffers.

`reset_update_rule_state!` is declared `public` in `NonlinearSolveBase` with a docstring and a `@docs` entry in the devdocs, rather than added to the ExplicitImports ignore list. (QA caught this: the first run failed `all_qualified_accesses_are_public`.)

## Failing before / passing after

New test: after `reinit!`, a reused cache must match a fresh cache **bitwise** in `u`, and exactly in `nsteps` and `stats.nf` — 7 algorithm configurations × oop/iip/scalar, 63 assertions.

Before (fix reverted, test kept):

```
Test Summary:                      | Pass  Fail  Total   Time
all                                |    9    54     63  16.8s
  Broyden                          |    1     8      9   4.9s
  Broyden (bad)                    |    3     6      9   0.9s
  Broyden (diagonal)               |    1     8      9   1.5s
  Broyden (true jacobian)          |          9      9   3.5s
  Klement                          |    1     8      9   0.8s
  Klement (true jacobian diagonal) |    2     7      9   1.0s
  LimitedMemoryBroyden             |    1     8      9   1.9s
```

After:

```
Test Summary: | Pass  Total     Time
all           |   63     63  1m13.7s
```

The issue's own reproducer, before and after:

```
                                   before              after
cache.update_rule_cache.dfu   [-1.66e-10, -3.06e-10]   [-8.0, -15.0]
f([1.0, 1.0], [9.0, 16.0])    [-8.0, -15.0]            [-8.0, -15.0]
nsteps  (fresh 14)            15                       14
nf      (fresh 16)            17                       16
```

## Why this needs #1145

With the reused cache now following the same trajectory as a fresh one, `core_tests__item8.jl` — a 200-point parameter continuation driven by `reinit!` — started failing. That turned out not to be a regression: a **fresh** cache on the state the continuation reaches at `i = 5` fails identically on unmodified master.

```
FRESH cache, u0=0.20000000003657056, p=0.05 -> u=0.2 (unchanged), retcode=Unstable
```

Sweeping fresh caches over all 199 continuation states on master: 4 fail (one diverges, three converge to the negative root). `item8` was passing only because the stale `dfu` happened to steer the loop around them. #1145 fixed that underlying `LimitedMemoryBroyden` bug and has since merged; `item8` is untouched by either PR.

## Verification

`GROUP=Core julia --project -e 'using Pkg; Pkg.test()'` in `lib/NonlinearSolveQuasiNewton` — 1064 assertions, all passing:

```
Broyden                                               |  594    594  8m35.3s
Broyden: Iterator Interface                           |    2      2  2.9s
Broyden Termination Conditions                        |   27     27  9.2s
Klement                                               |  216    216  4m45.9s
Klement: Iterator Interface                           |    2      2  1.3s
Klement Termination Conditions                        |   27     27  7.0s
LimitedMemoryBroyden                                  |   99     99  2m16.4s
LimitedMemoryBroyden: Iterator Interface              |    2      2  2.3s
LimitedMemoryBroyden Termination Conditions           |   27     27  7.1s
Postcondition iterate correction                      |    4      4  2.4s
LimitedMemoryBroyden: continuation from a nearby root |    1      1  0.9s
reinit! resets the update rule state                  |   63     63  11.2s
```

`GROUP=QA` — `NonlinearSolveQuasiNewton QA | 20 20 46.1s` and `NonlinearSolveBase QA | 20 20 50.8s`, both passing.

`julia --project=docs docs/make.jl` — exit 0, so the new `@docs` entry resolves.

Runic and `typos` clean.

## What I did not verify

- The OrdinaryDiffEq symptom described in the issue (`Broyden`/`Klement`/`FastShortcutNonlinearPolyalg` going `Unstable` on stiff problems because `NonlinearSolveAlg` calls `reinit!` on the inner cache every stage). This fixes the root cause the issue identifies, but I did not run OrdinaryDiffEq's suite to confirm the downstream symptom clears.
- GPU, Downstream, and `julia pre` jobs — not run.
- The `NonlinearSolve` top-level test groups — not run.

## Worth pushing back on

`reset_update_rule_state!` is new public API on `NonlinearSolveBase`, which is why that package gets a minor bump rather than a patch. The alternative was keeping it internal and widening the ExplicitImports ignore list; I went with public because the file's own comment flags those entries as "still non-public in their owning packages", i.e. a backlog rather than a target.

Versions bumped: `NonlinearSolveBase` 2.42.0 → 2.43.0 (new public API), `NonlinearSolveQuasiNewton` 1.15.1 → 1.15.2, `NonlinearSolve` 4.26.0 → 4.26.1, with the corresponding compat floors raised.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "NonlinearSolve"
 uuid = "8913a72c-1f9b-4ce2-8d82-65094dcecaec"
-version = "4.26.0"
+version = "4.26.1"
 authors = ["SciML"]
 
 [deps]
@@ -98,9 +98,9 @@ NLSolvers = "0.5"
 NLsolve = "4.5"
 NaNMath = "1"
 NonlinearProblemLibrary = "0.1.2"
-NonlinearSolveBase = "2.42"
+NonlinearSolveBase = "2.43"
 NonlinearSolveFirstOrder = "2"
-NonlinearSolveQuasiNewton = "1.12"
+NonlinearSolveQuasiNewton = "1.15.2"
 NonlinearSolveSpectralMethods = "1.6"
 PETSc = "0.4.2"
 Pkg = "1.10"
--- a/docs/src/devdocs/internal_interfaces.md
+++ b/docs/src/devdocs/internal_interfaces.md
@@ -63,6 +63,7 @@ NonlinearSolveBase.AbstractApproximateJacobianStructure
 NonlinearSolveBase.AbstractJacobianInitialization
 NonlinearSolveBase.AbstractApproximateJacobianUpdateRule
 NonlinearSolveBase.AbstractApproximateJacobianUpdateRuleCache
+NonlinearSolveBase.reset_update_rule_state!
 NonlinearSolveBase.AbstractResetCondition
 ```
 
--- a/lib/NonlinearSolveBase/Project.toml
+++ b/lib/NonlinearSolveBase/Project.toml
@@ -1,6 +1,6 @@
 name = "NonlinearSolveBase"
 uuid = "be0214bd-f91f-a760-ac4e-3421ce2b2da0"
-version = "2.42.0"
+version = "2.43.0"
 authors = ["Avik Pal <avikpal@mit.edu> and contributors"]
 
 [deps]
--- a/lib/NonlinearSolveBase/src/NonlinearSolveBase.jl
+++ b/lib/NonlinearSolveBase/src/NonlinearSolveBase.jl
@@ -133,6 +133,7 @@ include("forward_diff.jl")
 @compat(public, (get_u, get_fu, get_nsteps))
 @compat(public, (construct_linear_solver, needs_square_A, needs_concrete_A, get_linear_cache))
 @compat(public, (construct_jacobian_cache, reused_jacobian))
+@compat(public, (reset_update_rule_state!,))
 @compat(
     public,
     (
--- a/lib/NonlinearSolveBase/src/abstract_types.jl
+++ b/lib/NonlinearSolveBase/src/abstract_types.jl
@@ -588,6 +588,8 @@ Abstract Type for all Approximate Jacobian Update Rule Caches used in NonlinearS
 ### Interface Functions
 
   - `store_inverse_jacobian(cache)`: Return `store_inverse_jacobian(cache.rule)`
+  - `reset_update_rule_state!(cache, fu)`: Reseed any residual the cache carries between
+    iterations with `fu`.
 
 ### `InternalAPI.solve!` specification
 
@@ -602,6 +604,21 @@ abstract type AbstractApproximateJacobianUpdateRuleCache <: AbstractNonlinearSol
 function store_inverse_jacobian(cache::AbstractApproximateJacobianUpdateRuleCache)
     return store_inverse_jacobian(cache.rule)
 end
+
+"""
+    reset_update_rule_state!(cache::AbstractApproximateJacobianUpdateRuleCache, fu)
+
+Reseed the update rule cache with `fu`, the residual at the iterate the enclosing solver
+cache is being (re)initialized at, exactly as `InternalAPI.init` seeds it.
+
+Secant-type update rules difference the current residual against the previous iterate's,
+which they store across iterations. That stored residual is not reachable from
+`InternalAPI.reinit_self!`, which runs on the nested caches before the enclosing cache has
+evaluated the residual at the new iterate, so the enclosing cache calls this afterwards
+instead. The default is a no-op, which is correct for update rules whose cache holds only
+scratch buffers.
+"""
+reset_update_rule_state!(::AbstractApproximateJacobianUpdateRuleCache, fu) = nothing
 
 """
     AbstractResetCondition
--- a/lib/NonlinearSolveQuasiNewton/Project.toml
+++ b/lib/NonlinearSolveQuasiNewton/Project.toml
@@ -1,7 +1,7 @@
 name = "NonlinearSolveQuasiNewton"
 uuid = "9a2c21bd-3a47-402d-9113-8faf9a0ee114"
 authors = ["Avik Pal <avikpal@mit.edu> and contributors"]
-version = "1.15.1"
+version = "1.15.2"
 
 [deps]
 ArrayInterface = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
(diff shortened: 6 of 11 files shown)


ChrisRackauckas-Claude:
Rebased onto `master` now that #1145 has merged. The branch had a conflicting merge state because #1145 was squash-merged; replayed the single reinit commit onto the new `master` with `git rebase --onto`, so the PR is now that one commit and nothing else.

Re-validated after the rebase, in `lib/NonlinearSolveQuasiNewton`:

```
GROUP=Core   ->  1064 assertions, 0 failures
  Broyden                                               |  594    594  7m19.8s
  Klement                                               |  216    216  4m11.3s
  LimitedMemoryBroyden                                  |   99     99  2m30.9s
  LimitedMemoryBroyden: continuation from a nearby root |    1      1  1.7s
  reinit! resets the update rule state                  |   63     63  15.8s

GROUP=QA     ->  QA | 20  20  1m06.1s
```

Runic and `typos` clean. Version bumps still line up against the new `master`: `NonlinearSolveQuasiNewton` 1.15.1 (as merged by #1145) -> 1.15.2.

PR #1191, opened 2026-08-25
Add dependabot for julia updates

### LINKED ISSUE #1183: No dependabot or CompatHelper
It looks like since #954 there is neither CompatHelper nor dependabot used for updating Julia compat bounds. This means that newer versions of dependencies are not compatible with NonlinearSolve.jl and no automated PRs to make it compatible are created. This is, e.g., the case for NLSolve.jl v5. It would be good to add a `package-ecosystem: "julia"` section to the dependabot job.

JoshuaLampert:
## Checklist

## Additional context

This adds a dependabot setup for julia updates and therefore closes #1183. For some reason #954 just removed the CompatHelper setup without adding an analogous dependabot setting. This looks unintended to me (the bot wrote "Remove retired `CompatHelper.yml` (Dependabot drives `[compat]`)", which is wrong). I used the same setup as in OrdinaryDiffEq.jl adapted to this repo. Once this is merged, dependabot will probably create a few PRs updating compat bounds. I would be thankful if these are handled timely, such that packages depending on NonlinearSolve.jl can update their compats as well to allow for newer versions.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,26 @@ updates:
     directory: "/" # Location of package manifests
     schedule:
       interval: "weekly"
+  - package-ecosystem: "julia"
+    directories:
+      - "/"
+      - "/docs"
+      - "/lib/*"
+      - "/lib/*/test/*"
+      - "/test/Adjoint"
+      - "/test/Bounds"
+      - "/test/Downstream"
+      - "/test/gpu"
+      - "/test/qa"
+      - "/test/trim"
+      - "/test/Wrappers"
+    schedule:
+      interval: "daily"
+    open-pull-requests-limit: 10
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -5,3 +5,26 @@ updates:
...
+      - "/docs"
+      - "/lib/*"
+      - "/lib/*/test/*"
+      - "/test/Adjoint"
+      - "/test/Bounds"
+      - "/test/Downstream"
+      - "/test/gpu"
+      - "/test/qa"
+      - "/test/trim"
+      - "/test/Wrappers"
```
ChrisRackauckas:
```suggestion
      - "/test/*"
```

PR #1226, opened 2026-09-04
NonlinearSolveBase: rebuild a pre-wrapped function when its signatures do not match u0/p

ChrisRackauckas-Claude:

## What changed and why

`maybe_wrap_nonlinear_f` returned an already-wrapped `AutoSpecializeCallable` unchanged. A wrapper built by `get_concrete_problem` carries `FunctionWrappersWrapper` signatures for exactly the state and parameter types it was wrapped with, so a problem that was wrapped at construction time and then `remake`d with different types (another value eltype, a different array type, an outer-AD Dual) was handed a wrapper with no matching signature: the call throws `NoFunctionWrapperFoundError` (or falls into the slow fallback path, depending on the policy).

The check is now general: `wrapper_accepts(f, (u0, u0, p))` asks whether the stored wrapper has a signature for the current arguments. If it does, the wrapper is reused; if not, the raw callable is recovered with `get_raw_f` and goes through the normal wrapping path, which rebuilds the wrapper for the new types, or leaves the function raw when `u0` is Dual-typed (the existing rule for outer AD). `_despecialize_parameters` also skips its `remake` when the problem is already despecialized and wrapped, so concretizing an already-concretized problem is a no-op.

Motivation: https://github.com/SciML/ModelingToolkit.jl/pull/5062 stores initialization problems pre-wrapped; `ReconstructInitializeprob` / `remake` later change their `u0`/`p` types. That PR does not require this one (the `AllowNonIsBits` fallback keeps non-isbits cases working), but benefits from it.

## Verification

New testset `maybe_wrap_nonlinear_f rebuilds a pre-wrapped f whose signatures do not match u0` in `lib/NonlinearSolveBase/test/runtests.jl`: wrap at construction, then `remake` with (a) `Float32` `u0`/`p` — expects a rebuilt wrapper that accepts the new types and computes the right residual, and (b) an outer-AD Dual `u0`/`p` — expects the raw callable and correct values/partials; both for `AutoSpecialize` and `AutoDespecialize`.

Against **master** `NonlinearSolveBase` (`78b1484f`), same testset (`wrapper_accepts` does not exist there, so those assertions error; the Float32 call itself throws):

```
maybe_wrap_nonlinear_f rebuilds a pre-wrapped f whose signatures do not match u0: Error During Test at newtest_only.jl:17
  Expression: NonlinearSolveBase.wrapper_accepts(wrapped.f.f, (wrapped.u0, wrapped.u0, wrapped.p))
...
maybe_wrap_nonlinear_f rebuilds a pre-wrapped f whose signatures do not match u0: Error During Test at newtest_only.jl:2
  No matching function wrapper was found!
Test Summary:                                                                    | Pass  Error  Total  Time
maybe_wrap_nonlinear_f rebuilds a pre-wrapped f whose signatures do not match u0 |    3      4      7  3.8s
```

With this PR, `NONLINEARSOLVE_TEST_GROUP=Core julia --project=lib/NonlinearSolveBase -e 'using Pkg; Pkg.test()'` (Julia 1.12.7):

```
maybe_wrap_nonlinear_f rebuilds a pre-wrapped f whose signatures do not match u0 |   20     20  2.7s
...
```

`NONLINEARSOLVE_TEST_GROUP=QA`: `Testing NonlinearSolveBase tests passed`. Runic and typos clean on the changed files.

## Not verified

- Downstream solver-package groups (NonlinearSolveFirstOrder etc.), Enzyme paths, GPU arrays; only NonlinearSolveBase `Core` and `QA` were run locally.
- Julia 1.10.

## For the reviewer

- `wrapper_accepts` compares against the base signature only (`(u0, u0, p)`); the inner-AD Dual signatures the ForwardDiff extension adds are keyed off that base type, so a base match implies the Dual ones match too.
- `_despecialize_parameters`' early return assumes that when `f.f` is already wrapped and `p` is already `DespecializedParameters`, the other callbacks (`jac`, …) were wrapped in the same earlier pass. That holds for anything produced by `maybe_wrap_f`.
- Version bumped to 2.49.3.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (model: claude-fable-5-1)
https://claude.ai/code/session_01FeJYXni9MkJhPFA9yxYvfn

### DIFF
--- a/lib/NonlinearSolveBase/Project.toml
+++ b/lib/NonlinearSolveBase/Project.toml
@@ -1,6 +1,6 @@
 name = "NonlinearSolveBase"
 uuid = "be0214bd-f91f-a760-ac4e-3421ce2b2da0"
-version = "2.49.2"
+version = "2.49.3"
 authors = ["Avik Pal <avikpal@mit.edu> and contributors"]
 
 [deps]
--- a/lib/NonlinearSolveBase/src/autospecialize.jl
+++ b/lib/NonlinearSolveBase/src/autospecialize.jl
@@ -83,6 +83,8 @@ end
 
 function _despecialize_parameters(prob)
     SciMLBase.specialization(prob.f) === SciMLBase.AutoDespecialize || return prob
+    # Already despecialized and wrapped (e.g. concretized at construction): skip the `remake`.
+    prob.p isa SciMLBase.DespecializedParameters && is_fw_wrapped(prob.f.f) && return prob
     f = _map_parameter_callbacks(_wrap_parameter_callback, prob.f)
     p = SciMLBase.DespecializedParameters(prob.p)
     return SciMLBase.remake(prob; f, p)
@@ -222,7 +224,9 @@ Non-array state (e.g. scalar `Number` u0) is not wrapped because the Dual-aware
 because `promote_u0` upgrades `u0` to a `Dual`-eltype array whenever the user's
 outer-AD pass injected duals into `p`; in that case the wrapper's signatures would
 be keyed off the outer Dual tag and miss the inner value-typed dispatch that the
-forward-diff extension builds via `Utils.value` / `nodual_value`.
+forward-diff extension builds via `Utils.value` / `nodual_value`. For the same reason an
+already-wrapped function is unwrapped (via [`get_raw_f`](@ref)) when `u0` has since been
+promoted to a Dual eltype, e.g. by `remake` of a problem that was wrapped at construction.
 """
 function maybe_wrap_nonlinear_f(prob::AbstractNonlinearProblem)
     u0 = prob.u0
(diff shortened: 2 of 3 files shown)


ChrisRackauckas-Claude:
Release ordering: https://github.com/SciML/NonlinearSolve.jl/pull/1228 needs to own NonlinearSolveBase 2.49.3 because https://github.com/SciML/ModelingToolkit.jl/pull/5062 now requires its reverse-mode AD fixes. This PR currently also claims 2.49.3. After pull request 1228 is reviewed, merged, and released, this branch should rebase on the updated base and bump NonlinearSolveBase to 2.49.4 before merge. I have not rewritten or force-pushed this branch.

PR #1287, opened 2026-09-14
Solve the trust-region subproblem nearly exactly (Moré) and make it the TrustRegion default

ChrisRackauckas-Claude:
## Summary

Makes `TrustRegion()` solve its subproblem nearly exactly — Moré's safeguarded iteration on the damping parameter (MINPACK `lmpar`), paired with Moré's step-following radius update (`lmder`) — and flips the default from dogleg to `subproblem = TrustRegionSubproblem.More`. This is the recipe TrustRegionLeastSquares.jl and SciPy TRF use to reach 100% on the standard benchmark suite.

**Motivation.** Rerunning the 88-problem Moré–Garbow–Hillstrom + Hock–Schittkowski benchmark from [TrustRegionLeastSquares.jl](https://github.com/vcantarella/TrustRegionLeastSquares.jl)'s README against this branch, the released dogleg path solves **83/88** — identical results to the released 4.30/4.31 implementation. The failures (`mgh18`, `mgh19`, `tp267`, `tp272`, `tp333`) are ill-conditioned or badly scaled problems where the two-segment polygonal dogleg path approximates the trust-region solution curve `δu(λ)` poorly: the solver either stagnates inside the 32-step stall window or converges to a worse local basin (`mgh19` settles at cost 0.0438 vs. reference 0.0201). `mgh34` is a rank-1 linear least-squares problem whose `JᵀJ` is exactly singular. The released `LevenbergMarquardt` fails the same 5/88 in the benchmark's saved results, driven by damping-update pathologies on the same class.

With the new default `TrustRegion()`, all **88/88** problems converge to the reference-best cost.

## What changed

- **`MoreTrustRegionDescent`** (NonlinearSolveBase, exported): a safeguarded Newton iteration on `φ(λ) = ‖Dδu(λ)‖ − Δ` picks `λ` so the damped step lands on the trust-region boundary (relative step-norm tolerance 1e-4, ≤10 iterations, warm-started λ). A Gauss–Newton step lying inside the region short-circuits with `λ = 0`. On a linear-solve failure the bracket lower bound tightens and `λ` grows — matching `lmpar`.
- **Factorization reuse — the default dense path is MINPACK `lmpar`/`qrsolv`, not a per-λ refactorization.** `J` is factorized once per Jacobian by a pivoted Householder QR (a port of MINPACK `qrfac` storing the reflectors packed into the factor buffer — allocation-free, no LAPACK work-vector). Each λ then eliminates `√λD` against the triangular `R` with `n` Givens rotations (`qrsolv`), producing the upper-triangular `S` with `SᵀS = Pᵀ(JᵀJ + λD²)P` directly. Per-λ work is an O(n²) sweep plus triangular solves — `JᵀJ` is never formed, the `(m+n)×n` augmented system is never rebuilt, and the whole inner loop allocates nothing.
- **Representation-aware fallback** — non-default paths keep the damped solve formulation that fits:
  - *Explicit `linsolve`, other `linsolve_kwargs`, or `m < n`*: the augmented system `[J; √λD]` goes to `linsolve` (LinearSolve picks pivoted QR for rectangular `A`). κ(J) is never squared and `[J; √λD]` is full column rank for `λ > 0`; `vcat` keeps sparse Jacobians sparse.
  - *Square-only `linsolve`s* (e.g. `LUFactorization`, `KrylovJL_GMRES`), scalars, and `StaticArray`s: the equivalent damped normal equations `(JᵀJ + λDᵀD)δu = −Jᵀfu`, reusing the existing `needs_square_A` trait.
  - *Matrix-free Jacobians*: `[J; √λI]` is applied as a stacked `FunctionOperator` — a rectangular operator, so the default `linsolve` auto-selects LSMR and `J` is never materialized. Square-only Krylov solvers get a `JᵀJ + λI` operator instead (declared symmetric positive-definite, so CG applies).
  - *Convertible operator `jac_prototype`s* (e.g. `MatrixOperator`) are materialized once per solve only when `linsolve` needs a concrete `A`; a genuinely matrix-free `J` with a concrete-only solver throws a descriptive `ArgumentError` at init.
- **`RadiusUpdateSchemes.More`**: Moré's step-following radius update — on `ρ < shrink_threshold`, `Δ ← shrink_factor · min(Δ, 10‖Dδu‖)`; on `ρ ≥ expand_threshold` or `λ = 0`, `Δ ← expand_factor · ‖Dδu‖`. Auto-selected for `subproblem = TrustRegionSubproblem.More`, together with `lmpar`'s initial radius and 1e-3 step threshold.
- **Predicted reduction**: the descent reports the MINPACK-form model reduction `½‖Jδu‖² + λ‖Dδu‖²` through a `predicted_reduction` extra, avoiding the cancellation of `−gᵀδu − ½δuᵀJᵀJδu` on ill-conditioned `J`. A nonpositive predicted reduction (`denom ≥ 0`) now yields `ρ = −1` — previously a step predicted by the model *and* observed to worsen could produce a positive ratio and be accepted.
- **`TrustRegionSubproblem`** (exported `EnumX` enum, `Dogleg`/`More`) and **`AbstractDescentDirection`** are public API in NonlinearSolveBase: `subproblem` accepts the enum or a custom descent implementation, so both are documented extension points.
- `subproblem = TrustRegionSubproblem.Dogleg` remains fully supported and unchanged; the 23-problem suite keeps every radius-scheme variant on dogleg and adds `TrustRegion()` (the new default) as a ninth entry — broken set `[11, 21]`, identical to dogleg's.

## Performance

Benchmarked both subproblems end-to-end (median of repeated solves, this branch):

Dogleg vs Moré per-problem solve times

**88-problem NLLS suite** (TrustRegionLeastSquares README problems):

| | success | median solve | total | total steps |
|---|---|---|---|---|
| `TrustRegionSubproblem.Dogleg` | 83/88 | 0.219 ms | 238 ms | 5736 |
| `TrustRegionSubproblem.More` | **88/88** | **0.190 ms** | 372 ms | **3862** |

Moré takes **33% fewer steps** — each accepted step lands nearer the trust-region-optimal point — and after the `lmpar`/`qrsolv` rework its per-step cost is on par with dogleg's, so the median solve is now *slightly faster*. On the 5 problems dogleg fails, Moré is also faster in wall time (`tp267` 0.02×, `tp333` 0.05×, `mgh18`/`tp272` 0.14×, `mgh19` 0.36×) because dogleg burns all 400 iterations before failing.

**SciMLBenchmarks 23 NonlinearProblems** (rootfinding): identical success — 21/23 both, same broken set `[11, 21]` — median 1.31× per-solve (was ~5.9× before the `lmpar` rework). The remaining gap is dominated by the two Hammarling matrix-square-root problems (14→24.5 ms, 30→71.7 ms), which are pathological MaxIters cases where the safeguarded λ-iteration runs several sweeps per step.

**Allocations.** The dense `lmpar` path allocates nothing outside `init`: the `qrfac` refactorization, `Qᵀf` application, and per-λ Givens sweep + triangular solves all run in place on workspace buffers. A 2000-step MaxIters run on the Hammarling problem dropped from ~68k allocations (augmented per-λ path) to **87 total — fewer than dogleg's 90**. The λ-iteration runs behind a function barrier so the union-typed `lincache`/`dtd` fields dispatch once per call; listing the workspace variants concretely (`_MoreLmparWorkspace{Float32/Float64}`) lets `isa`-narrowed calls union-split to static dispatch, and every `DescentResult.extras` return builds the same NamedTuple shape so the result type stays concrete.

`subproblem = TrustRegionSubproblem.Dogleg` keeps the old behavior available for latency-sensitive rootfinding.

## Implementation notes

- Reuses the existing infrastructure end-to-end: `construct_linear_solver`, the shared `LinearSolveJLCache`, `needs_square_A`/`needs_concrete_A` traits, `linsolve_kwargs`, sparse `jac_prototype`s, `pre_inverted` Jacobians, and `reinit!`.
- The `lmpar` workspace (`_MoreLmparWorkspace`) replaces the linear cache on the dense default path: `qrfac` stores `R`'s off-diagonals in the strict upper of the factor buffer with reflector vectors in the lower trapezoid (MINPACK layout), `Rdiag`/`ipvt` carry the diagonal and permutation, and `qrsolv`'s Givens sweep leaves `S` as `sdiag` + the strict lower of the work matrix for the follow-up `q = (JᵀJ + λD²)⁻¹D²p` solve (`q = P S⁻¹ S⁻ᵀ D̃²z` — two triangular sweeps). The GN direction is a rank-truncated `P R⁻¹ Qᵀ(−f)` on the same factors. `nfactors` increments once per Jacobian, `nsolves` per λ.
- On the generic LinearSolve path, a single linear cache serves the GN solve (`λ = 0`) and every damped solve: for operator Jacobians the `FunctionOperator` is created once and its closures read a mutable `(J, λ)` cell, so `λ` updates never rebuild `A` (avoiding `set_lincache_A!`'s `copyto!` on operators). `reuse_A_if_factorization` lets the damped factorization serve the follow-up `q` solve — computed on the augmented path as `A \ [0; Dp/√λ]` since `Aᵀ[0; Dp/√λ] = D²p` exactly.
- A cached GN step/`Jᵀfu` is reused across rejected retries (Jacobian-reuse path) and invalidated via `callback_into_cache!` when a step is accepted. `dtd` stores `DᵀD` so `Dδu`/`D²δu` are diagonal scaling, not matrix products.
- Internal step/`Jᵀfu` buffers live in the vectorized linear-algebra space (`length(u)`/`length(fu)` vectors); `δu` keeps the state shape and `Utils.restructure` maps between them — matrix-shaped states work.
- Degenerate cases handled explicitly: `Jᵀfu = 0` (model-stationary → `p = 0`), and the λ floor `eps·maxdiag(JᵀJ)/min(diag D²)` under normal form, below which no positive-definite factorization can succeed (the augmented path needs only `eps` since it stays full rank).

## Test plan

- [x] New regression file `least_squares_tests__item6.jl` (58 tests): 16 hard benchmark problems (incl. `mgh16/18/19/34`, `tp267/272/312/333`) solved to reference cost under `:none` and `:jacobian` scaling; scalar states; `SVector`; sparse `jac_prototype` staying sparse; matrix-shaped `u`; `LUFactorization` normal-form routing; matrix-free LSMR (augmented op) and GMRES (normal-form op); `MatrixOperator` materialization under LU vs. matrix-free under Krylov; `reinit!`; invalid `subproblem`/scaling errors.
- [x] Full README benchmark rerun on this branch: **88/88** (`final_cost` at or below the reference best on every problem; `mgh34`'s saved CSV reference is corrupted — `-4.25` is impossible for a least-squares objective — so it is scored against the true optimum 3.0676 that every converged reference solver reaches).
- [x] 23-problem rootfind library: `TrustRegion()` broken set `[11, 21]` — identical to `subproblem = TrustRegionSubproblem.Dogleg`.
- [x] `qrfac`/`qrsolv` verified against `qr!(ColumnNorm())` and direct augmented/normal-equation solves to ~1e-15 on random, square, tall, rank-deficient, zero-column, ill-scaled, and `Float32` Jacobians.
- [x] NonlinearSolveFirstOrder Core items: General NLLS 480/480, preconditioning/postconditioning, statics, underdetermined, rank-deficient, non-finite-residual, operator-Jacobian, misc, sparsity, native bounded (6001), rootfind items, inference (`@inferred init(prob, TrustRegion())` — the descent cache's `dtd`/`lincache`/`augmented`/`rhs`/`qrhs` fields are pinned unions so runtime gates don't split the cache type).
- [x] Umbrella `NonlinearSolve` items exercising the default: reinit, matrix-state, polyalg-inclusion, ForwardDiff-through-solve (13,886 tests for `TrustRegion()` alone).
- [x] QA: NonlinearSolveFirstOrder 20/20. NonlinearSolveBase QA shows 2 failures that are **identical on master** (`implicit_sensitivity_solve` non-public import in the ForwardDiff ext; `nlls_solve_adjoint_dp`/`solve_call` unrendered public names) — pre-existing, unrelated.
- [x] Runic clean; `typos` clean.
- [x] Docs build verified locally with `linkcheck = false` (this machine has no external network — linkcheck is left to CI; the diff adds no external links). `checkdocs = :exports` passes.

## Known unrelated failure

`bounded_default_tests.jl` fails on master too: upstream commit `52b7e8c` made `SobolMultistart` the bounded default, so `init(prob)` returns a deferred `NonlinearSolveNoInitCache` and the test's `cache.caches` access is stale. Reproduced on a clean master checkout; a fix is prepared as a separate PR.

🤖 Generated with [Devin](https://devin.ai) (harness: Devin CLI 3000.10.27, model: SWE-2 Max)
Session: local Devin CLI transcript `/home/crackauc/.local/share/devin/cli/summaries/history_977c086cdf324660.md` (earlier sessions: `history_7011e97d708b489a.md`, `history_6a75b67bca6d40a2.md`, `history_6c57b556d8b14ef0.md`, `history_34d32feac8b94b6d.md`, `history_93828a790ce74f44.md`, `history_b90eec6928b74857.md`)


ChrisRackauckas-Claude:
Pushed `8b3aa4ed` — the dense default path now ports MINPACK `lmpar`/`qrfac`/`qrsolv` instead of refactorizing the `(m+n)×n` augmented system per λ:

- `J` gets one pivoted Householder QR per Jacobian (allocation-free port of `qrfac`); each λ eliminates `√λD` against the stored `R` with `n` Givens rotations, then triangular solves. Per-λ work is O(n²) with **zero allocations**.
- Effect on the 88-problem suite: median solve 1.06 ms → **0.190 ms** (now slightly *faster* than dogleg's 0.219 ms), still 88/88 with 33% fewer steps. On the 5 dogleg failures Moré is 0.02×–0.36× dogleg's wall time.
- 23-problem rootfinding suite: median slowdown 5.9× → **1.31×**.
- A 2000-step MaxIters run dropped ~68k allocations → **87 total** (dogleg: 90). The λ-loop runs behind a function barrier over the union-pinned `lincache`/`dtd` fields, and all `DescentResult.extras` returns share one NamedTuple shape so the result type stays concrete.
- Explicit `linsolve`/`linsolve_kwargs`, `m < n`, sparse, static, and matrix-free Jacobians keep the augmented/normal-equation LinearSolve paths unchanged; `qrfac`/`qrsolv` verified against `qr!(ColumnNorm())` and direct solves to ~1e-15 incl. rank-deficient and `Float32`.

Figure and CSVs on the `assets` branch updated; PR body revised with the new numbers.

ChrisRackauckas-Claude:
Pushed `b564dc4d` — degenerate paths now signal failure instead of fabricating zero-valued results:

- `_more_lmpar_gn!` returns `false` on a rank-deficient `R` (`nsing < n`) rather than zeroing the null-space components (MINPACK's "basic solution" — an arbitrary, generally not minimum-norm, point of the solution manifold). The caller falls through to the damped λ-iteration, whose `λ > 0` system is nonsingular and produces the regularized minimum-norm step.
- `_more_lmpar_qrsolv!` returns `false` when `S`'s diagonal contains a zero (unreachable for `λ > 0`, guarded anyway) instead of zero-filling the trailing solution entries; the λ-loop tightens the bracket and retries, same as a non-finite result.
- Failure-path `extras` no longer fabricate `λ = 0` — downstream, `iszero(λ)` means "undamped Gauss-Newton step" to `RadiusUpdateSchemes.More` and triggers radius expansion, which a failed solve must not claim. The extras now carry NaN sentinels (same NamedTuple shape, so inference is unaffected), and `skip_solve` reports `success = false` — a Moré direction is Δ-dependent and cannot honestly be skipped.

Verified: 88/88 on the NLLS suite (`mgh34`'s rank-1 Jacobian now routes through the λ-loop every step — same `StalledSuccess` at cost 3.0676), item6 regressions 58/58, edge cases (scalar/SVector/sparse/linsolve/scaling) pass, and the per-call descent path stays at 0 allocations.

ChrisRackauckas-Claude:
Added a `TrustRegionDogleg` alias in `d6e5eb016` so the classical dogleg variant is discoverable as a top-level algorithm:

```julia
TrustRegionDogleg()  # == TrustRegion(subproblem = TrustRegionSubproblem.Dogleg)
```

- `TrustRegion()` still defaults to `TrustRegionSubproblem.More` — the alias only makes the low-latency option easier to find.
- All `TrustRegion` keywords are forwarded (`linsolve`, `concrete_jac`, radius parameters, AD options); `subproblem` is fixed — a `Dogleg` descent instance is accepted, other values throw `ArgumentError`.
- `radius_update_scheme` defaults to `RadiusUpdateSchemes.Simple`, matching `TrustRegion(subproblem = ...Dogleg)`.
- Exported from `NonlinearSolveFirstOrder` and re-exported through `NonlinearSolve`; documented in the solver summary pages and the native solvers `@docs` block.
- New testset covers construction, descent/scheme selection, kwarg forwarding, rejection paths, the preserved Moré default, and a solve-side equivalence check.

🤖 Generated with [Devin](https://devin.ai) (harness: Devin CLI 3000.10.27, model: SWE-2 Max). Session transcript (local): /home/crackauc/.local/share/devin/cli/summaries/history_45276113e0414e71.md

ChrisRackauckas-Claude:
Latest push (`3d82d39d`) fixes the CI regressions from the earlier commits:

**Trim/JET (was 54 reports → now 0):** the `lincache` union field couldn't union-split its UnionAll members, leaving dynamic dispatch through `solve!`. All runtime gates are now compile-time inferable: `has_scaling` is a `Val` field, the `m ≥ n` gate is gone (the lmpar workspace zero-pads underdetermined Jacobians — pad rows stay zero through the Householder sweep, `R` is rank-deficient, GN declines, λ-loop solves the correct damped system), and the `linsolve_kwargs` whitelist is a `@generated` check on the NamedTuple keys so it folds. `@inferred init` holds for both the default and explicit-linsolve paths.

**CUDA:** augmented/rhs/qrhs/op-state buffers now allocate via `similar` on the state/Jacobian device; damping-block and scaling/maxdiag reductions broadcast instead of scalar-indexing under `allowscalar(false)`. GPU CI will be authoritative — I have no local CUDA device.

**downgrade-sublibraries (LinearSolve 5.6.0):** operator Jacobians with the default `linsolve` now take the normal-form operator `x ↦ (JᵀJ + λI)x` — square and SPD, so `DefaultLinearSolver` works. The non-square `[J; √λI]` operator hit a broken `init(::LinearCache)` MethodError under 5.6.0. Reproduced and verified locally at the compat floor: both paths converge to cost 2.9613.

**Also fixed:** `qtbf` was never refreshed when the rank-deficient Gauss–Newton solve returned early — the λ-loop read stale/uninitialized data. Regression tests added for `m < n` and rank-deficient systems (verified to fail on the unfixed ordering).

Re-verified on the refactored code: 88/88 on the least-squares benchmark suite, 78/78 on the least-squares test group, inference tests clean, zero allocations in the lmpar hot path and the solve-level step.

ChrisRackauckas-Claude:
CI on `35ec6f407`: **96 pass, 5 fail** — every failure is pre-existing on master or a known flake; nothing from this PR.

**Now green:**
- `Trim` (ubuntu + macos): JET reports eliminated (was 54, now 0).
- `CUDA`: real GPU run passes — device-local buffers + `diag` import fixed the scalar-indexing errors.
- `NonlinearSolveFirstOrder [QA]`, `Downgrade Sublibraries` (all sublibraries incl. FirstOrder's own tests), Runic, typos.

**Remaining 5 failures (all verified pre-existing/flaky):**
- `NonlinearSolveBase [QA]`: `implicit_sensitivity_solve` non-public import + `nlls_solve_adjoint_dp`/`solve_call` unrendered docs — fails on master runs too.
- `NonlinearSolveFirstOrder Core` (Julia 1 + lts) and `downgrade-sublibraries / FirstOrder`: `bounded_default_tests.jl:12` — `NonlinearSolveNoInitCache` has no field `caches` (`init(prob)` returns the no-init wrapper for the `SobolMultistart` default). Identical failure on master run 34835177218; the touched files (`multistart.jl`, `solve.jl`, the test) are byte-identical to master tip.
- `Downgrade Tests - Core`: Watson alg #2 = `Broyden(init_jacobian = :true_jacobian)` residual 3.32 — knife-edge quasi-Newton case; the same item file failed on master's own CI (Wood problem, Aug 15). Broyden's code path is untouched by this PR.

Locally re-verified after the inference refactor: 88/88 on the least-squares benchmark suite, 78/78 in the least-squares test group, inference tests clean, zero allocations in the lmpar hot path.

PR #1293, opened 2026-09-16
Fix ForwardDiff duals nested in structured parameters

### LINKED ISSUE #1283: Broyden/Klement on a Dual-valued NonlinearProblem: partials grow unboundedly while only value residuals are checked for convergence
`Broyden` (and `Klement`) on a `NonlinearProblem` whose `u0` is `Dual`-valued but whose `p` is not an `AbstractArray{<:Dual}` iterates directly in Dual arithmetic. Termination only checks the residual *values*, while the partials follow their own quasi-Newton recurrence and, on this problem, grow geometrically past 1e100 while the value residual is still bouncing around 1e3. Nothing ever flags the partials.

## How this surfaces

ModelingToolkit's `OverrideInit` initialization under `ForwardDiff.gradient` of a `remake`d `ODEProblem`: the initialization `NonlinearProblem` gets Dual `u0` but `p::MTKParameters`, so the `DualNonlinearProblem` hooks (value solve + implicit function theorem) do not fire and `default_nlsolve`'s `FastShortcutNonlinearPolyalg` runs `Broyden` in Dual arithmetic. If Broyden's *values* happen to reach `abstol` late (CPU/rounding dependent), partials of order 1e154 are handed to the ODE `u0`, and the whole gradient comes out as `9.898834314793689e154`. Tracked in https://github.com/SciML/ModelingToolkit.jl/issues/5125; failing CI job https://github.com/SciML/ModelingToolkit.jl/actions/runs/34798923758/job/103837606233.

## Standalone reproducer (no ModelingToolkit)

```julia
using NonlinearSolve, ForwardDiff, SciMLBase
const NLB = NonlinearSolve.NonlinearSolveBase
const x = [1.0, 0.5, 1 / 3, 0.25]
# unknowns u = [y4, y2, y1, dx2, dx3, y3, dx1, dx4]; parameters p.a (a NamedTuple field, so
# `p` is not an AbstractArray{<:Dual} and the DualNonlinearProblem hooks do not apply)
function resid!(r, u, p)
    y4, y2, y1, dx2, dx3, y3, dx1, dx4 = u
    a = p.a
    r[1] = -dx3 + y3 - a[3] * x[3]
    r[2] = -dx1 + y1 - a[1] * x[1]
    r[3] = -dx2 + y2 - a[2] * x[2]
    r[4] = -dx4 + y4 - a[4] * x[4]
    r[5] = -x[1] + y1 + y1^3
    r[6] = -x[2] - y1 + y2 + y2^3
    r[7] = -x[3] - y2 + y3 + y3^3
    r[8] = -x[4] - y3 + y4 + y4^3
    return nothing
end
struct T end
D(v, p) = ForwardDiff.Dual{T}(v, p)
a = [D(1.5, 1.0), D(2.5, 0.0), D(3.5, 0.0), D(4.5, 0.0)]      # d/da[1]
u0 = [D(0.5, 0.0), D(0.5, 0.0), D(0.5, 0.0), -a[2] * x[2] + 0.5, -a[3] * x[3] + 0.5,
      D(0.5, 0.0), -a[1] * x[1] + 0.5, -a[4] * x[4] + 0.5]
prob = NonlinearProblem(NonlinearFunction{true}(resid!), u0, (a = a,))
cache = init(prob, Broyden(); abstol = 1e-10, reltol = 1e-10)
for k in 1:1000
    SciMLBase.step!(cache)
    if k in (1, 10, 100, 200, 400, 600, 800, 1000)
        fu = NLB.get_fu(cache); u = NLB.get_u(cache)
        println("it=$k  max|value(f)|=$(maximum(abs, ForwardDiff.value.(fu)))  max|partials(u)|=$(maximum(abs, ForwardDiff.partials.(u, 1)))")
    end
    cache.retcode == ReturnCode.Default || (println("retcode=", cache.retcode, " at it=$k"); break)
end
# exact sensitivities of the solution: ∂y/∂a1 = 0, ∂dx1/∂a1 = -1
```

## Output

```
it=1  max|value(f)|=1.5630037291337027  max|partials(u)|=0.9031214116624805
it=10  max|value(f)|=876517.3917456791  max|partials(u)|=1082.8977390532436
it=100  max|value(f)|=3.1382993813680255e8  max|partials(u)|=4.7090397549673775e14
it=200  max|value(f)|=10918.409950542808  max|partials(u)|=2.9331601568920642e29
it=400  max|value(f)|=5846.287698571345  max|partials(u)|=7.271554284043444e68
it=600  max|value(f)|=1275.7841700819176  max|partials(u)|=1.157300298595181e112
it=800  max|value(f)|=1941.9099915791435  max|partials(u)|=3.183197749297849e154
it=1000  max|value(f)|=1365.3078588576664  max|partials(u)|=1.5282849763420413e198
```

`NewtonRaphson()` and the default polyalgorithm (which here gets to `NewtonRaphson` after Broyden hits `MaxIters` and Klement returns `Unstable`) converge in 5 steps with the correct partials (`∂y/∂a1 = 0`, `∂dx1/∂a1 = -1`). The Float64 version of the same problem also fails to converge with Broyden, so the values are chaotic on their own; the point is that the partials are never examined.

## Environment

```
Julia Version 1.12.7
Commit 6d172b025e4 (2026-08-15 08:05 UTC)
Build Info:
  Official https://julialang.org release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 128 × AMD EPYC 7502 32-Core Processor
  WORD_SIZE: 64
  LLVM: libLLVM-18.1.7 (ORCJIT, znver2)
  GC: Built with stock GC
Threads: 1 default, 1 interactive, 1 GC (on 128 virtual cores)
```

```
  [f6369f11] ForwardDiff v1.4.6
  [8913a72c] NonlinearSolve v4.31.0
  [0bca4576] SciMLBase v3.54.0
  (resolved) NonlinearSolveBase v2.51.0, NonlinearSolveQuasiNewton v1.15.3, NonlinearSolveFirstOrder v2.7.0
```

Same behaviour on Julia 1.10.11 (partials 1.6e221 at iteration 1000).

## Possible fixes

- For Dual-valued `u`, include the partials of the residual (or of the step) in the termination check, so a solve whose partials have diverged cannot return `Success`.
- Skip quasi-Newton members (`Broyden`, `Klement`) in the polyalgorithms when `u` is Dual-valued, since their secant updates are not tracked correctly in Dual arithmetic.
- Detect Dual tunables inside SciMLStructures-style parameter containers (`SciMLStructures.canonicalize(Tunable(), p)`), so the existing `DualNonlinearProblem` path (value solve + implicit function theorem) fires for `MTKParameters` and similar.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (model: claude-fable-5-1)
https://claude.ai/code/session_01VorhgJYhDasyVLpGNrviga

ChrisRackauckas-Claude:
## Summary

Fixes https://github.com/SciML/NonlinearSolve.jl/issues/1283 and the solver side of https://github.com/SciML/ModelingToolkit.jl/issues/5125.

`DualAbstractNonlinearProblem` only matches when `p` is literally a `Dual` or an `AbstractArray{<:Dual}`. Duals nested inside a `NamedTuple`, `Tuple`, `Dict`, or a `SciMLStructures` container (e.g. ModelingToolkit's `DespecializedParameters`) — or present only in `u0`/`tspan` — fell through to the generic solver path, which then evolved internal solver state in `Dual` arithmetic. For quasi-Newton updates the rank-one denominators vanish in value while their partials stay O(1), so `solve` could return `ReturnCode.Success` with partials off by many orders of magnitude (~1e153 on the MTK reproducer, where the answer should be ≈ -1.08).

- New `InternalAPI.forwarddiff_solve` / `forwarddiff_init` interception hooks, called from `solve_call`/`init_call` before `__solve`/`__init` dispatch (and from the `simplenonlinearsolve_solve_up` / `bracketingnonlinear_solve_up` entry points that bypass those funnels).
- The ForwardDiff extension now strips `Dual` leaves recursively out of `p`, `u0`, and `tspan` — arrays, tuples, named tuples, dicts, and `SciMLStructures` portions — solves the primal problem, recovers sensitivities via implicit differentiation (`-J_u⁻¹ J_p`), and re-wraps the solution under the original tag, including interval `left`/`right` for bracketing solvers.
- A rebuilt container that cannot actually hold primal values (e.g. `Dict{Symbol, Dual}`, whose `convert` re-wraps reals as `Dual`s) falls back to a plain `Dict` or throws a descriptive `ArgumentError` instead of recursing forever. Nested duals (Hessian) still work: each recursive solve strips one level.
- `init`/`solve!`/`reinit!` on the ForwardDiff cache handle structured `p` and dual-only `u0`, including a `reinit!` that removes all duals.
- Version/compat bumps per the monorepo pattern: `NonlinearSolveBase` 2.52.0 (new `InternalAPI` API), consumers' `NonlinearSolveBase` compat floors raised to 2.52 so the new `forwarddiff_solve`/`forwarddiff_init` symbols exist at release time.

## Reproducer (issue #1283 shape)

```julia
using NonlinearSolve, ForwardDiff

f!(du, u, p) = (du[1] = u[1]^3 + u[1] - p.tunable[1]; du)
p = (; tunable = [ForwardDiff.Dual{:d}(3.0, 1.0)])
sol = solve(NonlinearProblem(NonlinearFunction(f!), [1.0], p), Broyden())
ForwardDiff.partials(sol.u[1])  # du/dp should be 1/(3u²+1) ≈ 0.1846004942
```

**Before (unfixed `a2355937b`):** the structured `p` bypasses the ForwardDiff path; `Broyden()` runs raw duals. On the sharper Rosenbrock discriminant in the test suite the raw-dual solve returns `Success` with partials `[-7.5e-7, -8.2e-5]` where the true sensitivity is exactly `0`. On `NewtonRaphson` the nested tag mixing hard-errors:

```
Cannot determine ordering of Dual tags testtag and
ForwardDiff.Tag{NonlinearFunction{...}, ForwardDiff.Dual{:testtag, Float64, 1}}
```

The new test file `test/Core/forward_ad_tests__item4.jl` (101 tests: NamedTuple/Tuple/SciMLStructure/`DespecializedParameters`/`Dict` parameters, dual-only `u0`, interval + bracketing algs, simple solvers, NLLS, `init`/`solve!`/`reinit!`, polyalg/default) **errors on unfixed code** and passes 101/101 with this change.

**After:** partials are exact IFT values (`0.18460049422892552` on the cubic, ~1e-16 on Rosenbrock).

## Validation

- `test/Core/forward_ad_tests__item4.jl`: 101/101 pass; fails deterministically on the unfixed base (`DualMismatchError` / wrong partials).
- Real ModelingToolkit initialization (master MTK + this branch, dev'd): `ForwardDiff.gradient` through `initprob` now agrees with finite differences to ~1e-8 on **both** `use_scc = false` (AD `-1.0778858595152796` vs FD `-1.0778858785087664`) and `use_scc = true` (AD `-1.077885859515277` vs FD `-1.0778858833493388`), where it previously returned ~1e153. Per-block SCC subsolves funnel through `solve_call`, so the hook covers them.
- `GROUP=Core`, plus the `NonlinearSolveBase`, `SimpleNonlinearSolve`, and `BracketingNonlinearSolve` package suites: all green. Nested-dual (Hessian) regression `forward_ad_tests__item2.jl` and `reinit!` regression `item3` still pass.
- `NonlinearSolveBase` QA: 19/20; the only failure is the pre-existing "public API rendered in docs" flag for `:nlls_solve_adjoint_dp`/`:solve_call`, which reproduces on unmodified `a2355937b`. New internal names added to the ExplicitImports ignore list per the file's convention.
- Runic-formatted; `git diff --check` and `typos` clean.

Notes: this fixes solver-side corruption of ForwardDiff partials. MTK-side SCC initialization glue (e.g. how `use_scc` builds and links block sub-problems) remains a ModelingToolkit concern; https://github.com/SciML/ModelingToolkit.jl/pull/5158 is a separate, MTK-side workaround.

🤖 Generated with [Devin](https://devin.ai) — harness: Devin CLI 3000.10.27, model: SWE-2 Max
Session (local transcript): /home/crackauc/.local/share/devin/cli/summaries/history_6a409b39f8c647e2.md (continuation of history_c55d477b2a4d4f8b.md)


lib/NonlinearSolveBase/src/forward_diff.jl
```diff
@@ -4,9 +4,18 @@
     alg
     p
     values_p
-    partials_p
+    # Partials of the Dual leaves of `p`, or `nothing` when `p` carried no duals.
+    # Scalar, `Partials`, and vector forms all occur depending on the shape of `p`.
+    partials_p::Any
```
ChrisRackauckas: why any? this shouldn't be required.

ChrisRackauckas-Claude: Fixed in c502d01 — narrowed to `Union{Nothing, AbstractArray}`.

That bound covers everything assigned there: `ForwardDiff.Partials` (scalar `p`, from the nominal-path construction sites), arrays of partials (vector/structured `p`), and `nothing` (dual-free `p`). `ForwardDiff.Partials` itself can't be named since ForwardDiff is a weakdep of NonlinearSolveBase, so `AbstractArray` is the tightest bound available in `src/`. Bare `@concrete` inference wasn't an option because `reinit!` reassigns the field across types (e.g. `Vector{Partials}` → `nothing` when a reinit removes all duals).

ChrisRackauckas-Claude: Done in 4e49aa1 — `partials_p` is unannotated again, so `@concrete` types it concretely per construction.

The widening was only ever needed because `reinit!` reassigned the field across types (`Vector{Partials}` → `nothing` when a reinit removed all duals). But the field is write-only bookkeeping — nothing reads it — so `reinit!` now refreshes it only when the new partials fit the existing concrete type, and keeps the stale value otherwise. `u0duals` stays `Union{Nothing, AbstractVector}` since it *is* read in `solve!` and genuinely flips between absent and a dual-vector across `reinit!`.

ChrisRackauckas-Claude: Follow-up in 220fdece7 — everything is now concrete, plus JET coverage:

- `u0duals` lost its `Union{Nothing, AbstractVector}`: it's an unannotated `@concrete` field, always a `Vector{<:Dual}` — empty vector marks "no duals in u0/tspan" (`solve!` checks `isempty`, `reinit!` writes `similar(field, 0)`). The legacy 6-arg ctor synthesizes an empty vector matching `p`'s dual leaf type.
- Every field of the cache is now concretely typed per construction; `isconcretetype` assertions pin all `fieldtypes` through `init` and `reinit!` in the new testset.

JET tests added to `test/qa/qa.jl` ("JET type stability", 18/18 on Julia 1.12 / JET 0.12):
- `test_opt` on `InternalAPI.forwarddiff_solve`/`forwarddiff_init` for a non-dual problem → 0 reports: the hooks add no dispatch on the primal fast path.
- `test_call` on `solve`/`init`/`solve!`/`reinit!` across structured-`p`, `u0`-only, and flat-array dual shapes, `target_modules` filtered to repo-owned frames.
- `test_opt` on the strip/rebuild leaf helpers (`_dual_view` concrete leaves, `_zero_partials`, `_forwarddiff_combine_partials`, `nonlinearsolve_dual_solution`) — all fully inferred.
- Loading JET also auto-enables `run_qa`'s `test_package` typo-mode scan on the facade — now 0 reports.

The JET run surfaced three latent bugs, fixed in the same commit:
- `NonlinearSolveForwardDiffCache` was unbound in the facade's `src/forward_diff.jl` — `init(prob, alg)` on dual problems via extension solvers (NLsolveJL/CMINPACK/PETScSNES/…) would `UndefVarError`. Now `NonlinearSolveBase.NonlinearSolveForwardDiffCache`.
- `NonConcreteEltypeError` was unbound in `solve_call`'s u0 eltype check (dead throw path) — now `SciMLBase.NonConcreteEltypeError`.
- `step!` bound `time_start` only under `has_time_limit`; now always defined (same semantics — `time()` still only called under the guard).

Validation: `test/qa/qa.jl` 15/15 QA + 18/18 JET; NonlinearSolveBase suite green; `forward_ad_tests__item2/3/4` green (item4 = 101 tests incl. reinit! legs).

🤖 Generated with Devin CLI 3000.10.27 (model: SWE-2 Max)
Session: /home/crackauc/.local/share/devin/cli/summaries/history_d838c798eceb47f3.md (local transcript)

PR #1300, opened 2026-09-17
Add scaling = :auto to MoreTrustRegionDescent

ChrisRackauckas-Claude:
## Summary

`scaling = :jacobian` (Moré column-norm scaling) is a large win on badly scaled least-squares problems — on the MGH/HS benchmark tail it cut tp293 223→134 and tp292 135→41 steps — but it hurts well-conditioned ones (tp299 126→197), so it cannot be a blanket default. `scaling = TrustRegionScaling.Auto` makes it adaptive instead:

- The first Jacobian's **nonzero** column norms are gated on `maxⱼ‖J[:,j]‖ / minⱼ‖J[:,j]‖ > 20` (implemented on stored squared norms as `hi/lo > 20²`). Above the threshold the solve proceeds exactly as `Jacobian`; below it exactly as `None`.
- The decision is taken once per solve and **latched** — rejected-step retries never re-gate — and is re-decided after `reinit!`. Fewer than 2 nonzero columns or an all-zero Jacobian → off.
- Matrix-free Jacobians expose no column norms, so `Auto` degrades to scaling off (`Jacobian` still throws for them).
- Inference-safe: `Auto` sets `has_scaling = Val(true)` and always allocates a concrete `dtd`; an inactive gate is `dtd == 1`, which multiplies through as the identity — no runtime `Union{Nothing,Vector}` and no new allocations on the `None`/`Jacobian` paths.
- A reused first Jacobian still settles a pending gate via an `elseif` in `solve!`.

The `scaling` argument is a new `EnumX.@enumx TrustRegionScaling` (`None`/`Jacobian`/`Auto`), matching the `TrustRegionSubproblem` convention. The `:none`/`:jacobian` symbol spellings shipped in NonlinearSolveBase v2.52 are still accepted — the constructor converts them; other values raise `ArgumentError`.

`TrustRegionScaling.None` remains the default — `Auto` is opt-in, so no existing trajectory changes.

## Test plan

- New 33-assertion `"More subproblem: Auto scaling"` testset in `least_squares_tests__item6.jl`: direct `lmpar` and LU/normal-form paths, bitwise-equal trajectory vs `None` on a well-conditioned problem, gate engagement on a badly scaled one (14 → 3 steps), `reinit!` re-gating, Krylov LSMR/GMRES matrix-free fallback, scalar and StaticArray cases, `@inferred` init/solve!.
- API-surface coverage: enum values stored concretely, legacy `:none`/`:jacobian`/`:auto` symbols convert, `:bogus`/`"auto"` throw `ArgumentError`.
- Discrimination: `MoreTrustRegionDescent(scaling = TrustRegionScaling.Auto)` throws `ArgumentError` on unfixed master.
- `least_squares_tests__item6.jl`: **117/117 pass**. `inference_tests__item1.jl` clean; `operator_jacobian.jl` 14/14; `conditioning_tests__item1/2.jl` 38/38; umbrella `core_tests__item3.jl` 16/16.
- Runic `--check` clean; `typos` clean.
- Not run locally: full `Core` group (long) and GPU/Julia-pre CI lanes. `misc_tests__item9.jl` has 4 pre-existing master failures at lines 253–258 (reproduced on unmodified master; unrelated to this diff — fixed in #1302).

🤖 Generated with [Devin](https://devin.ai) — Agent-Harness: Devin CLI 3000.10.21, Agent-Model: SWE-2 Max. Session transcript (local): `/home/crackauc/.local/share/devin/cli/summaries/history_75aaeac8e79744c6.md`

Co-Authored-By: Chris Rackauckas <accounts@chrisrackauckas.com>


lib/NonlinearSolveBase/src/descent/common.jl
```diff
@@ -52,3 +52,43 @@ EnumX.@enumx TrustRegionSubproblem begin
...
+# keep accepting them so that spelling keeps working.
+_more_scaling(s::TrustRegionScaling.T) = s
+function _more_scaling(s::Symbol)
+    s === :none && return TrustRegionScaling.None
+    s === :jacobian && return TrustRegionScaling.Jacobian
+    s === :auto && return TrustRegionScaling.Auto
+    return throw(
+        ArgumentError("`scaling` must be a `TrustRegionScaling` (`None`, `Jacobian`, \
+                       or `Auto`), got `$(s)`.")
+    )
```
ChrisRackauckas: Get rid of all symbol use

ChrisRackauckas-Claude: Done in `a87fb15b2` — `_more_scaling(::Symbol)` is gone; `scaling` takes `TrustRegionScaling` and nothing else (`:auto`/`:jacobian`/`:none` now throw `ArgumentError`, covered by tests). This intentionally breaks the `scaling = :jacobian` spelling that shipped in NonlinearSolveBase v2.52.

